PR #58, opened 2015-09-07
HW0b

onefishy:
HW0b


HW0/P4/P4.py
```diff
@@ -1,28 +1,71 @@
...
+
+def kalman_filter(s_0, Sig_0, A, B, C, K, s_obs):
+	# initializations
+	Sig_k = Sig_0
+	Sig_t = Sig_k
+	s_tilde = s_0
+	s_f_pred = np.zeros((K, 6))
+	s_f_pred[0] = s_0
+	
+	for i in range(K-1):
```
danfortunato: Better to use `xrange` instead of `range`, so there's no chance of memory blowup if `K` is large. `range` creates a list in memory first and then iterates over it, whereas `xrange` is a generator, which changes the counter on the fly. (`range` was changed in Python 3 to do this.)

HW0/P5/P5.txt
```diff
@@ -1 +1,146 @@
...
+reduce the number of subtotals that remain to be summed by at most M. Let x be the least 
+natural number such that M - xN >= 2M. Then the fastest we can compute the total will be 
+x + ceil(log_2(M-xN)).
+
+*********************************
+In our case, we can compute the total for 256 bags in (30 + 4) seconds with 8 employees.
+*********************************
+
+Part 3. (Answer)
+[See answer to Part 2]
```
danfortunato: I'm not sure what your final answer to Part 2 actually is, which is when you have 8 total cashiers. You answered Part 3 in Part 2, which asks how fast we can compute the sum with infinite employees, which is 8 seconds.

HW0/P5/P5.txt
```diff
@@ -1 +1,146 @@
...
+
+Part 5. (Answer)
+Assume there's no time cost to distributing the initial pile of bags to the employees 
+(i.e. grabbing a bag doesn't count as communication). Assume it takes 1 second to communicate 
+a number between two employees and that an employee cannot simultaneously communicate with
+two people. Then, at time t, each of the subtotals being simultaneously computed will take 
+three seconds (one for receiving each summand and one for adding). Thus, the computation 
+time for our scheme (in Part 2) will take 3*ceil(log_2(N))-2 seconds with an infinite
+number of employees. This is still faster than computing the total with one employee (linear
+time).
```
danfortunato: What is the actual answer when N=256, though?

HW0/P4/P4.py
```diff
@@ -40,38 +91,95 @@
...
 
     #####################
     # Part 4:
     # Use the Kalman filter for prediction
     #####################
 
-    # B = ?
-    # C = ?
+	# define propagation error scaling matrix
+	B = np.diag(np.array([bx, by, bz, bvx, bvy, bvz]))
```
rohitahuja: Didn't know numpy had a diagonalize method. Very cool!

HW0/P5/P5.txt
```diff
@@ -1 +1,146 @@
...
+x + ceil(log_2(M-xN)).
+
+*********************************
+In our case, we can compute the total for 256 bags in (30 + 4) seconds with 8 employees.
+*********************************
+
+Part 3. (Answer)
+[See answer to Part 2]
+
+Part 4. (Answer)
```
rohitahuja: (Adding my comments here since I can't directly comment on the graph.) To graph the difference between the n and log_2(n), I made the y axis logarithmic. This would make the difference between the two clear for very large n.

HW0/P5/P5.txt
```diff
@@ -1 +1,146 @@
...
+receive each summand, summand 1 will already have been received when summand 2 become available.
+Then, it will take one additional second to receive summand 2 and one more second to add them.
+
+Note that, in this scheme, the last (k+1)-th subtotal is computed by time max_k+1 = max_k + 1, 
+where max_k is the time by which the last k-th subtotal become available. Since the last pair of 
+bags is distributed by time t=ceil(N/2), it follows that we can total N bags in, at worst, 
+ceil(N/2) + 2*ceil(log_2(N)) - 1 seconds. 
+
+*********************************
+In this case, it takes 143 seconds to total 256 bags using our scheme.
```
rohitahuja: I claimed it would take 133 seconds. This is because I drew out the cashiers and indexed them. Basically, every time a cashier gets a bag, he will spend an extra second summing it. We only need 128 cashiers since each cashier gets 2 bags. In second 2, cashier 1 receives his bags. Every second beyond that, we will have a cashier summed his bags. In seconds 4 and 5, cashier 1 and cashier 2 will communicate and sum their bags. From here, whenever a cashier sums the bags he received, he will sum with one of the two sets of people who just finished summing their bags. cashier ontinuing with the example, after second 5 (when cashier 4 finished summing his bags), cashier 1 will have the sum of his bags and cashier 2's bags, and sum that result with cashier 4's result. This will continue with the rest of the even indexed cashiers. This follows for odd indexed cashiers as well. By the end, we have cashier 128 sum his bags at the 129th second, and he communicates and sums that with two cashiers holding resulting sums. This process ends at 133 seconds.

PR #71, opened 2015-09-07
HW0b

draj36:


HW0/P6/P6B.py
```diff
@@ -15,17 +30,18 @@ def burnTime(t):
     pool = mp.Pool(P)
 
     # Use a variety of wait times
+    wait_time = [10**-6,5*10**-6,10**-5,5*10**-5,10**-4,5*10**-4,10**-3,5*10**-3,10**-2,5*10**-2,10**-1,.5,1]
```
shenjeffrey: You can also use np.logspace so you don't have to manually write it out!

HW0/P6/P6B.py
```diff
@@ -6,6 +7,20 @@
 def burnTime(t):
     time.sleep(t)
 
+def getSerialTime(waitTime,numIt):
```
shenjeffrey: I like the function!

HW0/P6/P6B.py
```diff
@@ -6,6 +7,20 @@
...
+    startT = time.time()
+    for ii in range(0,numIt):
+    	burnTime(waitTime)
+    endT = time.time()
+    return endT-startT
+
+def getParallelTime(waitTime,pool,numIt):
```
shenjeffrey: I like the function!

HW0/P4/P4.py
```diff
@@ -20,6 +20,10 @@
     rx = 1.0
     ry = 5.0
     rz = 5.0
+    C = np.zeros((3,3))
+    C[0,0] = rx
```
sbalanovich: Shouldn't this and all the ones following be reciprocals (ie 1/rx)? At least that's what I gathered from the HW spec on page 17/22.

HW0/P4/P4.py
```diff
@@ -30,47 +34,86 @@
...
 
     #####################
     # Part 4:
     # Use the Kalman filter for prediction
     #####################
 
-    # B = ?
-    # C = ?
+    B = np.zeros((6,6))
+    B[0,0] = bx
```
sbalanovich: Not that it really matters, but to keep code cleaner and to take advantage of numpy, you should try and consolidate things like this to take up less space and to be more concise. For example

HW0/P4/P4.py
```diff
@@ -30,47 +34,86 @@
...
+    sigmaCurrent = 0.01*np.identity(6)
+    C = np.zeros((3,6))
+    C[0,0] = rx
+    C[1,1] = ry
+    C[2,2] = rz
     # Compute the rest of sk using Eqs (2), (3), (4), and (5)
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    for ii in range(1,K):
```
sbalanovich: The spec asked you to create functions such as `predictS()`, `predictSig()`, etc. These would have been useful to make the below code much more readable.

HW0/P6/P6B.py
```diff
@@ -6,6 +7,20 @@
 def burnTime(t):
     time.sleep(t)
```
sbalanovich: The below functions are great, but they don't really solve the problem of code duplication (the only difference between the 2 functions is the 1-2 lines actually calling the wait time either in serial or in parallel). I think there could be a way to write a `getGeneralTime()` function and pass in some parameter to indicate serial vs parallel, to make this code even cleaner and to allow it to work beyond these two examples if ever it was necessary.

HW0/P6/P6B.py
sbalanovich: ^ Agreed with this, use linspace or logspace to make this cleaner and more maintainable.

HW0/P5/P5.txt
```diff
@@ -1 +1,79 @@
 # Your answers here.
```
oyuchoi: Your graph for 5 doesn't seem to imply any trends in performance (both look linear) and it's hard to make out that in parallel it would be logarithmic.

HW0/P5/P5.txt
```diff
@@ -1 +1,79 @@
...
+65                              130                           64
+97                              194                           33
+113                             226                           18
+121                             242                           9
+125                             250			      6
+127                             254                           3
+129                             256                           5
+136                             256                           1  
+
+c. Answer: Yes, 73 seconds. Give 4 bags to 1 employee at a time instead. Then it takes at most 2*log2(N) + 3 seconds to count N bags (where N is a power of 2) after the N bags are passed out to employees. This is because we can wait 3 seconds for the last employee to count his/her 4 bags then take 2*log2(N) seconds for the N employees to count their coins. See table below:
```
oyuchoi: Even 4 bags are suboptimal

PR #72, opened 2015-09-07
Hw0b

zhiqianseah:


HW0/P4/P4.py
```diff
@@ -40,37 +49,124 @@
...
 
     #####################
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
-    # s = ?
+    A = np.matrix([[1, 0, 0, dt, 0, 0],
```
fannyheneine: It would have been more efficient and more elegant to use a for loop to fill in the values of the array A: 
for i in range(3):
A[i][i]=1

for i in range(3,6):
A[i][i]=1-dt*c
A[i-3][i]=dt

HW0/P4/P4.py
```diff
@@ -40,37 +49,124 @@
...
+        sig_updated = inv(sigma_tilde + C.transpose() * C)
+
+        #equation 5: updateS
+        s_updated = sig_updated * (sigma_tilde * s_tilde + C.transpose() *  ob_matrix[x,:].transpose())
+
+        filtered_matrix[:,x] = s_updated.reshape(6)     
+
+        #save sigma for next iteration
+        sigma = sig_updated   
+
```
fannyheneine: It would have been easier to just call recalculate sigma in each for loop instead of calling it sig_updated and then equating sigma to it again. I understand the idea of updating the value at each loop, and hence using a sig_updated variable, but it would be more efficient to directly assign the sigma updated value.
You could directly have: sigma = inv(sigma_tilde + C.transpose() * C)

Instead of : sig_updated = inv(sigma_tilde + C.transpose() * C)
sigma=sig_updated

HW0/P5/P5.txt
```diff
@@ -1 +1,51 @@
 # Your answers here.
+1. 255 seconds
+
+2. Yes. First, split the bags into 8 groups of 32 for each cashier. The 8 cashiers will take 31 seconds to add the total sum of each group. Now, to add the 8 numbers together, it takes 1 second to add 4 pairs of numbers, then 1 second to add the remaining 2 pairs of numbers, and finally 1 second to add the last pair of numbers (i.e. it is a binary tree). Hence, it will take 34 seconds to add up the sum.
+
+3. 8 seconds. log(256)/log(2). Essentially, it is a binary tree.
+
+4. It will take log(N)/log(2)
+
+5. 16 seconds. The binary tree has a depth of 9, hence there are 8 seconds to communicate between levels.
```
fannyheneine: A bit more details or a short explanation would be good for questions 3, 4 and 5. I understand your answers as I did the same but if someone hadn't understood the logic, it would be hard to grasp the solutions with your answer.

HW0/P6/P6.txt
```diff
@@ -1 +1,5 @@
 # Your answers here
+
+1. the order of jobs ending is different every time. Some jobs may even take a long time to complete, as other jobs grab the lock first. A scenario where this is important is when we want certain jobs to finish in a particular order, for example: A program that process and output strings of text from a document sentence-by-sentence, such as a translator (Prcoess and Output are 2 different jobs). The processing part can be done in parallel, but the string output has to be in sequential order for the final output to make sense.
```
fannyheneine: Good example!

HW0/P6/P6.txt
```diff
@@ -1 +1,5 @@
...
+
+2. The serial version is faster than the parallel version for short sleep time. This is because there is overhead involved in switching between threads. When the job is finished very quickly, the overhead becomes significant, which makes the parallel version slower.
```
fannyheneine: I think that what you mean is not "When the job is finished very quickly, the overhead becomes significant" (as the overhead does not become significant but is simply significant with respect to the completion of the serial job itself), but that when the job is very quick, the overhead of parallel computing becomes longer than the serial process itself?

HW0/P4/P4.py
```diff
@@ -40,37 +49,124 @@
...
+    ob_matrix = np.matrix(observation)
+
+    #create a matrix for the un-stretched values. Populated with zeros first
+    fixed_matrix = np.zeros(ob_matrix.shape)
+
+    #calculate the un-stretched values based on matrix multiplication of
+    #each row with the C matrix. Transpose the row to a column for the multiplication
+    #then transpose it back to a row, and save it to fixed_matrix
+    for x in range (0, ob_matrix.shape[0]):
+        fixed_matrix[x,:] = (C * ob_matrix[x,:].transpose()).transpose()
```
tobytoyuito: It would have been faster to use matrix operation. For example, fixed_matrix = np.dot(C, ob_matrix.T)

HW0/P5/P5.txt
```diff
@@ -1 +1,51 @@
...
+3rd second: 20 bags to cashier 3, cashier 2 starts counting (it will take him 20 seconds)
+...
+21st second: 2 bags to cashier 21, cashier 20 starts counting (it will take him 2 seconds). Note: there are 4 bags left at this point
+22nd second: 4 bags to cashier 22. cashier 21 starts counting (it will takes 1 second)
+23rd second: cashier 22 starts counting. All the other cashiers finish counting their bags
+24th second: cashier 1 to 21 communicate to sum up the values. There will be 10 pairs with 1 lone cashier
+25th second: The cashiers adds up their pair of values
+26th second: the cashier communicate again to sum up the values, there will be 5 pairs with 1 lone cashier. At this point, cashier 22 finished counting. Therefore, there are 6 pairs to sum up. this will take another ceil(log(6)/log(2))*2 = 6 seconds
+
+Therefore, it will take a total of 32 seconds.
```
tobytoyuito: I think the fastest way takes 22 seconds.
1st second: 256 bags to cashier 1
2nd second: cashier 1 gives 128 bags to cashier 2
3rd second: cashier 1 and cashier 2 give 64 bags to cashier 3 and cashier 4
4th second: cashier 1, 2, 3, 4 give 32 bags to cashier 5, 6, 7, 8
5th second: cashier 1...8 give 16 bags to cashier 7...16
6th second: cashier 1...16 give 8 bags to cashier 17...32
7th second: cashier 1...32 give 4 bags to cashier 33...64
10th second: every cashier counts 4 bags in 3 seconds.
Now we have 32 pairs of values.
11th second: communicate among 64 cashiers
12th second: adds up and ends up with 32 values
13th second: communicate again
14th second: adds up and ends up with 16 values
15th second: communicate again
16th second: adds up and ends up with 8 values
17th second: communicate again
18th second: adds up and ends up with 4 values
19th second: communicate again
20th second: adds up and ends up with 2 values
21st second: communicate again
22nd second: 1 value. Done!

HW0/P6/P6B.py
```diff
@@ -10,22 +11,37 @@ def burnTime(t):
...
 
     for t in wait_time:
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
+
+        start = time.time()
+        for x in range(0,16):
+            burnTime(wait_time[t])
```
tobytoyuito: I think variable t is one element in wait_time. And I think this will raise error if t is not an integer. If you want to iterate through both index and element, this would be a better way:
for index, time in enumerate(wait_time)

HW0/P6/P6B.py
```diff
@@ -10,22 +11,37 @@ def burnTime(t):
...
+        end = time.time()
+        serialTime = end - start
+
+
+        start = time.time()
+        pool.map(burnTime(wait_time[t]), range(16))
```
tobytoyuito: I think map is a function that puts second variable ( which is one element in range(16) ) into first variable ( which is burnTime(wait_time[t]). However, the first variable should be a function. I would write it like this:
pool.map(burnTime, [t]*16)

HW0/P5/P5.txt
zhiqianseah: Wow, this is pretty cool. I didn't think of using the cashiers to pass out the bags.

HW0/P6/P6.txt
zhiqianseah: Yup, sorry that my explanation was a bit off.

HW0/P4/P4.py
zhiqianseah: I am not sure of the internal workings of python, but I will argue that instantiating the values directly is more efficient as it stores the values directly into the matrix, rather than instantiating a zero matrix, and then replacing with values again. Either way, it is a really slight overhead for such a small matrix

HW0/P6/P6B.py
zhiqianseah: I think that is why my graph is different from others. Thanks for pointing out this mistake.

HW0/P5/P5.txt
oyuchoi: Great suggestion @tobytoyuito!

HW0/P6/P6.txt
```diff
@@ -1 +1,5 @@
 # Your answers here
```
oyuchoi: The graph for P6 looks very noisy and doesn't seem to indicate much of a speed-up when moving to 4 processes.

PR #83, opened 2015-09-08
HW0b

foo-bar-baz-qux:


HW0/P4/P4.py
```diff
@@ -2,6 +2,23 @@
 import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
 
+def predictS(A, s_k, a):
```
chuckyouliu: I'm not sure if they have to be defined above where the functions are defined but you don't have to pass in global variables as parameters

foo-bar-baz-qux: True, though I tend to avoid using globals like this as it can lead to unexpected behaviour.

chuckyouliu: It was the only thing I could think of to comment on

HW0/P5/P5.txt
```diff
@@ -1 +1,92 @@
...
+
+It takes 256 seconds for all cashiers at the base layer to obtain bags. They then communicate the bag totals to the 128 cashiers on the 7th layer in 1 second. Then, the analysis proceeds as in question 5 with calculation and communication taking 15 seconds. It is possible for cashiers to start calculating their totals as soon as they receive their bags, but we will reach a bottleneck since only after 256 seconds do the cashiers on the base layer all have a bag. Therefore, the total time taken is the same regardless of whether they start calculating straight away or only after all base layer cashiers have bags.
+
+The total time taken is 256 + 1 + 8 + 7 = 272.
+
+### 2 bags at a time
+
+On the 7th layer, there are 128 cashiers who each require two bags. This will take 128 seconds if we can only hand bags out at a rate of 2 bags per second (per cashier). Whilst cashiers could get started calculating totals of the upper layers while we are still handing out bags, there is going to be a bottle neck since only after 129 seconds can the 7th layer be entirely completed.
+
+So, including communication time, it takes 128 + 8 + 7 = 143 seconds if we hand each cashier 2 bags at a time.
```
chuckyouliu: The cashiers don't have to wait for all the other cashiers to get a bag before they can start adding/communicating. I don't know if my answers are optimal, but I got 258 and 132 seconds respectively writing a script to do it. Here's my output for 8 bags handing 1 out at a time:

Second 1
Empl:1 
Bags:1

Second 2
Employee 1 communicates to 2
Empl:2 
Bags:2

Second 3
Employee 2 sums
Empl:2 3 
Bags:1 1

Second 4
Employee 2 communicates to 4
Empl:3 4 
Bags:1 2

Second 5
Employee 3 communicates to 5
Employee 4 sums
Empl:4 5 
Bags:1 2

Second 6
Employee 4 communicates to 6
Employee 5 sums
Empl:5 6 
Bags:1 2

Second 7
Employee 5 communicates to 7
Employee 6 sums
Empl:6 7 
Bags:1 2

Second 8
Employee 6 communicates to 8
Employee 7 sums
Empl:7 8 
Bags:1 2

Second 9
Employee 7 communicates to 8
Employee 8 sums
Empl:8 
Bags:2

Second 10
Employee 8 sums
Empl:8 
Bags:1

Completed in 10 seconds

(Sorry for the long comment LOL)

HW0/P6/P6B.py
```diff
@@ -16,16 +16,25 @@ def burnTime(t):
...
+        serialTime = 0.
+        parallelTime = 0.
+
+        s_start = time.time()
+        for count in xrange(N):
+            burnTime(t)
+        serialTime = time.time() - s_start
+
+        p_start = time.time()
+        pool.map( burnTime(t), range(N) )
```
chuckyouliu: You shouldn't be able to get 16x speedup with 4 processes. I suspect the error is here where you're passing in burnTime(t), instead of having an array of t's of length N as your second argument.

foo-bar-baz-qux: I was curious about this, but couldn't come to a definitive answer. The expectation is that it should be a 4x improvement, but it appears that the sub-processes created by the Python multiprocessing library don't block when you call the sleep function. Instead, it appears to use some kind of threading mechanism to continue execution on the other tasks its assigned.

The implementation details for map are sparse, but it does split up the input array into chunks and sends it to the processes. I tried your suggestion, but it results in a similar graph on my system (`pool.map( burnTime, [t]*N )`). It should be equivalent to the original code in any case.

Any thoughts on this issue, @thouis?

chuckyouliu: I'm consistently getting 16x with pool.map(burnTime(t), range(N)) and 4x with pool.map(burnTime, [t]*N). Pretty strange

chuckyouliu: Yup, just tested - your speedup corresponds to the value of N given the argument above, and since N=16 you get that speedup

foo-bar-baz-qux: It seems to be system dependent; I'm still getting 16x and the same graph. When I output the order of the function calls, it sleeps 16 times in a row, and then they all wake up around the same time.

HW0/P4/P4.py
```diff
@@ -40,37 +62,86 @@
...
 
-    # B = ?
-    # C = ?
-
     # Initial conditions for s0 and Sigma0
-    # Compute the rest of sk using Eqs (2), (3), (4), and (5)
+    sig_0 = 0.01 * np.eye(6)
+
+    B = np.zeros((6,6))
+    B[np.diag_indices(6)] = [bx, by, bz, bvx, bvy, bvz]
```
fpeng01: np.diag_indices() is a good function. I'll try to use this function next time to avoid the long code block creating a 6 by 6 matrix with only diagonal entries nonzero.

HW0/P5/P5.txt
```diff
@@ -1 +1,92 @@
...
+Suppose there is an incredible communicator who can simultaneously communicate the labels on the bags to all cashiers on the 7th layer (without moving any of the bags.) Then it would make sense for us to give this person all 256 bags in 1 second. This person then takes 2 seconds to concurrently communicate 2 numbers to each cashier on the 7th layer (128 cashiers.)
+
+The total time taken would then be 1 + 2 + (8 + 7) = 18 seconds.
+
+Even if we cannot assume such superhuman abilities, we can still do better by giving cashiers more bags and then having them communicate or give bags to other cashiers. Let’s assume we need to distribute 6 bags to 3 cashiers (A, B, C) such that each cashier has 2 bags.
+
+1 second: We give cashier A 4 bags.
+2 seconds: We give cashier B 2 bags. Cashier A gives C 2 bags.
+
+The total time is now 2 seconds instead of 3 seconds since we benefit from A’s help. The point is, we want to reduce the amount of times we need to hand out bags by increasing the number of bags we give out each time.
```
fpeng01: I guess for this problem, more employees means more handing&communication time but less counting time.
So the problem is a trade-off. Not quite sure that communication between cashiers is easier and takes less time. but the idea of superhuman ability is interesting.

HW0/P6/P6.txt
```diff
@@ -1 +1,62 @@
...
+
+## How this affects how we program
+
+As commands can be executed in a seemingly random fashion, additional tools such as mutexes and semaphores need to be used if particular guarantees are required. We also need to be mindful of strange effects that can occur due to race conditions. These might only occur once in a while, making debugging particularly difficult.
+
+## Scenarios where this is important
+
+Whenever we have multiple threads/processes potentially reading and modifying shared memory, this becomes critical. Suppose we are trying to execute a particular function `doTenTimes()` 10 times in parallel. The logical thing is to try and keep a counter of how many times the function has executed; let this be stored in some shared variable `counter`. Each thread/process checks the counter variable with something like 
+
+```python
```
fpeng01: Good example with python code. I never think of actually writing a code snippet as an example. This is a more "engineering" way of answering the question.

HW0/P5/P5.txt
zezhouliu: @chuckyouliu is correct, they shouldn't wait for everyone to receive their items before they start computing

HW0/P6/P6B.py
zezhouliu: @chuckyouliu  Yes, the speed-up should only get up to 4x if you are using 4 processes.

PR #97, opened 2015-09-08
HW0b

twkillian:
Included in the pull request are the updated/new files generated to fulfill the requirements of HW0b (problems 4-6).


HW0/P4/P4.py
```diff
@@ -30,47 +32,114 @@
...
 
     #####################
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
+	#Propagation Matrix
+    A = np.diag([1, 1, 1, 1-c*dt, 1-c*dt, 1-c*dt]) + np.diag([dt, dt, dt],3)
```
ymiyamot: This is a cute way of defining the matrix!  Good to know.

HW0/P4/P4.py
```diff
@@ -30,47 +32,114 @@
...
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-k', label='Blind trajectory')
+    sK = np.matrix(np.zeros((6,np.amax(TruTraj.shape)))) #Initializing array for state predictions
+    sK[:,0] = s0                              #Placing initial measurement as first column
+    
+    for i in range(1,np.amax(TruTraj.shape)):
+         sK[:,i] = A * sK[:,i-1] + a
+    
+    sK = np.asarray(sK) # Converting state prediction matrix to an array (for plotting??)
```
ymiyamot: An alternative is to start off with sK being an np.array (rather than np.matrix) and then use np.dot to do matrix multiplication, but this looks like a perfectly good way too.

HW0/P4/P4.py
```diff
@@ -30,47 +32,114 @@
...
-    #         '-r', label='Filtered trajectory')
+    
+    def predictS(priorState): # Eq(2)
+        #Will fill out code structure here based on state update from part 3
+        predS = np.add(np.dot(A,priorState),a)
+        return predS
+         
+    def predictSig(priorSig): # Eq(3)
+        #Predict SigmaK update
+        predSig = np.linalg.inv(np.add(np.dot(np.dot(A, priorSig),np.transpose(A)),np.dot(B,np.transpose(B))))
```
ymiyamot: You could condense the notation a bit if you wanted to: (1) np.transpose(A) is the same as A.T, (2) np.dot(A, B) is the same as A.dot(B)

HW0/P6/P6.txt
```diff
@@ -1 +1,5 @@
...
+
+2.As our wait time increases the time to run to the the parallel program increases relative to the serial program. This is due to the wait time stacking up the flow of jobs getting completed serially while the parallel loop is pushing through multiple jobs at a time despite the longer wait time being enforced.It is possible for a parallel program to take longer to run due to communication overhead when the wait (job) times are small enough to be kept from bogging down the serial program.
```
ymiyamot: Slightly confused by this statement. As wait time increases, the plot indicates that the ratio (serial / parallel) increases – so the time to run the parallel program decreases relative to the serial program, no? I could be misinterpreting something.

The shape of the plot looks correct, but I'm a little puzzled by the ratio reaching asymptote near 1. This would suggest no benefit of parallelizing. Maybe the units on the y-axis are off?

PR #99, opened 2015-09-09
HW0b

Peilin-D:
First pull request on Sep 8th


HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+
+#4 
+(P5.py) 
+When infinite employees, the calculation is just like question #3. So the shortest time to count should be ceil(log_2(N)), which means round up.
+When a lone employee count all the bags, the time will be linear to the #bags, which is obvious.
+
+#5 
+The shortest time it costs for the employees to count should now add the communication time which is 1 second for a level in the "binary tree" except for the highest level.
+So for 8 employees, it will now take 37 seconds.
+For infinite employees, it will now take 16 seconds for fastest counting.
```
virgodi: I think that for infinite employees, they would actually need 15 seconds since there should not be any lag when performing the first sums. I thought that we needed 1s to go from 256 to 128 amounts, then 2s to go to 64 (1 for communication and 1 for computing), 2s to go to 32 etc until 1 amount: 1+7*2=15 s.

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
+     #s=predictS(s,A,a)
+     s=updateS(predictSig(A,Sig,B),predictS(s,A,a),C,m[k,:])
+     Sig=updateSig(predictSig(A,Sig,B),C)
+     S[:,k+1]=s
+     k=k+1
+    ax.plot(S[0], S[1], S[2],
+             '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
```
virgodi: Your code is very readable ! Well done !

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
-# Your answers here.
```
virgodi: I don't think the answer is Log base 2. It is the case for powers of 2 but for a general time T, you need to account for an extra second. I believe the answer is int(log_2(N)) where int rounds up the number to the next biggest integer.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+When a lone employee count all the bags, the time will be linear to the #bags, which is obvious.
+
+#5 
+The shortest time it costs for the employees to count should now add the communication time which is 1 second for a level in the "binary tree" except for the highest level.
+So for 8 employees, it will now take 37 seconds.
+For infinite employees, it will now take 16 seconds for fastest counting.
+
+#6
+Assume that the employees starts to count and communicate the next second I give them the bags, and they just communicate in serial, not parallel.
+If each employee takes 1 bag from me, that will be 256 seconds for taking the bag, and for 256 employees to sum up all the bags, it will take 2*255 seconds(communications and add numbers), but they can start to count when I give out bags, and the overlap time will be 256-3+1 because the overlap begins at the time I give bag to the 3rd person. So the total time it costs will be 512 seconds.
```
adisuissa: As you noted in the previous parts of the question, the communication and summation can be done in parallel and require less than 2*255 seconds.

PR #103, opened 2015-09-09
Hw0b

ppgmg:
Submission for HW0 Problems 4-6
Kendrick Lo (Harvard ID: 70984997) for CS 205


HW0/P5/P5.py
```diff
@@ -0,0 +1,25 @@
...
+z = x - 1
+
+plt.plot(x, y, 'o', label='as many cashiers as you want')
+plt.plot(x, z, '.', label='one cashier')
+plt.legend(loc = 'lower right')
+plt.xlabel("number of bags")
+plt.ylabel("counting time (seconds) - log scale")
+plt.title("Counting Time for N Bags")
+
+plt.yscale('log')  # show log scale on y-axis
```
jaco8: It is good to do log scale since it is better to see the shape.

HW0/P5/P5.txt
```diff
@@ -1 +1,327 @@
...
+
+Summary
+-------
+
+1. 255 seconds
+2. 34 seconds
+3. 8 seconds
+4. log N seconds
+5. 15 seconds
+6. 272 seconds, 143 seconds
```
jaco8: What my understanding for 1 bag is : For total 256 bags, we ignore the handing time first, each person takes 1 bag, so we have 256 groups. Same idea as problem 5, but a little bit different is 256 people should communicate first, it takes an extra 1 second. So the time without handing time is 16 seconds. Now for the handing time, from 256—>1, it takes 8 seconds, but we need 1 extra second to pass bags to cashier. Therefore, the total time will be 16 + 8 + 1 = 25 sec

HW0/P5/P5.txt
```diff
@@ -1 +1,327 @@
...
+   Wait-and-Combine (16 workers): 39 seconds
```
jaco8: 2 bags: Same idea as problem 5, for the handing time, from 256—>1 which is cost 8 seconds. Therefore, the total time will be 15 + 8 = 23 seconds.

HW0/P6/P6.txt
```diff
@@ -1 +1,199 @@
...
+Problem 4, for columns in matrix S as an example. Each set of values for
+a given column in s depended on values of the preceding column (which is
+associated with the previous time step); if we try to apply parallel
+programming techniques and do so in a way that changes the order of the
+columns for example, the resultant model will be different (and likely
+incorrect).
+
+2. Running P6B.py
+
+From the plot, in general there seems to be a trend that suggests the
```
jaco8: I pretty agree with your answer. Under the condition from 10^(-6)—>10^(-4), since the task is relatively small, serial version is faster than parallel version. However, under the condition from 10^(-3)->10^(0), the task is relatively large, parallel version will take shorter than serial version. Therefore, if we program this in parallel, it will take much more time to run this program if the task is relatively small, since opening a processor costs much more time. But if the task is relatively large, using multiprocessor will make the whole process much faster.

HW0/P5/P5.txt
ppgmg: Thanks for your comment.  I'm not sure I agree with that interpretation though. The question asks what happens if each cashier comes up to you and takes 1 bag. The question also says that you can't hand bags to more than one person at a time. So I'm not sure for the case with 1 bag how it can be faster than 256 seconds, if each cashier can take 1 bag from you, and that individual action takes 1 second long...

HW0/P5/P5.txt
ppgmg: Thanks again for this note. I guess it depends on how you interpret the question being asked.

I have (now) seen some solutions to the very last part of question 6, where all 256 bags are passed off to the first cashier, and then the cashiers themselves divide up bags amongst themselves in a binary-treelike format. This is actually quite clever, although I do think that type of solution requires certain assumptions to be made that are arguably not clear from the question. 

For example, the question says that YOU can't be handing bags to more than one person at a time, but is completely silent on whether cashiers can be handing bags to more than one other cashier at a time. Also, we have to assume that cashiers can perform distributing tasks in addition to merely counting and communicating tasks -- there is some suggestion of this from the question when it mentions "between two people". My answers assuming cashiers can only count & communicate (in the spirit of questions 1-5), but if they can do more I'd agree applying divide-and-conquer to the distribution task itself is very fast too!

HW0/P6/P6.txt
ppgmg: Thanks for your feedback!

HW0/P5/P5.txt
kevinchen1000: We are not concerned about the offset of (+1/-1 seconds). The key is to realize the log(N) trend. The discussion given is very thorough, great work!

HW0/P6/P6.txt
kevinchen1000: Good discussion !

PR #108, opened 2015-09-09
Hw0b

sdaulton:


HW0/P4/P4.py
```diff
@@ -30,47 +29,112 @@
...
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
-    # s = ?
-
     # Initial conditions for s0
+    A = np.matrix([[1,0,0,dt,0,0],[0,1,0,0,dt,0],[0,0,1,0,0,dt],[0,0,0,(1 - c * dt),0,0],[0,0,0,0,(1 - c * dt),0],[0,0,0,0,0,(1 - c * dt)]],dtype=float)
```
rpestourie: Good handling of matrices in python!
You can also use sparse numpy arrays with np.diag()

HW0/P4/P4.py
```diff
@@ -30,47 +29,112 @@
...
 
-    # B = ?
-    # C = ?
+    B = np.zeros((6,6))
+    B[0,0] = bx
+    B[1,1] = by
+    B[2,2] = bz
+    B[3,3] = bvx
+    B[4,4] = bvy
+    B[5,5] = bvz
```
rpestourie: same comment here, using np.diag() is a little faster but this works too!

HW0/P5/P5.txt
```diff
@@ -1 +1,30 @@
...
+
+If each cashier takes 2 bags from you, only 128 cashiers need to collect bags and no communication occurs before the first calculation.
+Note: this answer assumes that a cashier with 2 bags can instantly read the number from both bags.
+# of seconds = distribution time + calculation time + communication time = 128 + 8 + 7 = 143 seconds
+
+The better strategy though would be for you to hand each cashier that approaches you half of the bags in your possession, while you have more that 2 bags.  When you have 2 bags, give both to the next cashier.  Each cashier with more than 2 bags, should also give half their bags to a bag-less cashier.  
+Here, you would give 128 bags to the first cashier ==> 1 second
+Then you would each give 64 bags to 2 other cashiers ==> 1 second
+Note that there are only 7 rounds of bag distribution since each cashier would keep 2 bags.  Thus, distributing bags would take 7 seconds.
+# of seconds = distribution time + calculation time + communication time =7 + 8 + 7 = 22 seconds
```
rpestourie: I am not sure it is correct. The first cashier will take 127 sec to compute the 128 bags which is way more than the 22 seconds that you found. Good explanation though.

HW0/P6/P6B.py
```diff
@@ -16,16 +16,22 @@ def burnTime(t):
...
+    wait_time = [10**(x/10.0) for x in range(-60,1,1)]
 
     for t in wait_time:
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
-
+        parallelStart = time.time()
+        pool.map(burnTime(t), range(N))
```
rpestourie: I don't think it is correct here. 
You want pool.map(burnTime, [t]*N)
[t]*N is the vector on which you map, you want it to wait 16 times during t (not during range(16))
But it is a good try!

I think that is why you get a speedup up to 16, it should not be better than 4 since you only have 4 processors

HW0/P5/P5.txt
ppgmg: Hi Sam! If I understand what you've said correctly, I think part of your strategy is to offload the task of distributing bags to other cashiers, since "each cashier with more than 2 bags should give half their bags to a bag-less cashier".  Very clever! This gets the bags distributed faster and then you can apply the log n algorithm to add/combine.  But I think if you're going to do this, you have to explicitly assume that unlike the initial distribution where bags can only be passed to one cashier at a time, that the cashiers themselves are not bound by such a restriction (e.g. when the first cashier passes 64 bags to two other cashiers, that he can do this simultaneously and in 1 second rather than sequentially). So I'd probably just state this assumption (or add another second to each distribution).

HW0/P6/P6B.py
ppgmg: I think Raphael is correct here. I believe the map function takes each element from the second argument and feeds it into the function. So we want to feed in [t, t, t, t, .... ,t ] into burnTime so that the sleep lasts t seconds. In this case, rather than feeding in [t, t, t .... t] you'd be feeding in [1, 2, 3.... N], which means that the parallel version is sleeping at varying times while the serial version is always sleeping for t second.

sdaulton: Despite the notation being very confusing, map can actually be used in this way.  If you test the code, it is as if I fed in [t,t,t,t,t,t,...,t].  I shouldn't use map this way in the future.  It is not clear code.  Thanks for pointing it out.

HW0/P4/P4.py
```diff
@@ -30,47 +29,112 @@
...
-    #         '.g', label='Observed trajectory')
+    measurements_filename = "P4_measurements.txt"
+    measurements = np.loadtxt(measurements_filename,delimiter=',')
+    C = np.matrix(np.zeros((3,6)))
+    C[0,0] = rx
+    C[1,1] = ry
+    C[2,2] = rz
+    C_inv = C.getI()
+    x_approx = (C_inv * measurements.T).T
+    ax.plot(x_coords, y_coords, z_coords,
```
adisuissa: You use the s_true values, and not the x_approx values.

HW0/P6/P6B.py

PR #109, opened 2015-09-09
Hw0b

zxjyjr:
Jing's HW0b


HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
     #
     # Load true trajectory and plot it
     # Normally, this data wouldn't be available in the real world
+
```
vivjay30: Because this is all within the if statement on line 6 you'll want to keep at least one tab indent throughout, more if you have more conditionals/loops

HW0/P5/P5.txt
```diff
@@ -1 +1,12 @@
...
+2.it's faster for 8 cashier to do it. They can finish it in 31+log2(8)=34 seconds
+
+3.if the store has infinite employers, then it takes log2(256)=8 seconds
+
+5.it takes 15 seconds for infinite cashiers to count the coins.
+
+6.If cashier come and take 1 bag each time, then the total time would be 15+256=271 seconds
+  If cashier come and take 2 bags each time, then the total time would be 15+128=143 seconds
+
+A better strategy would be give 16 person each 16 bags, each cashier add 16 bags alone, then they can communicate to sum all the bags. In this case, the total time would be: 16+15+8=39 seconds
```
vivjay30: I had a slightly faster way. If you give half the bags to one person then those two give half to two more and so on until everyone has 2 bags and you count back you can do it in 20 seconds

HW0/P6/P6.txt
```diff
@@ -1 +1,11 @@
 # Your answers here
+In each parallel run, the order to run the four function in parallel is random. For example, the order of 0,1,2,3 can be random, it could be 1,2,3,0
+For example, we have multiple steps to finish a calculation, and we also have several sub-steps in each steps. If the steps need to be finished step by step, and the sub-steps is independent of other sub-steps in each step. Then we can run sub-steps in parallel, while we mustr run steps in series but not in parallel.
+Otherwise, we would get total wrong results if we run steps in parallel.
+
+For P6B
+When the sleep time is very small compared to communicating time, then the total time depends on the communicating time. Serial computing is even faster than parallel computing, that might due to parallel computing spend lots of time on communicating since it has 4 processes, while serial computing does not. 
+
+But when sleep time is large compared with communicating time, then multi-processes will greatly reduce the total time by a factor of 4 since it has 4 processes. That is, the ratio will eventually reach a plateau of 4. But there is a transition from communicating dominant to sleep time dominant.
```
vivjay30: Your graph actually suggests that after 10^-5, more wait time means that serial is getting better. This isn't quite reflected in your analysis.

HW0/P6/P6.txt
```diff
@@ -1 +1,11 @@
 # Your answers here
+In each parallel run, the order to run the four function in parallel is random. For example, the order of 0,1,2,3 can be random, it could be 1,2,3,0
+For example, we have multiple steps to finish a calculation, and we also have several sub-steps in each steps. If the steps need to be finished step by step, and the sub-steps is independent of other sub-steps in each step. Then we can run sub-steps in parallel, while we mustr run steps in series but not in parallel.
+Otherwise, we would get total wrong results if we run steps in parallel.
+
+For P6B
+When the sleep time is very small compared to communicating time, then the total time depends on the communicating time. Serial computing is even faster than parallel computing, that might due to parallel computing spend lots of time on communicating since it has 4 processes, while serial computing does not.
```
vivjay30: If this is the case then why does 10^-5 give such a peak (meaning serial is much slower at 10^-5)?

HW0/P6/P6B.py
```diff
@@ -15,14 +16,30 @@ def burnTime(t):
     pool = mp.Pool(P)
 
     # Use a variety of wait times
-    ratio = []
-    wait_time = []
-
+    ratio =np.zeros(7)
```
vivjay30: Your graph is different from what I got, or what other people I checked with got. It might be because you are using numpy which has other performance complexities. In general the ratio should increase as wait time increases meaning parallel becomes faster as wait time increases. This is opposite of your results.

HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
...
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
-
-    # A = ?
-    # a = ?
+s0=np.matrix([[0],[0],[2],[15],[3.5],[4.0]])
+A=np.matrix([[1,0,0,dt,0,0],[0,1,0,0,dt,0],[0,0,1,0,0,dt],[0,0,0,1-c*dt,0,0],[0,0,0,0,1-c*dt,0],[0,0,0,0,0,1-c*dt]])
+a=np.matrix([[0],[0],[0],[0],[0],[g*dt]])
+s=np.zeros((120,6))
```
isadoranun: There is an easier way of creating A and a matrix that might help you in the future:
A = np.matrix(np.zeros([6,6]))
np.fill_diagonal(A, [1,1,1,1 - cdt,1 - cdt,1 - c*dt])

A[0,3] = dt
A[1,4] = dt
A[2,5] = dt

 a = np.matrix(np.zeros([6,1]))
a[5,0] = g*dt

HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
...
+def updateSig(sigma_bar):
+	C=np.matrix([[rx,0,0,0,0,0],[0,ry,0,0,0,0],[0,0,rz,0,0,0]])
+	sigma_kplus=(sigma_bar+(np.transpose(C))*C).I
+	return sigma_kplus
+
+def updateS(s_bar,sigma_bar,sigma_kplus,m):
+	C=np.matrix([[rx,0,0,0,0,0],[0,ry,0,0,0,0],[0,0,rz,0,0,0]])
+	mm=np.asmatrix(m)
+	s_kplus=sigma_kplus*((sigma_bar*s_bar)+(np.transpose(C))*(np.transpose(mm)))
+	return s_kplus
```
isadoranun: For all the functions, there is no need to redefine the matrices "A", "a", "B", "C", etc. if you already defined them before. Consider them as constant variables.

HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
...
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+# Show the plot
+ax.legend()
+plt.show()
 
-    # Show the plot
-    ax.legend()
-    plt.show()
```
isadoranun: I think you forgot to upload the plot :(

HW0/P5/P5.txt
isadoranun: I think you forgot to upload the plot P5.png :(

HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
...
+    # calculate each position
+s_temp=s0
+s[0]=np.transpose(s0)
+i=1
+while i < 120:
+	s_temp1=A*s_temp+a
+	s[i]=np.transpose(s_temp1)
+	s_temp=s_temp1
+	i+=1
+
```
isadoranun: I have two recommendations:
1- Instead of using 120 it is better to use for example len(read_temp_meas), in this way your code will be more robust.
2- Instead of using a "while" condition, you should make a "for" loop.  "while" is usually used when a command must be executed until a defined condition is met. "For" is used for repeating the same action a defined number of times. For example:

for i in xrange(120):
        s_temp1=A*s_temp+a
	s[i]=np.transpose(s_temp1)
	s_temp=s_temp1

HW0/P4/P4.py
```diff
@@ -29,49 +29,118 @@
...
+	s_bar=predictS(s_temp)
+	sigma_bar=predictSig(sigma_temp)
+	sigma_kplus=updateSig(sigma_bar)
+	s_kplus=updateS(s_bar,sigma_bar,sigma_kplus,read_temp_meas[i])
+	ss[i]=np.transpose(s_kplus)
+#update the temporary variable s and sigma
+	s_temp=s_kplus
+	sigma_temp=sigma_kplus
+	i+=1
+
```
isadoranun: Same here, "for" is the proper way of iterating this loop.

HW0/P5/P5.py
```diff
@@ -0,0 +1,35 @@
...
+print i
+x=range(0,n)
+plt.plot(x,s,'.b',label='unlimited cashiers')
+plt.plot(x,s_one,'.k',label='one cashier')
+plt.xlabel('number of bags')
+plt.ylabel('time(s)')
+legend=plt.legend(loc='upper left')
+plt.show()
+			
+
```
isadoranun: I think you forgot to upload the plot P5.png ...
But I ran your code and the resulting plot looks correct :+1:

HW0/P5/P5.txt
```diff
@@ -1 +1,12 @@
...
+
+2.it's faster for 8 cashier to do it. They can finish it in 31+log2(8)=34 seconds
+
+3.if the store has infinite employers, then it takes log2(256)=8 seconds
+
+5.it takes 15 seconds for infinite cashiers to count the coins.
+
+6.If cashier come and take 1 bag each time, then the total time would be 15+256=271 seconds
+  If cashier come and take 2 bags each time, then the total time would be 15+128=143 seconds
+
```
isadoranun: I got different answers for this question but I don't know what was your reasoning nor the right answer :(

HW0/P6/P6.txt
```diff
@@ -1 +1,11 @@
 # Your answers here
+In each parallel run, the order to run the four function in parallel is random. For example, the order of 0,1,2,3 can be random, it could be 1,2,3,0
+For example, we have multiple steps to finish a calculation, and we also have several sub-steps in each steps. If the steps need to be finished step by step, and the sub-steps is independent of other sub-steps in each step. Then we can run sub-steps in parallel, while we mustr run steps in series but not in parallel.
+Otherwise, we would get total wrong results if we run steps in parallel.
+
```
isadoranun: I believe that they do run in the same order they were submitted, but they take different times to be done and that is why you get the outputs in a different order.

HW0/P6/P6B.py
isadoranun: Yes, I obtained this same graph when running my code in the virtual machine. When I ran it in the Ipython notebook I got what everybody else has. I have no idea why....
I ran your code in an ipython notebook and the graph is ok :+1:  , I attach it here so you can see.
real_graph

HW0/P5/P5.txt
oyuchoi: How did you arrive at this conclusion? And also @vivjay30  is correct that a logarithmic solution is even faster!

PR #113, opened 2015-09-10
Hw0b

chuckyouliu:


HW0/P4/P4.py
```diff
@@ -40,37 +49,89 @@
...
+    B = np.matrix([[bx, 0., 0., 0., 0., 0.], 
+                   [0., by, 0., 0., 0., 0.],
+                   [0., 0., bz, 0., 0., 0.],
+                   [0., 0., 0., bvx, 0., 0.],
+                   [0., 0., 0., 0., bvy, 0.],
+                   [0., 0., 0., 0., 0., bvz]]) 
+    C = np.matrix([[rx, 0., 0., 0., 0., 0.], 
+                   [0., ry, 0., 0., 0., 0.],
+                   [0., 0., rz, 0., 0., 0.]]) 
+
```
Elena-Zhao: There might be a better way to construct matrix B and C by calling some functions provided by numpy:
B = np.diag((bx, by, bz, bvx, bvy, bvz))
C = np.hstack([np.diag((rx,ry, rz)), np.zeros((3,3))])

This looks more elegent than simply hardcoding :)

HW0/P5/P5.txt
```diff
@@ -1 +1,49 @@
...
+
+3) Using the same process of adding numbers, then taking half the employees to add the remaining numbers...
+	128 employees sum 2 numbers - 1 second
+	64 employees sum 2 numbers - 1 second
+	...
+	1 employee sums 2 numbers
+
+	For a total of 8 seconds
+
+4) ceil(log_2(N)) seconds.
```
Elena-Zhao: I agree the answer is ceil(log_2(N)) seconds, but this can be hardly seen on your corresponding plot. Why not narrow the range of x, or better, compute log for y axis

HW0/P5/P5_6.py
```diff
@@ -0,0 +1,82 @@
...
+                    header += " "
+            else:
+                for y in range(0, len(header) - len(bags)):
+                    bags += " "
+    print header + "\n" + bags + "\n\n"
+            
+
+if __name__ == '__main__':
+    calculateTime(256,1)
+    calculateTime(256,2)
```
Elena-Zhao: I must say this piece is really cool!

HW0/P6/P6.txt
```diff
@@ -1 +1,47 @@
...
+
+The Bye Job print statements are out of order. 
+If the called function is completely independent then there shouldn't be an issue - 
+but suppose the function called on an outside process that was used in the calculation
+of the function and then was updated for the next time the function was called (similar
+to the kalman filter problem), the ordering then would have an effect on the outcome.  
+If we want to do things in parallel in this situation we'd have to implement locks to 
+do the completely independent calculations in parallel but make sure the updating 
+procedures are done sequentially.
+
```
Elena-Zhao: Another issue:
What I occasionally find in output:
Hi Job 0
+Hi Job 1
+Hi Job 2
+Hi Job 3
+Bye Job 0Bye Job 2
+
+Bye Job 3Bye Job 1

It seems sometimes two processors output to the single io at the same time. This might cause deadlock due to shared resources.

HW0/P4/P4.py
```diff
@@ -40,37 +49,89 @@
...
+             '-k', label='Blind trajectory')
 
     #####################
     # Part 4:
     # Use the Kalman filter for prediction
     #####################
 
-    # B = ?
-    # C = ?
+    B = np.matrix([[bx, 0., 0., 0., 0., 0.],
```
sbalanovich: I like the way you did the matrices here, because it's very visually clear what they actually are. I personally used numpy to make these matrix declarations clean wherever possible (such as `B = np.diag([bx, by, bz, bvx, bvy, bvz])`), but I think that's more important when you have a lot of declarations like this, so for something like this problem set, I like this approach better.

HW0/P4/P4.py
sbalanovich: Completely missed this note when I wrote the above comment, but like I said, I think hardcoding in this case is fine because it shows you at a glance what the matrix actually looks like.

HW0/P5/P5.txt
sbalanovich: Agreed with this. Also it looks like your plot above is smooth, which indicates that you may not have actually plotted it as a `ceil()` function.

HW0/P5/P5_6.py
```diff
@@ -0,0 +1,82 @@
+# -*- coding: utf-8 -*-
+"""
```
sbalanovich: Was this necessary? I applaud your dedication and commitment, but doesn't this seem redundant given that we already have a mathematical solution derived in P5? Maybe I'm misreading this, but couldn't all of this be expressed with the one-liner `ceil(log_2(N))`?

HW0/P5/P5.txt
chuckyouliu: I think I did, just plotted it in the console didn't write a script so can't really remember

HW0/P5/P5_6.py
chuckyouliu: This was only to solve the part where you hand out 1 or 2 bags per second, so at the very least the solution requires N or N/2 seconds to just hand out the bags. The comment at the top is misleading, it wouldn't be optimal if you were handing out say 64 bags per second because you would then prefer a more tree-like approach that gets to log_2(N)

sbalanovich: Ah, I see, got it. Well done then! :+1:

HW0/P4/P4.py
zezhouliu: Yes, np.diag is a cleaner implementation

PR #115, opened 2015-09-10
Hw0b

kevinkeyjkw:


HW0/P6/P6.txt
```diff
@@ -1 +1,26 @@
...
+
+How could this affect how we program in parallel?
+This affects the way we program because many problems will occur if our processes share a resource.
+
+Describe a scenario where this would be important.
+For example, if a two processes want to add 1 to a shared variable x only if x==0. Process A checks and sees that x==0, but is interrupted by Process B. B checks x==0 is true and adds one to x. A continues to execute without realizing that B already changed the value of x and thus adds another 1 to x. Now x's value is 2. This could have serious consequences if not noticed.
+
+Try to explain the trend you observe. Is it possible that a parallel program could take longer than it’s serial version? Under what conditions?
+From the results and graph provided, we see that serial execution of burntime takes more time than parallel execution. I think this is due to multiple processes calling burnTime in parallel, thus there is no need to wait for one process to finish before processing the next one. Parallel programs could take longer than serial if there is a shared resource. Communication between processes for that resource could result in a longer execution time.
```
jonathanhfriedman: For small wait times, the serial code was faster, so a parallel program could also be slower when the time required to execute the process is not much larger than the time lost to communication overhead.

jpbm: Is it really the scheduler that determines how long a job is allowed to take? I would have expected that the job just runs as long as it runs...

PR #117, opened 2015-09-10
Hw0b

btweinstein:
Submitting my homework.


HW0/P4/P4.py
```diff
@@ -30,48 +35,142 @@
...
     #
     # Read the observation array and plot it (Part 2)
     #####################
+    measurements = pd.read_csv('./P4_measurements.txt', header=None).values # Keep it as a ndarray
+    # We now have to rescale x, y, and z appropriately. Luckily, the matrix
+    # is diagonal so we can just decouple everything.
+    rescaled_measurements = measurements.copy() #Copy...we don't want to overwrite the original measurements
+    rescaled_measurements[:, 0] *= 1./rx
+    rescaled_measurements[:, 1] *= 1./ry
+    rescaled_measurements[:, 2] *= 1./rz
```
jack91an: Good observation! 
Alternatively, you could construct the matrix in case the distortion is more complex.

HW0/P4/P4.py
```diff
@@ -30,48 +35,142 @@
...
-    # a = ?
-    # s = ?
+    A = np.array([
+        [1, 0, 0, dt, 0, 0],
+        [0, 1, 0, 0, dt, 0],
+        [0, 0, 1, 0, 0, dt],
+        [0, 0, 0, 1-c*dt, 0, 0],
+        [0, 0, 0, 0, 1-c*dt, 0],
+        [0, 0, 0, 0, 0, 1-c*dt]
+    ], dtype=np.double)
```
jack91an: Instead of hard coding, you could also construct two arrays and use the np.diag() function to insert the arrays into a zero matrix. This should help if we are recording more parameters and have a huge A matrix. (This comment also applies to later matrix definitions, but I think it's a trivial point)

HW0/P4/P4.py
```diff
@@ -30,48 +35,142 @@
...
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    ax.plot(x_filtered, y_filtered, z_filtered,
+             '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
-    plt.show()
+    plt.show()
```
jack91an: Very clear variable naming, and appreciate the comments.

HW0/P5/P5.txt
```diff
@@ -1 +1,133 @@
...
+ t_minimum = [(N/zeta - 1) + Log2(zeta)]t_add
+
+
+The two limits we are interested in yield
+
+t_one = (N - 1)t_add
+t_infinite = Log2[N]t_add
+
+See P5_part4.py for the python program to produce the plot. See P5.png
+for the plot.
```
jack91an: In my opinion, the time should be a step function instead of a continuous one, because between (2^n, 2^(n+1)] bags, the time spent by infinite number of workers to reduce the number of bags to the 2^n would be the same (i.e. 1 second). I would add a ceil() to your function. Let me know if you disagree :)

This applies to P5.4 as well. In P4.4, we got lucky because the number of bags is a perfect 2^8.

HW0/P5/P5.txt
```diff
@@ -1 +1,133 @@
...
+there is a huge bottleneck! The employee will *at least* have to wait
+
+t_wait = N*t_hand
+
+before having all of the bags to total. After receiving the
+final bag, the employee will have to add that one. Therefore,
+
+t_one_bag_at_a_time = t_wait + 1 = N t_hand + 1
+t_one_bag_at_a_time = 257 seconds
+
```
jack91an: What about the 255th bag? If you hand to the same person as the last bag, then s/he will need to somehow sum the previous sums with the 255th bag, and then the 256th bag, which takes 1 additional second to 257s. If you hand it to a different person, then s/he'll need to add+communicate, which also takes more than 1 second before the last person could do his/her calculation. So I think the theoretical minimum is 258s- am I missing anything?

HW0/P6/P6.txt
```diff
@@ -1 +1,46 @@
...
+It may be hard to parallelize some code due to these issues.
+
+One scenario where process timing would be important is having multiple
+threads reading from a set of data that another thread is producing.
+If timing is not carefully controlled, the data-analyzing
+threads may try to read a dataset that does not yet exist, causing problems!
+Furthermore, if the data-analyzing threads somehow hogged the computational resources,
+the data-producing thread may be unable to function, rendering the data-analysis
+threads useless. This problem could deadlock the entire program!
+
```
jack91an: Good point! This reminds me of the Kalman Filtering method in P3, whereas the later state depends on adjustment input from the previous state.

HW0/P4/P4.py
btweinstein: Thanks!

HW0/P5/P5.txt
btweinstein: Agreed! I decided not to put it in for simplicity, but it definitely should be there.

HW0/P5/P5.txt
btweinstein: I think it depends if someone can be handed a bag and can add at the same time. That's what I assumed. If they can't add & grab a bag at the same time, I think you're right.

HW0/P6/P6.txt
btweinstein: Agreed. Splitting that program into multiple threads could definitely be a disaster if not managed correctly!

btweinstein:
Thanks for your comments!

HW0/P6/P6B.py
```diff
@@ -33,4 +55,6 @@ def burnTime(t):
     plt.xlabel('Wait Time (sec)')
     plt.ylabel('Serial Time (sec) / Parallel Time (sec)')
     plt.title('Speedup versus function time')
+    plt.savefig('P6.png', dpi=200, bbox_inches='tight')
     plt.show()
+
```
ymiyamot: I had no useful tips to add for the code, it looks great. In general, I appreciated your detailed explanations in your text answers of how you came to your conclusions.

HW0/P5/P5.txt
```diff
@@ -1 +1,133 @@
...
+only N/q employees will have bags at the top of the tree. Therefore,
+
+t_q_bags_at_a_time = t_wait + t_sum_down_tree
+t_q_bags_at_a_time = (N t_hand)/q + Log2[N/q](t_add + t_communicate)
+
+Plugging in N=256, q=2, we find
+
+t_total = 142 seconds
+
+c) The most efficient strategy is to distribute bags as fast as
```
oyuchoi: Good!

PR #120, opened 2015-09-10
HW0b Submission

apetschek:
This is my submission of HW0b :)


HW0/P6/P6B.py
```diff
@@ -11,21 +12,27 @@ def burnTime(t):
...
+    #wait_time = np.array([10**-6,10**-5,10**-4,0.5*10**-2,0.5*10**-1,10**-1,1])
+    wait_time = np.array([10**-6,10**-5,10**-4,10**-3,10**-2,10**-1,1])
     for t in wait_time:
         # Compute jobs serially and in parallel
-        # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
-
+        begin = time.time()
+        pool.map(burnTime(t),range(N))
```
kevinkeyjkw: Hi Andrew, I saw this line and am not sure what it is doing. I may be wrong but shouldn't the second parameter be the parameters passed to burnTime. Say for wait time of 1 you would pass in pool.map(burnTime,[1]*16) and it would execute burnTime(1) sixteen times. But since your graph looks correct I am curious to see whether this is another way of using map().

apetschek: thanks, kevin for the note! that may be correct. i think the shape of my curve is right...but i couldn't get the graph to approach 4 (instead of 16 as it is currently doing). I just gave pool.map(burnTime,[1]*16 a try and that seemed to have worked! though i'm not entirely sure of now what the code means..why don't we have to put burnTime(t) anymore, and instead just use burnTime? and what does the brackets around the 1 signify here? :) this was a tricky one for me.

kevinkeyjkw: No problem Andrew. The [1]*16 is just means [1,1,1,1,1,1,1,1,1,1,1,1,1,1,1]. Then map will pass in each value in the array(in this case will pass 1 each time) to burnTime. Thus executing burnTime(1) 16 times

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
...
+
+To improve this process, i would try to minimize the bottleneck that exists here where i am the only distributor of bags. As you can see from my descriptions above, there are times where workers are idle because there is no one yet to interact with (i.e., not everyone has bags). Thus, to speed up this process i could hand 128 bags to the first worker in the 1st second and in the 2nd second i could hand 64 bags to a third worker while the 2nd worker could give 64 of his bags to a 4th worker, and in the 3rd second i could give 32 bags to a 5th worker, worker #1 could give 32 of his bags a 6th worker etc … This would continue until there are 128 workers with 2 bags each. Then 15 seconds of counting would commence. The time this would take would be 7 seconds of bag distribution + 15 seconds of counting among the 128 workers = 22 seconds.
```
zxjyjr: That's a very good strategy! 22 seconds is impressive

HW0/P6/P6.txt
```diff
@@ -1 +1,12 @@
...
+
+6B. For very small sleep times (e.g., 10^-6, 10^-5), we see the ratio between serial time and parallel time to be less than 1, meaning that it actually takes longer for parallel processes to complete the loops than the serial process. However, serial's advantage begins to give way (exponentially) to the parallel processes being faster as wait time increases (e.g., 10^-4, 10^-3). The growth in advantage begins to slow however and the ratio approaches a value of 16 as time gets larger and more substantial and closer to wait time of 1 second (e.g., 10^-1, 10^0). I initially expected to see the graph approach a value of 4, as that is the number of processes we have created and 16 is just an arbitrary number of loops.
```
zxjyjr: I would think it could also be that the parallel strategy cost more time to communicate than serial strategy for very small sleep time, so the parallel strategy becomes more time consuming than serial strategy.

HW0/P4/P4.py
```diff
@@ -40,38 +41,68 @@
...
+                         
+    def predictSig(A,Sigma,B): #to be called Sig_predicted in loop or Sigk
+        return inv(np.add(np.dot(np.dot(A,Sigma),A.transpose()),np.dot(B,B.transpose())))
+        
+    def updateSig(x,y): #to be called Sig_updated in loop or Sig K+1
+        return inv(np.add(x,np.dot(y.transpose(),y))) 
+    
+    def updateS(x,y,z,i,j): #to be called s_new in loop or  Sk+1
+        return np.dot(x,np.add(np.dot(y,z),np.dot(i.transpose(),j))) 
+
```
clim16: If you want to avoid writing out the operations like .add() and .transpose() to make it clearer and to use less parentheses you can convert to matrices and use the mathematical symbols like + and *.  You can also transpose using .T and invert using .I

HW0/P5/P5.txt
clim16: you have thorough explanations!

PR #121, opened 2015-09-10
Hw0b

jonathandlee12:


HW0/P4/P4.py
```diff
@@ -30,47 +30,101 @@
...
-    # s = ?
-
-    # Initial conditions for s0
-    # Compute the rest of sk using Eq (1)
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-k', label='Blind trajectory')
+    
+    # initialize A matrix and a vector
+    A = np.matrix([[1,0,0,dt,0,0],[0,1,0,0,dt,0],[0,0,1,0,0,dt],[0,0,0,1-c*dt,0,0],[0,0,0,0,1-c*dt,0],[0,0,0,0,0,1-c*dt]])
```
xdai1091: I did the same thing. But my friend said this is actually not a very good style of coding

HW0/P5/P5.py
```diff
@@ -0,0 +1,15 @@
+import numpy as np
+import matplotlib.pyplot as plt
+
+if __name__ == '__main__':
+    x = np.arange(1,500)
+    y1 = x - 1
+    y2 = np.ceil(np.log2(x))
```
xdai1091: I didn't use "ceil" function here. Do you mind elaborate your thought here for me?

HW0/P5/P5.txt
```diff
@@ -1 +1,23 @@
...
+4. With a countably "infinite" number of employees, the time required would be, in seconds, a ceiling of log base 2 of the number of bags, whereas a single employee would require N - 1 seconds.
+5. Going back to #3, neglecting communication delay, it requires 8 s to complete this task. Now, with communication delay, each round of computation requires a round of cross-talk between cashiers, which means that there is a 1 s delay between each of the 8 stages. Therefore, there is an additional 7 s delay on top of the 8 s = 15 s.
+6. It will take 256 s to distribute 1 bag to 1 cashier, so let's look at the terminating stages: once all 256 cashiers have received 1 bag, the last two need to (1) talk, (2) sum, (3) communicate with the rest of the "sum", and (4) sum for the final value. Hence, 256 + 4 s = 260 s.
```
xdai1091: Is it just 4 s?

HW0/P5/P5.txt
```diff
@@ -1 +1,23 @@
...
+
+It will take 128 s to distribute 2 bags to each cashier. Each cashier would do a summation, then communicate with whatever cashier is "available" and sum the available numbers. At the end of the distribution, there will be a leftover sum at this point, so we need to communicate and then add the two values to give the total value, thus requiring 128 s + 5 s = 133 s.
+
+Consider the following cases, where I hand off 256 bags to one cashier -- I cut down significantly on the transfer time (1 s), but the serial processing is going to take time (255 s), for a total of 256 s. Given our answers to the cases of lower numbers, there is clearly at least one optimal value of bags for me to be able to hand off so as to minimize total processing time. 
+
```
xdai1091: I thought the same way before, then I feel like the point here is let them to finish all work at the same time. With that's been said, the first person who gets the bags could start counting immediately. Does this make sense?

HW0/P6/P6.txt
```diff
@@ -1 +1,86 @@
...
+Hi Job 8
+Bye Job 5
+Hi Job 9
+Bye Job 6
+Bye Job 7 ### <0.25 s pause>
+Bye Job 8
+Bye Job 9
+[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+What we get in practice is not the expected ordering, although we obtain our "sorted" ouput in the end. This may be a consequence of differences in latency, compute time, and bandwidth between processors.
```
xdai1091: quote from student's user,patrickday9, "What happens in practice is that the program actually wakes new processes before closing the previous ones. This is worrisome when programming in parallel as we must be concerned about data dependancies."

jdlkem: I agree with both statements independently, but I don't understand how either is applicable in this case.

HW0/P5/P5.txt
jdlkem: My line 21 is just to give an upper bound on the efficiency of the problem.

HW0/P5/P5.py
jdlkem: The np.ceil function is used here because you can't evenly divide 256 bags across 3 cashiers (it would be most efficient to use numbers of cashiers that are exponents of 2).

HW0/P5/P5.txt
jdlkem: Yes, because the time it takes for two cashiers with one bag each (e.g., beginning after 2 seconds)  to sum (1 s to communicate, 1 s to add) is the same time it takes for you to give the two bags (2 s), one to each cashier. Thus, by the time the 256th bag is given and the 255th and 256th cashiers have communicated their bags and the 256th cashier has summed, the 254th cashier will already be poised with the cumulative sum from all other 254 cashiers (thus, only 2 s more = 2 + 2 + 256 = 260 s). Try drawing it out if you aren't convinced (start with 8 bags).

HW0/P4/P4.py
jdlkem: Yeah, I thought about changing it to something nicer (e.g., np.diag(np.array([])) and then concatenate), but I got lazy =P

zezhouliu: Yes, @jdlkem , np.diag is a good method to use in this case!

HW0/P6/P6.txt
```diff
@@ -1 +1,86 @@
-# Your answers here
+Problem 6.
```
zezhouliu: Your graph for problem 6 looks like it has a lot of noise and doesn't give much insight into the trend.

jdlkem: @zezhouliu is there something in my code that could be causing the noise (I was using VirtualBox)? Given the fluctuations, I could have easily data snooped and chosen a graph that I would expect would be most representative of the trend (indeed, I had several graphs that look similar to those of my peers), but that would be unscientific and misleading. This is why I chose to implement error bars and repetitions here. 

The distribution of standard errors itself is very indicative of a trend in performance. Narrow error bars at 1e-6 s and above 1e-2.5 s wait times are indicative of clear performance differences between the parallel and serial methods, and there is an obvious transition zone between these two values. The wide error bars in between these two wait times may suggest that the speedup values at these wait times are sensitive to fluctuations in bandwidth or other components inherent to my computer or other concurrent processes (if not my code, possibly due to VirtualBox, my computer, etc., although I made it a point to kill as many irrelevant running tasks on my computer prior to running this). The best way to better determine the exact interval in which the transition occurs would have been to increase the number of reps of my run.

PR #125, opened 2015-09-11
Hw0b

zelong430:


HW0/P4/P4.py
```diff
@@ -30,33 +30,60 @@
     # Load true trajectory and plot it
     # Normally, this data wouldn't be available in the real world
     #####################
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '--b', label='True trajectory')
+    # load txt file
+    s_true = np.loadtxt("P4_trajectory.txt", delimiter = ",")
+    ax.plot(s_true[:,0],s_true[:,1] , s_true[:,2] ,'--b', label='True trajectory')
```
samighoche: nice use of python syntax:)

HW0/P4/P4.py
```diff
@@ -30,33 +30,60 @@
...
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    s_measure = np.loadtxt("P4_measurements.txt", delimiter = ",")
+    # construct the streching matrix
+    s_streching = np.matrix([[1/rx,0,0], [0,1/ry,0],[0,0,1/rz]])
+    # caliberate measurement by s_streching * s_measure
+    s_caliberate = []
+    for x in range(len(s_measure[:,0])):
+	s_caliberate.append(np.dot(s_streching,s_measure[x].T))
+    result = np.asarray(s_caliberate)
```
samighoche: Another option would be to multiply the s_measure matrix by the s_stretching matrix instead of appending results of matrix * vector multiplications because python for loops are slower than numpy matrix operations (which are actually done in C)

HW0/P4/P4.py
```diff
@@ -65,13 +92,66 @@
...
-
+    # initialize matrix B and C
+    b = np.zeros([6,6])
+    B = np.asmatrix(b)
+    B[0,0] = bx
+    B[1,1] = by
+    B[2,2] = bz
+    B[3,3] = bvx
+    B[4,4] = bvy
+    B[5,5] = bvz
```
samighoche: Nice initialization!

HW0/P5/P5.txt
```diff
@@ -1 +1,24 @@
...
+	8 people in total
+	First, each person count 32 cases, take them 31 ms, result in 8 new numbers
+	Second, 4 people each sum 2 of the 8 numbers, take them 1 m, result in 4 new numbers
+	Thirdly, 2 people each sum 2 of the 4 numbers, take them 1 m, result in 2 new numbers
+	Finally, sum the 2 numbers, take them 1m
+	In total, they spend 34 ms
+
+	They can count all cases in 34ms.
+3.8ms
+4.log2(n)
```
samighoche: I think this would be the ceiling of log2(n), because in case n is not a power of 2, there will be one last operation that will take a second

HW0/P5/P5.txt
```diff
@@ -1 +1,24 @@
...
+5.8 + 7 = 15
+6. if one person take one bag at a time: it tooks 8 + 8 + 256 = 272 ms
+   if one person take two bags at a time: 15 + 128 = 143 ms
```
samighoche: I agree with both those times:) but if you allow cashiers to perform additions and communications at the same time that the rest of the bags are being distributed, then you can add more parallelism and cut down on the time cost

HW0/P6/P6.txt
```diff
@@ -1 +1,17 @@
 # Your answers here
+P6A:
+	The output of print Jobs has different order each time the script run.
+	After several times of running, I found the order of each sentence appearing is random.
+	I think the randomness happened because there is only one input flow for the I/O. 
+	Since the burnTime function is run on 4 processor simultaneous, the processors turn the data into I/O at the same time.
+	Then the I/O randomly pick one statement to print out.
```
samighoche: I don't fully agree with this but I might be wrong. I think that a processor might take longer on the same job than another processor, for many reasons like background processes happening, the processor being busy, etc. But I don't think that the processors finish at the same time and then the I/O selects randomly

HW0/P6/P6.txt
```diff
@@ -1 +1,17 @@
...
+	We need to make sure that in parallel programming, given specific instructions to handle the situations that severals happened at same time.
+	This would be important if we want to print out result of calculation which are solved parallelly and finish at same time.
+	In this situation, we may want the result to be printed at a specific order instead of random order.
+
+P6B:
+	When a task take relative short to complete, in our case when the sleep time is 10^-6 s, complete the task in serial take same time as complete the task in parallel.
```
samighoche: I agree with most of this:) However I think that for very short jobs, it's even faster to run serially because multi-processing has overhead that can outweigh the benefit of parallelism for jobs that are not very demanding

HW0/P6/P6B.py
```diff
@@ -16,17 +16,27 @@ def burnTime(t):
 
     # Use a variety of wait times
     ratio = []
-    wait_time = []
+    wait_time = [10**-6, 10**-5,10**-4,10**-3,10**-2,10**-1,1]
```
samighoche: I think you might want to have more data points, here is a good way to generate a lot of points (50 by default) 
np.logspace(-6, 0), and you can specify how many points you want as a third argument: np.logspace(-6, 0, 20) for 20 points

HW0/P4/P4.py
```diff
@@ -30,33 +30,60 @@
...
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
     #####################
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    s_measure = np.loadtxt("P4_measurements.txt", delimiter = ",")
+    # construct the streching matrix
+    s_streching = np.matrix([[1/rx,0,0], [0,1/ry,0],[0,0,1/rz]])
```
jdlkem: Another option would be to use np.diag(np.array()), rather than write out all those zeroes.

HW0/P5/P5.txt
jdlkem: Ditto here, although the graph seems to suggest the ceiling interpretation. I would have liked to see the code for confirmation of understanding.

HW0/P5/P5.txt
jdlkem: Yes, I agree with samighoche as well. All of the other numerical answers I agree with as well.

HW0/P6/P6.txt
jdlkem: Once again, I agree with samighoche. Think of it this way: parallel jobs can process in parallel (lol redundancy), but a serial process must first occur to initialize the parallelization. My understanding is that each job theoretically has an expected order of completion, assuming perfect conditions across processors and equal communication distances/latency/etc; however, there are other processes, physical phenomena, and other noise that may slow down an individual processor, thereby leading to seemingly random outputs.

HW0/P6/P6.txt
```diff
@@ -1 +1,17 @@
 # Your answers here
+P6A:
+	The output of print Jobs has different order each time the script run.
+	After several times of running, I found the order of each sentence appearing is random.
+	I think the randomness happened because there is only one input flow for the I/O. 
+	Since the burnTime function is run on 4 processor simultaneous, the processors turn the data into I/O at the same time.
+	Then the I/O randomly pick one statement to print out.
+	We need to make sure that in parallel programming, given specific instructions to handle the situations that severals happened at same time.
+	This would be important if we want to print out result of calculation which are solved parallelly and finish at same time.
+	In this situation, we may want the result to be printed at a specific order instead of random order.
```
jdlkem: Definitely true.

HW0/P6/P6.txt
```diff
@@ -1 +1,17 @@
...
+
+P6B:
+	When a task take relative short to complete, in our case when the sleep time is 10^-6 s, complete the task in serial take same time as complete the task in parallel.
+	As the task take longer time to complete, in our case when the slepp time increase to 10^-3 or 10^0 s, completing the task in parallel become more effienct than completing in serial.
+	However, since we only have 4 processors, the best we can do in parallel computing in this case is complete the tasks in 4 times effiency.
```
jdlkem: Compared to the overhead of parallelism, I agree with this.

HW0/P6/P6B.py
```diff
@@ -16,17 +16,27 @@ def burnTime(t):
...
+	    burnTime(t)
+	ts_end = time.time()
 
-        # Compute the ratio of these times
-        # ratio.append(serialTime/parallelTime)
+        serialTime = ts_end - ts_start
+
+	tp_start = time.time()
+	result = pool.map(burnTime, [t]*16)	
+	tp_end = time.time()
```
jdlkem: I'm not sure if this is relevant, but the times of initialization of the range(16) list vs. [t]*16 may be different and could potentially impact the results. I would look into this to check if these are different, or this could be avoided better by initializing this before running time.time().

samighoche: Good point:)

PR #126, opened 2015-09-11
HW0 Part 2

newtestsubject:


HW0/P4/P4.py
```diff
@@ -29,48 +29,123 @@
...
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
-    #####################
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    s_measure = np.loadtxt('P4_measurements.txt', delimiter=',')
+    
+    V = np.matrix([[1/rx, 0, 0], [0, 1/ry, 0], [0, 0, 1/rz]])
```
jdlkem: Another option could be np.diag(np.array()), so you don't have to type all those 0's :)

HW0/P5/P5.txt
```diff
@@ -1 +1,29 @@
...
+      - Each of 128 cashiers gets 2 bags and adds them up. It takes 1 second.
+      - There are now 64 bags. 31 cashiers add 2 bags each. It takes another 1 second.
+                                       .
+                                       .
+                                       .
+                                2^x = 256 
+                                    x = 8
+                          It takes 8 seconds.
+
+4.) I takes logN seconds for >= N/2 cashiers to add up N bags. It takes N - 1 seconds for one cashier to add up N bags. See plot.
```
jdlkem: I think there should be a ceiling function involved here.

HW0/P5/P5.txt
```diff
@@ -1 +1,29 @@
...
+                                    x = 8
+                          It takes 8 seconds.
+
+4.) I takes logN seconds for >= N/2 cashiers to add up N bags. It takes N - 1 seconds for one cashier to add up N bags. See plot. 
+
+5.) Each step when two cashiers add up their bags takes 2 seconds instead of 1 second. For 256 it takes 2*8 = 15 seconds (for >= N/2 cashiers).
+
+6.) First 256 bags are handled over to 256 cashiers. This takes 256 seconds. The next 8 consecutive steps require 1, 2, 4, 8, 16, 32, 64 and 128 additional seconds for handing over the bags. This adds up to 511 additional seconds (+ 16 seconds for counting and communication).
+
+If 2 bags are handled to each cashier at the beginning the handling takes 510 seconds (+ 15 seconds for counting and communication), since the first step (128 cashiers give one bag each to 128 other cashiers = 1 second) no longer happens.
```
jdlkem: Hypothetically, handing the bags takes 256 seconds, but once the bags have been handed, only numbers need to be communicated (i.e., no bag hand-offs need to occur at this point). Thus, as an upper bound, it should take 256 + 15 (from part 5) (and I think it can be even lower with concomitant parallelism).

HW0/P6/P6.txt
```diff
@@ -1 +1,10 @@
...
+
+A non-parallelized program can be thought of a linear (and branched) chain of events. Parallelization disturbs this linearity. This can cause problems when outputs of different steps depend on each other. For example when calculating a + b = c, a and b can be calculated in parallel, but before c can be calculated the results of the both previous calculations must be available. Thus the processes that calculate and b need to communicate with one another.
+
+2.) The serial version of the program is faster when the waiting times are short. The parallel version is faster when the waiting times are long. Parallel execution requires the processor to coordinate the processes that are running in parallel. This takes time and resources. When the waiting times are short this overhead slows the execution down. When the waiting times are long, the processor has enough time to coordinate the execution of the instances that are running in parallel. This leads to a speed up.
```
jdlkem: Yes, I agree broadly. I think the better phrasing to use would be that "relative to the parallelization time overhead," serial processing is slower than parallelization for long wait times.

HW0/P6/P6B.py
```diff
@@ -18,14 +18,31 @@ def burnTime(t):
...
+        
+        now1 = time.time()
+        resultParallelTime = pool.map(burnTime, [t]*N)
+        parallelTime = time.time() - now1
+        #print parallelTime
+        
+        now2 = time.time()
+        resultParallelTime = map(burnTime, [t]*N)
+        serialTime = time.time() - now2
+        #print serialTime
```
jdlkem: I like the use of map for both. Also, good use of many steps.

HW0/P5/P5.txt
tzhou22: As jdlkem said, for #4 we would expect the graph to be shaped like stair steps and not a smooth line (made the same mistake).
Expanding on above, you can think about the part where each cashier has one bag as some sort of base case, and once we reach that point, the final solution can be reached in the 15 seconds calculated in #4.

HW0/P5/P5.txt
tzhou22: ^^

PR #127, opened 2015-09-11
work on P4.py for HW0b

xiaweiwang:


HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
 # Your answers here.
+1. It takes 255 seconds.
+
+2. Yes, working together makes the verification faster. 
+Each cashier counts the sum of 32 bags using 31 seconds and adding the subtotals up takes 3 seconds, so it takes 34 seconds in total.
+
+3. 2^8 = 256. It takes 8 seconds.
+
+4. See plot. Since the time need to be an integer in seconds, it is [ln(N)/ln(2)]+1, where [] refers to the integer part.
```
rohitahuja: (Adding my comment here since I can't directly comment on the graph.) To graph the difference between the n and log_2(n), I made the y axis logarithmic. This would make the difference between the two clear for very large n.

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
...
+
+5. If it takes a second to communicate, then it takes an addtional second to enter the next round of counting. So it takes 8s (counting time) + 7s (communication time)=15s.
+
+6. 1) If each cashier takes 1 bag, the distribution time will be 256s, communication time will be 8s (8 rounds of communication) and counting time 8s (8 rounds). The total is 272s.
+
+2) If each cashier takes 2 bag, the distribution time will be 128s, plus what we calculated in the previous question, the total time will be 128s+15s=143s.
```
rohitahuja: I claimed it would take 133 seconds. This is because I drew out the cashiers and indexed them. Basically, every time a cashier gets a bag, he will spend an extra second summing it. We only need 128 cashiers since each cashier gets 2 bags. In second 2, cashier 1 receives his bags. Every second beyond that, we will have a cashier summed his bags. In seconds 4 and 5, cashier 1 and cashier 2 will communicate and sum their bags. From here, whenever a cashier sums the bags he received, he will sum with one of the two sets of people who just finished summing their bags. cashier ontinuing with the example, after second 5 (when cashier 4 finished summing his bags), cashier 1 will have the sum of his bags and cashier 2's bags, and sum that result with cashier 4's result. This will continue with the rest of the even indexed cashiers. This follows for odd indexed cashiers as well. By the end, we have cashier 128 sum his bags at the 129th second, and he communicates and sums that with two cashiers holding resulting sums. This process ends at 133 seconds.

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
...
+
+A better strategy would be to distribute bags from top down, i.e., we distribute 128 bags to the first two cashiers and they each distribute bags evenly to two cashiers and so on. In this case, the total distribution time will be 8 seconds. As we calculated in Q5, the communication + counting time is 15 seconds. So the total time is 23 seconds.
```
oyuchoi: This is a great logarithmic solution (but you can do a bit faster!)

PR #128, opened 2015-09-11
Hw0b

vinayps:
Submitting HW0b


HW0/P5/P5.txt
```diff
@@ -1 +1,11 @@
 # Your answers here.
+1) 255 seconds
```
oyuchoi: Your P5 graph makes it look like both lines are linear!

HW0/P5/P5.txt
```diff
@@ -1 +1,11 @@
...
+With two bags being handed out at the same time, the lower bound on number of total operations comes down to 128 initial hand overs + 255 additions = 383 operations. Irrespective of how many of those additions/communications/handovers can happen in parallel, it will still take more than 128s to get the total. But this is much better than the earlier case with "one bag handovers".
+Thus clearly, improving the I/O bandwidth is the way to bring down the processing time rather than throw more processors at the problem (we are already at infinite!) The better strategy would be to fix the initial handover bottleneck or in other words try to reduce the I/O bottleneck time by increasing the bandwidth by say placing the bags on a large table from which 32 cashiers could pick two bags each in 1s. That would bring down the total processing time to 17s, about as good as we can get.
```
oyuchoi: This changes the requirements that each agent (either yourself or a cashier) can only communicate with one person at a time.  It would be faster if multiple cashiers could take money from you at the same time, but sometimes this is not possible (as in our case), and instead you have to factor in the communication cost.

PR #136, opened 2015-09-11
Hw0b - Virgile Audi

virgodi:


HW0/P6/P6B.py
```diff
@@ -1,36 +1,43 @@
...
+	for t in wait_time:
+    	# Compute jobs serially and in parallel
+		# Use time.time() to compute the elapsed time for each
+		before = time.time()
+		for i in range(16):
+			burnTime(t)
+		serialTime = time.time()-before
+		
+		before = time.time()
+		pool.map(burnTime(t), range(N))
```
apetschek: I had the same code written on my assignment as well, but the graph it produced, like yours, approached a ratio of 16 instead of 4 which i believe should the right answer. I've been informed (through code peer review) that the ratio of 16 (and not 4) resulted from how we called pool.map(burnTime(t), range(N))...

instead we should have said pool.map(burnTime,[1]*16) which passes in an array of 16 ones: [1,1,1,1,1 etc...] That way it passes in burnTime(1) 16 times to align with what serial process is doing.

hope that helps! i'm looking into it further.

README
```diff
@@ -0,0 +1 @@
+My name is Virgile Audi
```
apetschek: this part looks correct :)

HW0/P6/P6B.py
virgodi: Thanks for the explanation ! This is all very new to me...

HW0/P4/P4.py
```diff
@@ -30,48 +30,105 @@
...
+ax.plot(s_true[0,:], s_true[1,:], s_true[2,:],
+             '--b', label='True trajectory')
 
     #####################
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
     #####################
+m=np.loadtxt("P4_measurements.txt", delimiter=",",unpack=True)
+notc=np.diag([1/rx,1/ry,1/rz])
```
aidizhang: I like the variable name notc :)

HW0/P4/P4.py
```diff
@@ -30,48 +30,105 @@
...
+s = np.array([0.,0.,2.,15.,3.5,4.])
+
+S=np.zeros(shape=(6,K))
+S[:,0]=s
+S=np.asmatrix(S)
     # Initial conditions for s0
     # Compute the rest of sk using Eq (1)
+for i in range(0,(K-1)):
+	S[:,(i+1)]=np.asarray((np.asmatrix(A)*np.asmatrix(S[:,i]))+np.asmatrix(a).T)
+S2=np.asarray(S)
```
aidizhang: Cool! Perhaps setting up the variables as matrices, conducting all operations on them, and changing final matrix to an array would be cleaner code, but I'm nitpicking here.

HW0/P4/P4.py
```diff
@@ -30,48 +30,105 @@
...
-    #         '-r', label='Filtered trajectory')
+Skalman2=np.asarray(Skalman)
+ax.plot(Skalman2[0], Skalman2[1], Skalman2[2],
+            '-r', label='Filtered trajectory')  
 
     # Show the plot
-    ax.legend()
-    plt.show()
+ax.legend()
+plt.show()
```
aidizhang: This is in reference to P5.png below. Perhaps changing y-axis to log scale would have better visualization effects :)

HW0/P5/P5.txt
```diff
@@ -1 +1,61 @@
...
+ 3      5      7      8    11      12       15      17       here it only takes 1s as no communication is needed
+     7      						Now we need 2 sec as we need communication in between cashiers
+          9        
+                 11  
+                       13
+                              15
+                                     17
+                                             19
+
+So we spot a pattern that would lead to an added 251s for 256 bags for a total of 259 seconds. And this is without adding time for bringing the bags from one cashier to another. If cashiers were allowed to take 2 bag at a time, then we would reduce the time.
```
aidizhang: This is great! The diagram you drew definitely helps. I got 260 seconds, which is 1 second slower than yours I guess :( I guess the question is how you would definitively prove that having each cashier going twice in a row to pick up a bag is the most optimal, other than gut feel/heuristics. And also whether we can generalize your pattern to allowing cashiers to take multiple bags at a time. Cool solution though!

README
aidizhang: :) extra points

HW0/P6/P6B.py
aidizhang: Yep! I called pool.map(burnTime, [t]*N) and it approached a 4x parallelism factor. Why does pool.map(burnTime(t), range(N)) not work though? Is the key in repeating the SAME call 16 times?

adisuissa:
In P4.py your code inside the "main" function should be indented. This encourages a reuse of this code (or its functions) and allows one to import the file and use these functions.

HW0/P6/P6B.py

HW0/P4/P4.py
adisuissa: Adding a legend to the graph can ease understanding.

HW0/P5/P5.txt
```diff
@@ -1 +1,61 @@
...
+          9        
+                 11  
+                       13
+                              15
+                                     17
+                                             19
+
+So we spot a pattern that would lead to an added 251s for 256 bags for a total of 259 seconds. And this is without adding time for bringing the bags from one cashier to another. If cashiers were allowed to take 2 bag at a time, then we would reduce the time. 
+
+The strategy is to find the right balance between how many bags to take and how many employees to use. We might be able to gain some time by taking fewer employees in order to save the time they have to take to transfer the bags and communicating.
```
adisuissa: The proposed strategy is unclear and should be more detailed.

PR #141, opened 2015-09-11
Hw0b

serenabooth:


HW0/P5/P5.py
```diff
@@ -0,0 +1,31 @@
...
+
+    parallel_time.append(parallel_time_calc)
+
+plt.plot(num_bags, serial_time, color = 'red', label='Single Cashier')
+plt.plot(num_bags, parallel_time, color = 'blue', label = 'Infinite Cashiers')
+plt.xlabel('Number of Bags')
+plt.ylabel('Time to Sum (sec)')
+plt.title('Number of Bags vs. Time to Verify Total')
+plt.legend()
+plt.show()
```
foo-bar-baz-qux: I think using a logarithmic scale here would help with the visualization.

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
-# Your answers here.
+1. 255 seconds. 
+
+2. Yes. Each cashier can take 1/8th of the bags of pennies, 32 per person, and add the totals for their 32 bags. The cashiers could then pair off to sum their totals, and pair off again, and pair off again. Each cashier can sum their 32 bags in parallel, taking 31 seconds. The eight cashiers can pair off and sum their totals, taking another second. The four remaining cashiers can pair off, another second. And, finally, the remaining two cashiers pair off, another second. In total, 34 seconds. 
+
+3. Assuming no costs for communication or for handing out the bags, 128 cashiers each have two bags. In 1 second, the cashiers each sum their two bags. The cashiers pair off and communicate, which takes no time by assumption. In 1 additional second, the 64 remaining cashiers add two bags again. The employees can do this 7 times (as 2^7=128), thus taking 7 seconds.
```
foo-bar-baz-qux: I think this should be 8 seconds as there's a final calculation on the 0th layer to sum the last two numbers.

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
...
+
+3. Assuming no costs for communication or for handing out the bags, 128 cashiers each have two bags. In 1 second, the cashiers each sum their two bags. The cashiers pair off and communicate, which takes no time by assumption. In 1 additional second, the 64 remaining cashiers add two bags again. The employees can do this 7 times (as 2^7=128), thus taking 7 seconds. 
+
+4. It would take the infinite number of employees log(N) time to count the total. See plot.
+
+5. Assuming 8 cashiers, as in question 2, each time the cashiers pair off, the communication cost takes a second. As this pairing interaction occurs 3 times in question 2, this takes an additional 3 seconds, taking 37 seconds in total. 
+
+Assuming infinitely many cashiers, wherein 128 are useful, the cost of communication in pairing off (from 128 employees to 64 to 32 ...) takes an additional 7 seconds, and so takes 14 seconds total.
```
foo-bar-baz-qux: Based on Q3 being 8 seconds, I think this should be 15 seconds.

HW0/P5/P5.txt
thereisnoquarter: Agreed, I think it should be log_2(256) = 8s.

HW0/P4/P4.py
```diff
@@ -40,37 +42,132 @@
     # Read the observation array and plot it (Part 2)
     #####################
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    C_partial = np.matrix([
```
thereisnoquarter: It's generally recommended to use NumPy ndarrays rather than matrices, although I think they said to use matrices in class so it should be fine.

HW0/P4/P4.py
```diff
@@ -40,37 +42,132 @@
...
+    # construct s using Eq (1)
+    s_column = s_0
+    s[:,0] = s_column
+    for k in range(1,K):
+        s[:,k] = A*s_column + a
+        s_column = s[:,k]
+
+    s = np.array(s)
+    ax.plot(s[0], s[1], s[2],
+             '-k', label='Blind trajectory')
```
thereisnoquarter: This applies throughout, but I would add more comments (maybe referencing equation numbers) describing what everything is doing.

HW0/P4/P4.py
```diff
@@ -40,37 +42,132 @@
...
+
+    def predictSig(A,sig_k,B):
+        return np.linalg.inv(A*sig_k*A.transpose() + B*B.transpose())
+
+    def updateSig(sig_tilda, C):
+        return np.linalg.inv(sig_tilda + C.transpose()*C)
+
+    def updateS(sig_k1, sig_tilda, s_tilda, C, m_k1):
+        return sig_k1 * (sig_tilda*s_tilda + C.transpose()*m_k1)
+
```
thereisnoquarter: This might just be me, but for readability I like to split up my equations into several terms and combine them at the end.

serenabooth: What do you mean?

HW0/P5/P5.py
serenabooth: Yeah, you're right. Thanks!

HW0/P5/P5.txt
```diff
@@ -1 +1,17 @@
...
+
+6. Assuming 256 cashiers, let each cashier take 1 of the 256 bags; this takes 256 seconds. As each cashier only takes 1 bag, the cashiers cannot start their sums until all bags have been distributed. Then, the cashiers pair off, communicate, and add their bags. They repeat this process 8 times with a cost of 2 seconds per iteration. This takes 16 seconds. The total time is 272 seconds.
```
adisuissa: You can overlap communication and computation, and get a better result by doing that (don't need to wait until all cashiers receive all bags).

PR #142, opened 2015-09-11
Hw0b

abhishekmalali:
Completed HW0b assignment.


HW0/P5/P5.txt
```diff
@@ -1 +1,36 @@
...
+6. Approaching the problem by considering that employees are not allowed to pass the bags and only communicate with each other, lets take employees = 8 and each employee gets 32 bags. For this case, the time required for execution is 43 seconds. Similarly for number of employees  = 16 and each employee getting 16 bags, the time is  35 seconds. And for number of employees = 4 and each employee getting 64 bags, the time is 71 seconds. Using mathematical induction, we figure that if the number of employees is n, then the time taken is (256/n) + n +3 seconds. Using the above formula for 1 bag per employee, we get a time of 260 seconds. For 2 bags per employee, we get 133 seconds.
+
+A better approach will be to give 128 bags to each employee who further pass it down the chain in the binary tree. Including the lag in the second set of bags, the total time for the bags to reach the end of the chain is 8 seconds. Adding this time to the time we previously calculated in 5.5 which is required to communicate results up the chain we get an answer of 23 seconds.
```
matheuscfernandes: I wouldn't agree with this strategy. From when I wrote this down, giving the 128 bags to each worker will not be any faster than the previous strategy because now you stop using yourself as a distributor as a resource. Only after the workers distribute the work to other workers are the able to perform the adding themselves.

HW0/P5/P5.txt
```diff
@@ -1 +1,36 @@
...
+4. int(Log2(N)) time since the problem will be a binary tree if we have infinite employees. N will be the next largest value in the 2^n series.
+5. Considering the one second required to communicate between the levels, each level in the Binary tree takes 2 seconds to communicate the result to the next level except for the highest level where the sum is directly calculated and hence has no communication cost. The time required for 256 bags will be 2*Log2(N) - 1 which equals 15 seconds.
+6. Approaching the problem by considering that employees are not allowed to pass the bags and only communicate with each other, lets take employees = 8 and each employee gets 32 bags. For this case, the time required for execution is 43 seconds. Similarly for number of employees  = 16 and each employee getting 16 bags, the time is  35 seconds. And for number of employees = 4 and each employee getting 64 bags, the time is 71 seconds. Using mathematical induction, we figure that if the number of employees is n, then the time taken is (256/n) + n +3 seconds. Using the above formula for 1 bag per employee, we get a time of 260 seconds. For 2 bags per employee, we get 133 seconds.
```
matheuscfernandes: Why are you assuming that emplyees are not allowed to pass the bags? I am confused. The problem asks how does the time change if each employee can only take 1 bag per time vs. 2 bags per time, given that it takes 1 second to hand the bag from you to the employee.

HW0/P5/P5.txt
```diff
@@ -1 +1,36 @@
 # Your answers here.
+
+1. 255 seconds required to sum total 256 bags.
+2. The process of counting would be faster. The time required to calculate the sum total of all the bags would be 34 seconds.
+3. If the store has infinite number of employees, the problems becomes a binomial tree where each level takes 1 second to be solved. Hence the employees can verify the sum in log(N) to the base 2 time where N is the next largest value in 2^n series. Hence the time required for 256 bags will be log(256) to base 2 which is 8 seconds.
+4. int(Log2(N)) time since the problem will be a binary tree if we have infinite employees. N will be the next largest value in the 2^n series.
+5. Considering the one second required to communicate between the levels, each level in the Binary tree takes 2 seconds to communicate the result to the next level except for the highest level where the sum is directly calculated and hence has no communication cost. The time required for 256 bags will be 2*Log2(N) - 1 which equals 15 seconds.
```
matheuscfernandes: Everything up to this point, I have the same results.

HW0/P6/P6.txt
```diff
@@ -1 +1,27 @@
...
+Bye Job 6
+Bye Job 5
+Bye Job 4
+Hi Job 8
+Hi Job 9
+Bye Job 7
+Bye Job 8
+Bye Job 9
+
+It might be so that job 5 may get completed before job 4. A problem where this might actually matter is where the parallel processing job is trying to modify shared data. Any program requiring serialized data where the past data affects the future operations might be affected by this behaviour.A fitting example would be high-frequency trading where there are multiple transactions ordered at the same instant of time. And incase a job where assets might be sold to recoup money to buy in assets be slower to process than the later job, this might cause insufficient balances.
```
matheuscfernandes: This looks good. I would also like to add what I wrote which is that if the input coming in is in a particular order you want your outcome to come out as. Obviously, in such case the order would not be the same as the order the inputs were provided. You would have to somehow sort that or identify which output corresponds to which input.

HW0/P6/P6.txt
```diff
@@ -1 +1,27 @@
...
+
+2.The trend that we observe in the graph is that with increasing wait times the serialised execution takes more time than the parallel execution. Initially the serial execution is faster since parallel execution has a certain overhead to the entire process which leads to slower results. With increasing timings, the overhead slowly becomes negligible and as the graph shows, the parallel execution becomes nearly 4 times faster than the serial execution. And as mentioned, the problem uses 4 threads which means that the longer the delay, the more the parallel program achieves better results since the overhead becomes negligible.
```
matheuscfernandes: Precisely what I have down! :100:

HW0/P6/P6B.py
```diff
@@ -16,16 +16,21 @@ def burnTime(t):
...
+	start = time.time()
+	pool.map(burnTime(t),range(16))
+	parallelTime = time.time() - start
 
         # Compute the ratio of these times
-        # ratio.append(serialTime/parallelTime)
+        ratio.append(serialTime/parallelTime)
 
     # Plot the results
     plt.plot(wait_time, ratio, '-ob')
```
matheuscfernandes: Plt looks good. I got a similar behavior.

evanyao27:
Great work! You seem to have really good mastery of numpy (seeing how you use very concise and very effective function calls). You seem to have a very good grasp of matrix functions (maybe have experience with matlab?).Thanks for teaching me about np.eye for constructing the identity matrix. One small stylistic improvement would be to use multiple lines to define constant matrices; just makes things easier to read. But otherwise, everything looks really good! 

-Evan

HW0/P5/P5.txt
abhishekmalali: This was an assumption that I had to take since I allow only communication and not passing since communication has the same overhead as passing and while the employee is calculating all by himself, he is being much more efficient since the only time spent is in calculation.

PR #145, opened 2015-09-11
[Anna Whitney] HW0b

annawhitney:


HW0/P5/P5.txt
```diff
@@ -1 +1,57 @@
...
+    cashier, and the latter cashier adds their 128th value to their running
+    total; in the 258th second, the former cashier adds the last number to
+    their total; in the 259th second, one communicates their total to the
+    other; and in the 260th second, they add their totals).
+
+    If each cashier takes 2 bags, 133 seconds (the last bags are handed to a
+    cashier in the 128th second; one second later, the last cashier to receive
+    a bag adds their two bags together; and from that point, as above, an
+    additional four seconds are necessary to reach the final total).
+
```
ppgmg: I like your analysis for this part. I did have a bit of trouble following the logic for question 6, but I think I roughly understand your strategy and it seems clever. Is it true that you just use the first two cashiers to collect all the counts from everyone else, and to do the additions, except that they alternate?  If so, does that mean that once one of the other cashiers passes their counts to these first two cashiers, they become idle?

HW0/P6/P6.txt
```diff
@@ -1 +1,51 @@
...
+    processes modify the same resource, for example, and one relies on the
+    output of the other, unexpected behavior may result (e.g., if a bank is
+    processing many transactions in parallel and a user deposits their paycheck
+    and then pays a bill, the deposit may not have completed before the bill
+    payment is deducted from the user's account, resulting in overdraft fees).
+
+PART B.
+    The speedup of parallel relative to serial execution increases as the wait
+    time increases, finally plateauing at roughly the theoretical maximum of
+    4x speedup for wait times on the order of 10^-2 through 10^0 seconds.
```
ppgmg: Good explanations for these.  One thing I thought was a bit curious on your graph was that there were values above the 4.0 theoretical maximum mark.  I guess this could probably be caused from factors causing a slowdown in the execution of the serial process, perhaps unrelated to this experiment (e.g. 'outside' processes running concurrently causing it)?

HW0/P5/P5.txt
annawhitney: Yes, in the everyone-takes-n-bags-from-me-directly scenario, I'm pretty sure this is the best you can do - each cashier adds up their own n bags and then tells the total to a cashier who isn't currently adding (and since only one of the first two cashiers to receive their bags is adding at a time, because they're offset by one second, this means that there's no advantage - at least, not that I could find - to having more than two cashiers collecting the totals). It's still much less efficient than some methods that don't require all the cashiers to come up to me and take bags directly, though.

HW0/P6/P6.txt
annawhitney: Yeah, in practice the serial running time isn't completely fixed, so if on one iteration the serial process was slowed down (other concurrent processes seems like a good guess for why) and the parallel processes weren't, they could be more than 4x faster. If we ran many more iterations with the same wait time and averaged them, we could get a smoother curve that wouldn't go above 4.0, but this is a pretty small sample size so it's noisy.

HW0/P4/P4.py
```diff
@@ -40,37 +46,70 @@
     # Read the observation array and plot it (Part 2)
     #####################
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    ms = np.loadtxt('P4_measurements.txt', delimiter=',')
+
+    unstretch = np.diag(1./np.array([rx,ry,rz]))
```
xinyanhan: Clever way of constructing unstretch matrix!

HW0/P5/P5.txt
```diff
@@ -1 +1,57 @@
...
+    take 31 seconds to add), then four of them each add their total to the
+    total of one of the other four (1 additional second), then two of the
+    remaining four add the remaining two pairs (1 more second), then one
+    cashier adds these results (1 more second), for 34 seconds total.
+
+3.  8 seconds (in the first second, 128 cashiers each add a pair of bags;
+    the next second, 64 add pairs of those results; and so on, until one
+    remaining cashier adds the final total in the 8th second).
+
+4.  Log_2(N) seconds (because the number of sums is halved each second).
```
xinyanhan: For the infinite cashiers time, is it int(log_2(N)), since the seconds cashiers take are all in integers.

HW0/P5/P5.txt
```diff
@@ -1 +1,57 @@
...
+    Using this same basic algorithm (give some number of bags to each cashier,
+    have them add up all those bags, then combine their total with one of two 
+    cashiers keeping a running total by alternately talking and adding), the 
+    best we can do is handing 16 bags to each cashier, which takes 35 seconds 
+    (the last set of bags is handed to a cashier in the 16th second; 15 
+    seconds later, the last cashier to receive their bags finishes totaling 
+    their bags; and four more seconds are required to reach the final totals).
+    Going to 32 bags doesn't decrease the time further, taking 43 seconds 
+    (only 8 to hand out the bags, but 31 for each cashier to do their sums, 
+    plus 4 to finish).
```
xinyanhan: I like your clear explanation in so much details! My answer to the better strategy is also to hand 16 bags to each of the 16 cashiers, but the optimal time they take that I can think of is 16+15+5 = 36, where 16 is the hand-bag time, 15 is the time for each cashier to count 16 bags, and 5 is the extra time to add up and communicate. Could you please explain a   bit more on how you derived the 4 seconds to reach the final totals?

HW0/P5/P5.txt
```diff
@@ -1 +1,57 @@
...
+
+    However, we can do even better (23 seconds) with a different approach:
```
oyuchoi: yes, this is better!

PR #146, opened 2015-09-11
Hw0b

jaco8:
HW0b


HW0/P5/P5.py
```diff
@@ -0,0 +1,16 @@
...
+import matplotlib.pyplot as plt
+import math
+
+# Main
+if __name__ == '__main__':
+	x = np.linspace(1, 1000, 20)
+	y1 = [math.log(i, 2) for i in x]
+	y2 = [(i - 1) for i in x]
+	plt.plot(x, y1, '-b', label = 'Infinite number of employees')
+	plt.plot(x, y2, '--r', label = 'Lone cashier')
```
jacobbaron: It would be great to plot this on a log scale, since we can't really see what the parallelized plot is doing.

HW0/P5/P5.txt
```diff
@@ -1 +1,21 @@
...
+
+6. 1 bag: For total 256 bags, we ignore the handing time first, each person takes 1 bag, so we have 256 groups. Same idea as problem 5, but a little bit different is 256 people should communicate first, it takes an extra 1 second. So the time without handing time is 16 seconds. Now for the handing time, from 256—>1, it takes 8 seconds, but we need 1 extra second to pass bags to cashier. Therefore, the total time will be 16 + 8 + 1 = 25 sec
```
jacobbaron: My understanding of this problem was that you had to hand each employee one bag, and you can only hand them bags one at a time. It also seems that once the bags are handed out from you to the employees, no more handing needs to be done. Thus, when you distribute the bags serially - one bag to one worker - it should take 256 seconds to distribute.

TuanTrain: I interpreted the question slightly differently, with each bag having to go through you personally, meaning 256 seconds to hand out every bag. While that happened 2 other cashiers can add up the bags in parallel so it would take 256 seconds + remaining time to finish counting the last bag you gave out, which in my method would be 4 seconds. I think here you are explaining the process of handing out the bags in a log_2(N) fashion, the way it was communicated and counted in part 5.

HW0/P6/P6.txt
```diff
@@ -1 +1,34 @@
...
+Bye Job 8
+Bye Job 9
+[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+Since Job 0 to Job 3 is in ascending order, but it not terminate with in this order. Since Synchronization is very important in parallel which effect the whole process. 
+
+ 
+How much faster? 
+
+Under the condition from 10^(-6)—>10^(-4), since the task is relatively small, serial version is faster than parallel version. However, under the condition from 10^(-3)->10^(0), the task is relatively large, parallel version will take shorter than serial version. Therefore, if we program this in parallel, it will take much more time to run this program if the task is relatively small, since opening a processor costs much more time. But if the task is relatively large, using multiprocessor will make the whole process much faster.
```
TuanTrain: I agree with your answers here. Synchronization of processes is important in parallel programming to ensure the data is not messed up in separation. 
On my computer the switch from parallel being slower to it being faster was at around 10^(-5), and I also reasoned that it was caused by overhead of parallelizing.

HW0/P5/P5.txt
```diff
@@ -1 +1,21 @@
...
+
+   2 bags: Same idea as problem 5, for the handing time, from 256—>1 which is cost 8 seconds. Therefore, the total time will be 15 + 8 = 23 seconds. 
+
+   Yes. There is a better strategy which is 4 bags. Same idea as above, the total time will be 15 + 7 = 22 seconds. Since 8 bags if one person do lone, it takes 7 seconds; If 8 bags spilt into 2 groups, it will take 3 (4 bags takes 3 seconds)+ 2 + 1 = 6 seconds. Therefore, 4 bags is the best strategy.
```
oyuchoi: What about more than 4 bags?

PR #149, opened 2015-09-11
Hw0b

matheuscfernandes:
Finished all parts of HW0


HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
     # Load true trajectory and plot it
     # Normally, this data wouldn't be available in the real world
     #####################
+    s_true=np.loadtxt('P4_trajectory.txt',delimiter=',')
```
btweinstein: You can also use pandas to load information from text files; I prefer that. But what you did is great too!

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
 
     #####################
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
-    # s = ?
+    A = np.matrix([[1,0,0,dt,0,0], [0,1,0,0,dt,0], [0,0,1,0,0,dt],
```
btweinstein: Although what you did is fine here, you can also initialize matrices like

A = np.matrix([
                      [1, 0, 0...],
                      [2, 0, 0...],
                      ])

Which might be easier to read.

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
+        [0,0,0,1-c*dt,0,0], [0,0,0,0,1-c*dt,0], [0,0,0,0,0,1-c*dt]])
+
+    a = np.matrix([0,0,0,0,0,g*dt]).transpose()
+    s = np.zeros((6,K))
+    s0 = np.matrix([0,0,2,15,3.5,4.0])
+    s[:,0] = s0
+    s = np.asmatrix(s)
+
+    for i in xrange(0,K-1):
+        s[:,i+1]=np.add(np.dot(A,s[:,i]),a)
```
btweinstein: I think this line can be rewritten a little more simply, although what you did is perfectly fine! I would probably do

s[:,i+1]=A.dot(s[:, i]) + a

Arrays often have functions like np.dot implemented within them to make your code easier to read. But again, what you did is fine!

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
+    B[0,0],B[1,1],B[2,2],B[3,3],B[4,4],B[5,5]=bx,by,bz,bvx,bvy,bvz
+    C[0,0],C[1,1],C[2,2]=rx,ry,rz
+    B=np.asmatrix(B)
+    C=np.asmatrix(C)
+    
+    Sigma=np.identity(6)*0.01
+
+    def predictS(A,s,a):
+        return np.dot(A,s)+a
+
```
btweinstein: Similar to what I wrote above, predictSig can be rewritten a little more intuitively as

lin.inv(SigmaBar + C.T.dot(C))

This is easier to read I think.

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
+        Sig_updated = updateSig(Sig_predicted,C)
+        s[:,i+1]=updateS(Sig_updated,Sig_predicted,s_predicted,C,m[:,i])     
 
     # Initial conditions for s0 and Sigma0
     # Compute the rest of sk using Eqs (2), (3), (4), and (5)
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    s=np.asarray(s)
+    ax.plot(s[0,:], s[1,:], s[2,:],
```
btweinstein: Nice job! Your plot is beautiful. I generally prefer to work with numpy arrays directly instead of the numpy.matrix wrapper (as it always confuses me), but it is a matter of taste.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
-# Your answers here.
+By: Matheus C. Fernandes
+
+Question 1: 255 seconds
+	If one cashier were counting the add the totals of 256 bags it would take 255 seconds because the first count 1+1=2 would count as one secong and every count thereafter would also count as one second
```
btweinstein: Yep, I agree.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
-# Your answers here.
+By: Matheus C. Fernandes
+
+Question 1: 255 seconds
+	If one cashier were counting the add the totals of 256 bags it would take 255 seconds because the first count 1+1=2 would count as one secong and every count thereafter would also count as one second 
+
+Question 2: 34 seconds
+	Yes, assuming they imidiatelly divide up the work evenly, they can deivide the 256 bags by 8 workers. Each worker than has to count 32 bags which would take 31 seconds to count. Then 4 workers can add up the other bags, taking 1 additional second. Then 2 workers take the sum of 4 people, taking 1 additional second. Finally, the 1 worker adds the final sum of two people, during the last 1 second.
```
btweinstein: I got the same answer is you, but I found the explanation slightly confusing. You might consider putting in a simple asci graphic, i.e. 

x x x x x x x x
 |      |     |    |
 x   x   x   x
     |       |
   x       x
        |
        x

To show the part where they sum up a tree.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+
+Question 3: 8 seconds
```
btweinstein: Nice, I used log2 as well and found the same answer.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+	Assuming there is no communication time between workers, it would take 8 seconds. This starts with having 256 workers and taking each second to add each of the half of the workers namely it takes log2(256)=8.
+	
+Question 4: log2(N), please find attached P5.png
```
btweinstein: I agree with this, but it doesn't look like your plot is correct. Somehow the infinite number of workers plot is flat. Is it an issue with the scale of the plot? I would use a log-log plot to better show the two methods over many orders of magnitude.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+	It would take infinite number of employees log2(N) seconds to count N bags.
+
+Question 5: 15 seconds
```
btweinstein: Can you expand on what you mean by "except for the first case?" I wrote that the required time was 16 seconds. The first step, you still have two communicate two numbers to each other and add them, taking two seconds, right?

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+	Every time there is a sum between two numbers there is one additional second, except for the first case. 
+
+Question 6: **this part was done in colaboration with Thibaut Perol**
+	(each cashier comes and takes 1 bag): 230 seconds
```
btweinstein: This doesn't make sense to me as it will take 256 seconds for you to hand out all the bags. So I don't think you can total the amount before then! I think that the answer is 257 seconds. Please let me know if I'm wrong!

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+		picture attached 'P6_1bagperworker-8bags.jpg,P6_1bagperworker-16bags.jpg'
+
+	(each cashier comes and takes 2 bags): 132 seconds
```
btweinstein: I found 142 seconds, but you might be correct. A formula explaining the general case would be nice, but is probably not necessary. Did you take into account the fact that you have to communicate in the last bit?

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+		picture attached 'P6_2bagperworker-16bags.jpg'
+
+	(Better strategy?): No, for having a system where we count two bags at one time, we cannot do better than handing each person two bags at a time. If they were able to count more than two bags at a time, then we would be able to change this strategy.
```
btweinstein: This doesn't make sense to me. I will copy my answer:

The most efficient strategy is to distribute bags as fast as
possible. It only takes t_hand to hand *any number of bags* to an employee.
Therefore, we should employee another binary tree method to hand bags
out to employees as fast as possible. Sketching the idea,

      N
    |   |
   N/2 N/2
|    |   |   |
N/4 N/4 N/4 N/4

etc. You start with N bags and hand half to the next employee.
That employee hands half to another employee, and this repeats until
each employee has 1 bag. At this point, employees sum down the binary
tree via communicating with each other. The solution will then scale as
Log2[N] which is quite fast. We thus find

t_minimum = Log2[N] t_hand + Log2[N](t_add + t_communicate)

Simplifying,

t_minimum = Log2[N](t_hand + t_add + t_communicate)

Evidently, the fastest handing strategy results in 3 times the original
infinite solution (without communication lag). Plugging in numbers,
we find

t_minimum = 24 seconds

HW0/P6/P6.txt
```diff
@@ -1 +1,31 @@
...
+
+		Bye Job 7
+		Hi Job 8
+		Hi Job 9
+		Bye Job 8Bye Job 9
+
+		[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+		[Finished in 1.0s]
+
+	The order in which the jobs are completed change from one run to the other, this is a result of doing the jobs in parallel. This could be a problem when the input of a job requires an output of a previous job or a job that is in the order of the process. Another possible issue when we want the output to be in order of processes. For example, I know I have 4 cores, so I have the dependence of job 5 on one of the 4 first jobs. This scenario could create a conflict.
```
btweinstein: Good explanation. Even in the same run, however, you can't control precisely when jobs run and finish!

HW0/P6/P6.txt
```diff
@@ -1 +1,31 @@
...
+
+How much faster?:
```
btweinstein: Do you know why your plot seems to shoot above 4 at certain times? That is pretty weird. Naively, the maximum speedup should be 4x, right?

HW0/P6/P6B.py
```diff
@@ -16,16 +19,29 @@ def burnTime(t):
...
+
 
     for t in wait_time:
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
+        timeBegining = time.time()
+
+        pool.map(burnTime, t*np.ones(N))
```
btweinstein: This is a good way of calling 16 processes; I did it by 

for i in range(N/4): 
            pool.apply(burnTime, args=(t,))

I like your way better.

HW0/P4/P4.py
drishi: Pandas is especially great to serialize python objects to "save" your progress, but I used np.loadtxt here as well

HW0/P4/P4.py
drishi: ^^Good tip for readability, I will also keep that in mind for next time.

HW0/P4/P4.py
drishi: Just as a note, I always use np.dot(arg 1, arg2) because that's always been the notation I've seen when working in numpy. Could be a matter of personal preference, though I do see the left-to-right readability benefit from what btweinstein mentioned.

HW0/P4/P4.py
drishi: Great! Your plot looks just like mine

HW0/P5/.ipynb_checkpoints/P5-checkpoint.ipynb
```diff
@@ -0,0 +1,115 @@
...
+   "mimetype": "text/x-python",
+   "name": "python",
+   "nbconvert_exporter": "python",
+   "pygments_lexer": "ipython2",
+   "version": "2.7.10"
+  }
+ },
+ "nbformat": 4,
+ "nbformat_minor": 0
+}
```
drishi: Hmm what is this for exactly? Didn't have this bit in my code.

HW0/P5/P5.txt
drishi: +1

HW0/P5/P5.txt
drishi: I agree that this seems low but I got an even higher number of 272s. I wrote:
It takes us 256 seconds to hand out every bag to all the cashiers,
+and one second to pass bags between workers so that they are able to have two bags to be able to add
+together. From there on, we have the same situation as above, and can do 256 + 1 + 15 to get a total of
+272 seconds.

HW0/P5/P5.txt
drishi: I got 143 seconds here

HW0/P5/P5.txt
drishi: ^^Great explanation +btweinstein

HW0/P6/P6B.py
```diff
@@ -33,4 +49,5 @@ def burnTime(t):
     plt.xlabel('Wait Time (sec)')
     plt.ylabel('Serial Time (sec) / Parallel Time (sec)')
     plt.title('Speedup versus function time')
+    savefig('P6.png', bbox_inches='tight')
     plt.show()
```
drishi: Yep, more or less exactly how I did it but I think you did a better job labeling your graph

HW0/P6/P6.txt
matheuscfernandes: Hi Bryan, you are right, I was slightly confused as well. But they I realized that I think this has something to do with when the timer starts and ends. I think it is an artifact of timer error. I refined my points for the different pause times too much and that is why you see some of the differences in more detail.

HW0/P4/P4.py
matheuscfernandes: Thanks for these suggestions! That is very helpful. I will make sure to use that next time I load the text files.

HW0/P4/P4.py
matheuscfernandes: Thanks Bryan, I will be sure to do that next time. I actually had trouble typing this up too. That way I am less prone to making mistakes.

HW0/P4/P4.py
matheuscfernandes: Good point, I would have done it if it were on MATLAB. I am still working on my transition to Python and this is certainly very useful feedback!

HW0/P4/P4.py
matheuscfernandes: Thanks to both suggestions. I will keep this in mind.

HW0/P4/P4.py
matheuscfernandes: Excellent, thanks!

HW0/P5/P5.txt
matheuscfernandes: Yeah, I think I went a little to far on my x-axis. You are right, I should have used a log-log plot for this.

HW0/P5/.ipynb_checkpoints/P5-checkpoint.ipynb
matheuscfernandes: This is for the iPython notebook I used to create the plot. It was just saving a checkpoint of my files. Nothing you really need to have.

HW0/P4/P4.py
```diff
@@ -30,47 +30,97 @@
...
+
+    s = np.zeros((6,K))
+    s[:,0] = s0
+    s = np.asmatrix(s)
+
+    m=np.asmatrix(measurements).transpose()
+
+    for i in xrange(0,K-1):
+        s_predicted = predictS(A,s[:,i],a) 
+        Sig_predicted = predictSig(A,Sigma,B)
```
adisuissa: Sigma is passed to predictSig, and it doesn't change in the loop.
In the next iteration the same Sigma value will be passed, instead of the new value (look again at how Kalman filter is defined).

HW0/P5/P5.txt
adisuissa: Indeed if it takes you at least a second to hand one bag to a cashier, then it will take at least 256 seconds to hand all bags.

HW0/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+	
+Question 4: log2(N), please find attached P5.png
+	It would take infinite number of employees log2(N) seconds to count N bags.
+
+Question 5: 15 seconds
+	Every time there is a sum between two numbers there is one additional second, except for the first case. 
+
+Question 6: **this part was done in colaboration with Thibaut Perol**
+	(each cashier comes and takes 1 bag): 230 seconds
+		picture attached 'P6_1bagperworker-8bags.jpg,P6_1bagperworker-16bags.jpg'
```
adisuissa: It's not really clear from the pictures what the strategy is.
I would suggest that next time you accompany the pictures with some text that explains the algorithm you are proposing.

HW0/P4/P4.py
matheuscfernandes: Adi, I do not understand what your point, the value of Sigma is updated at the end of the for loop so that the initial value of sigma is taken into account. Could you please clarify what you mean?
Thanks!

HW0/P5/P5.txt
matheuscfernandes: The strategy was that we looked at the combinations of workers. The graph represents merges in one color and counts in another color. Given the particular amount of workers and what they can handle, we drew up different schemes where the work could be handed off to each worker. Each increment of time (flop) is represented by the number inside the node in the graph. Therefore the most efficient way we could think of, is drawn through the nodes and through the connections. This, by nature of the problem, is difficult to explain in words. That is why we drew up the diagram and made is as clear as we could through the legend and the numbering of the nodes.

HW0/P4/P4.py
adisuissa: Look again at the problem description, Eq. (3).
The new value of sigma is dependent on A, B, and Sigma_k, where Sigma_k is the value of the previous iteration (Eq. (4)).
In your code you Sig_predicted uses A, B, and Sigma_0 (the initial value of Sigma).
I think that this value is not updated in the rest of loop execution, but do let me know if I've missed something.

matheuscfernandes: @adisuissa, I see it now. Yeah, you are right, I forgot to update sigma. Thank you for the feedback.

PR #151, opened 2015-09-11
Hw0b

quezacot:


HW0/P4/P4.py
```diff
@@ -29,20 +29,27 @@
...
     #         '--b', label='True trajectory')
-
+    ax.plot(s_true[:,0], s_true[:,1], s_true[:,2], '--b', label='True trajectory')
+    
     #####################
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
+    s_obs = np.loadtxt('P4_measurements.txt', delimiter=',')
+    R = np.array([[1.0/rx, 0, 0], [0,1.0/ry,0], [0,0,1.0/5.0]])
```
dominedo: This is a small detail, but I would reference rz (as you did with rx and ry) instead of hard-coding it.

HW0/P4/P4.py
```diff
@@ -51,12 +58,31 @@
     # A = ?
     # a = ?
     # s = ?
-
+    
+    A = np.array( [ [1,0,0, dt,  0,  0],
```
dominedo: I defined A as you did, but one of my peer reviewers suggested that np.diag() could be a more efficient method, especially for larger matrices than this one.  I found it a useful comment so I thought I'd pass it on!

HW0/P4/P4.py
```diff
@@ -51,12 +58,31 @@
...
+                    [0,0,0, 1-c*dt, 0, 0],
+                    [0,0,0, 0, 1-c*dt, 0],
+                    [0,0,0, 0, 0, 1-c*dt] ])
+    
+    a = np.array( [0,0,0,0,0,g*dt] ).reshape([-1,1])
+    
+    s_prop = np.empty([6,K])
+    
     # Initial conditions for s0
+    s0 = np.array([0,0,2,15,3.5,4.0]).reshape([-1,1])
```
dominedo: Instead of hard-coding the parameters, you could reference s_true[0,:] instead.  That way you only have to update the values in one place if they change.

HW0/P5/P5.txt
```diff
@@ -1 +1,27 @@
...
+3. Now each round only needs 1 second to count:
+   1 + 1 + 1 + 1 + 1 + 1 + 1 + 1 = log2(256) = 8 seconds
+
+4. See P5.png. Single cashier takes N - 1 seconds. Infinite cashiers takes ceiling(log2(N)) seconds.
+   Plot number of bags from 2 to 1200
+
+5. The communicating time takes effect since the second round count. Each round increases 1 second.
+   1 + 2 + 2 + 2 + 2 + 2 + 2 + 2 = 15 seconds
+ 
+6. The cashier who holds all the 256 bags in the beginning distributes the bags to other cashiers. They distribute the bags based on this rule: give the half number of bags in hand to another cashier who has no bags in hand.
```
dominedo: I interpreted this question to mean that the person holding all the bags in the beginning can only distribute them one or two at a time. It looks like your interpretation was different, i.e. the end result of the distribution is that each cashier is only holding one or two bags, but the person initially holding the bags can distribute more than one/two at a time.  Your interpretation is definitely faster (e.g. 25 seconds in part 1, as opposed to 260 seconds in my interpretation)!

HW0/P6/P6.txt
```diff
@@ -1 +1,37 @@
...
+Hi Job 9
+Bye Job 8
+Bye Job 9
+
+We can see there are unexpected results of the job orders:
+1) The order jobs begin does not match the order we send in code.
+2) The order jobs end does not match the order they begin.
+3) Run this program several times, the sequences may different every time.
+
+1) would be a concern when we need to serve many identical tasks with different owners. In general, we should follow the rule: first come, first served. However, this program cannot guarantee this fiarness.
```
dominedo: I'm not sure I fully understand your concern #1. Is the only problem that the order of allocation is unfair, or do you mean that this can also create problems with the running/output of the jobs?

HW0/P4/P4.py
```diff
@@ -65,12 +91,43 @@
...
+        mk1 = s_obs[i].reshape([-1,1])
+        s_nextkal = updateS(sigma_nextkal, sigma_tilt, s_tilt, mk1)
+    s_kal = s_kal.T
 
     # ax.plot(x_coords, y_coords, z_coords,
     #         '-r', label='Filtered trajectory')
+    ax.plot(s_kal[:,0], s_kal[:,1], s_kal[:,2], '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
```
matheuscfernandes: I got the same plot. I would suggest adding the plot as a .png just for reference, even though the problem doesn't explicitly ask for it. Otherwise all looks good to me!

HW0/P4/P4.py
matheuscfernandes: I agree! Another way to do it would be to create a matrix of zeros and set different indices to have certain values. You can just write the variable name and the index of the variable you would like to change.

HW0/P4/P4.py
```diff
@@ -65,12 +91,43 @@
...
+        s_tilt = predictS(s_nextkal)
+        sigma_tilt = predictSig(sigma_nextkal)
+        sigma_nextkal = updateSig(sigma_tilt)
+        mk1 = s_obs[i].reshape([-1,1])
+        s_nextkal = updateS(sigma_nextkal, sigma_tilt, s_tilt, mk1)
+    s_kal = s_kal.T
 
     # ax.plot(x_coords, y_coords, z_coords,
     #         '-r', label='Filtered trajectory')
+    ax.plot(s_kal[:,0], s_kal[:,1], s_kal[:,2], '-r', label='Filtered trajectory')
```
matheuscfernandes: I got a similar plot on the end. :100:

HW0/P6/P6B.py
```diff
@@ -16,16 +16,24 @@ def burnTime(t):
...
+        pool.map(burnTime, [t]*N)
+        parallelTime = time.time() - parallelTime
+        
+        print "serialTime", serialTime, "parallelTime", parallelTime
         # Compute the ratio of these times
-        # ratio.append(serialTime/parallelTime)
+        ratio.append(serialTime/parallelTime)
 
     # Plot the results
     plt.plot(wait_time, ratio, '-ob')
```
matheuscfernandes: I think the question explicitly asks to add the plot for this problem and name it P6.png. I would also suggest adding the graphs for the previous problems even though the problem may not explicitly ask for them. Just makes it easier for us to compare.

HW0/P5/P5.txt
matheuscfernandes: I interpreted this question the same way as @dominedo.

HW0/P6/P6.txt
```diff
@@ -1 +1,37 @@
...
+
+2) and 3) would be a concern when we care about the schedule. For example, when a user split one tasks to multiple independent tasks to benefit from parallel computing. The user still needs to wait all the tasks finished. This program cannot predict the ending sequence and thus schedule the tasks to end close to the same time that minimizes the user's waiting time.
```
matheuscfernandes: I think another point here would be that if the data is shared and one compute node relies on the output data from another compute node that is in a different ordered outcome. This would cause a wait time on the advanced compute node who has to wait for the other node to be finished.

HW0/P5/P5.txt
adisuissa: Yep, it seems like you misinterpreted the question.

PR #152, opened 2015-09-11
HW0b

danfortunato:


nithintumma:
Hi,
I thought your coding style was great; clear and easy to follow. I thought your comments were great as well, especially explaining the constants in the beginning of main() P4.py. I really liked your concise construction of the matrices in P4.py
+    A = np.eye(6)
+    A[3,3] = A[4,4] = A[5,5] = 1-c*dt
+    A[0,3] = A[1,4] = A[2,5] = dt
+
+    a = np.zeros(6)
+    a[-1] = g*dt

Best,
Nithin

xiaweiwang:
Nice and clean code. Your comments are also clear and easy to follow.
For Q6, what do you think of the following improved strategy?
Distribute bags from top down, i.e., we distribute 128 bags to the first two cashiers and they each distribute bags evenly to two cashiers and so on. In this case, the total distribution time will be 8 seconds. As we calculated in Q5, the communication + counting time is 15 seconds. So the total time is 23 seconds.

danfortunato:
Thanks guys! I agree @xiaweiwang , that is a better strategy!

PR #153, opened 2015-09-11
Hw0b

vivjay30:


HW0/P5/P5.txt
```diff
@@ -1 +1,10 @@
...
+3) If everyone is working it is the log base 2 of 256 which is 8 seconds. This only requires 128 people, any more would be redundant
+4) Like part 3 it's the log base two of N since you add two numbers at a time and combine them and repeat. On the plot the blue is the linear time it takes one person to add the numbers, and the green is the logarithmic time 
+5) This is 15 seconds. It takes twice as long as in part three because after every round of counting, you need exactly one second for half the people to tell the other half. It is 1 less than double because at the end you don't need to tell anyone the final count. 
+6) Assuming it still takes one second to communicate a number, the answer becomes 512. Starting with the simple scenario of two bags, after two seconds you have handed out both bags, third second to communicate a number, fourth second to add. With a third bag, the person has to communicate the number, and an addition has to be made again giving six seconds. The addition of a bag leads to an additional two seconds in every case.
```
annawhitney: The problem with this analysis is that it doesn't scale linearly from the two-bag case. With more bags and more cashiers, different cashiers can be doing different things at the same time, so it doesn't have to take as long as 512 seconds. Each pair of bags might take 4 seconds to process, but those 4 seconds can overlap with the 4 seconds of processing another pair of bags, so the total time in seconds doesn't have to be twice the number of bags.

HW0/P5/P5.txt
```diff
@@ -1 +1,10 @@
...
+
+If you give out two bags at a time it takes 257 seconds. That is because each additional two bags you give out requires the person to add them, and tell it to another person while the next bags are being given out and added. Four bags takes five seconds, at the end of 2s person two has the bags, at the end of 3s person two has added, at the end of 4s person two tells person one, at the end of 5 person one has added together. Then you go from there and the answer becomes 257. I'm assuming each person can only do one thing at a time, for example you can't give a bag while recieving a number at the same time. 
+
+The fastest way to do it is as follows. Hand half to one person. Then you and that person hand half to two others and so on until everyone has 2 bags after 7 seconds. Then everyone adds those two, and half tell someone else. That continues until there is one sum after 20 seconds.
```
annawhitney: Since you're not a cashier and are presumably unwilling to help the cashiers count your pennies, the first step should be you handing all the bags to one cashier, and then the rest progresses like you said (this only adds one second, though).

HW0/P6/P6B.py
```diff
@@ -16,16 +17,25 @@ def burnTime(t):
 
     # Use a variety of wait times
     ratio = []
-    wait_time = []
+    # Get 16 random wait times
+    wait_time = sorted([random.uniform(.00000000001, .001) for i in range(16)])
```
annawhitney: This is an interesting choice for how to generate the wait times. Since we're plotting on a log scale, it would be nice to have logarithmically uniformly distributed values rather than linearly uniformly distributed values - I wonder whether the random module provides an analog to random.uniform that does that?

HW0/P5/P5.txt
chengchengqin: I also think that Q6 may be not correct. It can be much much shorter.

HW0/P5/P5.txt
chengchengqin: Nice plot but I suggest that the plot should not be continuous.

HW0/P6/P6B.py
chengchengqin: Your example of translating some text is very vivid. I haven't thought of such suitable example before.

HW0/P6/P6.txt
```diff
@@ -1 +1,6 @@
 # Your answers here
```
zezhouliu: Your graph for P6 does not give any insight on how the performance scales

PR #156, opened 2015-09-11
Hw0b

thereisnoquarter:
Finished part b of HW0.


HW0/P5/P5.txt
```diff
@@ -1 +1,22 @@
...
+
+6) We can write down a cost function for N bags and q workers as C(N, q) = q + (N/q - 1) + 2 * log_2(q). The first (q) term is the amount of time it takes to hand out the bags to the workers, it takes (N/q - 1) seconds for each worker to count their own bags, and then it takes 2 * log_2(q) to communicate and add the sums between the workers in parallel. Thus we have:
+
+Each takes one bag = C(256, 256) = 256 + 2 * log_2(256) = 272s.
+Each takes two bags = C(256, 128) = 128 + 1 + 2 * log_2(128) = 143s. 
+
+Taking the derivative of C(256, q) with respect to q and solving the resulting quadratic equation and rounding down gives an optimal q ~ 14.622. This gives C(256, 14.62) = 38.87. Clearly, we need to choose between q = 14 and q = 15 in real life, and q = 15 tives the slightly better answer of C(256, 15) = 38.88.
```
zxjyjr: I would say q here can only be intergers, since it is workers number.

HW0/P5/P5.txt
```diff
@@ -1 +1,22 @@
...
+
+An added complication is that for q = 14 or 15, 256 / q is not an integer. This means that, for q = 14, the best (even) distribution we can do is 18 bags per person. Because 14 * 18 = 252, we will assign one extra bag to four separate people. This will addone extra computation to the first layer, and thus our optimal time is 39.88s.
```
zxjyjr: For the same reason above, optical time could be only intergers. The time of passing a bag and count a bag is an integer, so the total time could only be integers.

HW0/P5/P5.txt
```diff
@@ -1 +1,22 @@
...
+
+Each takes one bag = C(256, 256) = 256 + 2 * log_2(256) = 272s.
+Each takes two bags = C(256, 128) = 128 + 1 + 2 * log_2(128) = 143s. 
+
+Taking the derivative of C(256, q) with respect to q and solving the resulting quadratic equation and rounding down gives an optimal q ~ 14.622. This gives C(256, 14.62) = 38.87. Clearly, we need to choose between q = 14 and q = 15 in real life, and q = 15 tives the slightly better answer of C(256, 15) = 38.88.
+
+An added complication is that for q = 14 or 15, 256 / q is not an integer. This means that, for q = 14, the best (even) distribution we can do is 18 bags per person. Because 14 * 18 = 252, we will assign one extra bag to four separate people. This will addone extra computation to the first layer, and thus our optimal time is 39.88s. 
+
+One final complication is that, because addition can occur while the bags are still being passed out, the time will actually be slightly less than stated here. Finding the optimal division appears to be an integer programming problem that I do not know how to solve, so I felt that this was sufficient.
```
zxjyjr: Yes, i agree that it's not easy to give the best strategy. But I think it's possible to give a better strategy that cost less time than passing 1 or 2 bags to each cashier. For example, give 16 person each 16 bags, then it would take 29 seconds in total

HW0/P5/P5.txt
araftery: Since the number of steps in the combining phase is actually ceil(log_{2}{256/q}), For the optimal `q`, `256 / q` must be an integer (i.e., a power of 2). For example, if we used 13 cashiers, handing out 20 bags to 12 of them and 16 bags to 1 of them, it would still take 4 rounds, not 3.68, to combine.

So `q = 16` is the closest number to the optimum you found that works.

HW0/P6/P6B.py
```diff
@@ -16,19 +18,31 @@ def burnTime(t):
...
     for t in wait_time:
+        print 'On wait time', t, '!'
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
+
+        # Calculate the time it take serially
+        serialTime = time.time()
+        for ii in range(N):
```
araftery: You should probably use `xrange` here, which returns a generator instead of a list, as it is an advantage of the serial program that it does not need to save a list of numbers to memory, it just needs each number once.

HW0/P6/P6B.py
```diff
@@ -16,19 +18,31 @@ def burnTime(t):
...
+            burnTime(t)
+        serialTime = time.time() - serialTime 
+
+        # Calculate the time it takes in parallel (why are we doing this in a loop)
+        parallelTime = time.time()
+        pool.map(burnTime, [t for x in range(N)])
```
araftery: You could just use `[t] * N` for the second argument, which is faster:

```
In [1]: import timeit

In [2]: timeit.timeit("y=5; [y for x in range(10)]")
Out[2]: 1.1208350658416748

In [3]: timeit.timeit("y=5; [y] * 10")
Out[3]: 0.2912890911102295```

File: HW0/P4/P4.py [hunk 6]
```diff
@@ -2,6 +2,73 @@
...
+    second_term = np.dot(B, np.transpose(B))
+
+    # Combine for the parenthetical term
+    paren = first_term + second_term
+
+    # Invert and return sigma twiddle
+    return np.linalg.inv(paren)
+
+
+def updateSig(Sig_twid, C):
```
araftery: Try to avoid using uppercase variable names like `Sig_twid`, as these are generally reserved for class names in Python (see [PEP8](https://www.python.org/dev/peps/pep-0008/#method-names-and-instance-variables))

HW0/P4/P4.py
```diff
@@ -40,37 +116,145 @@
...
+                        [0, ry_inv, 0],
+                        [0, 0, rz_inv]])
+
+    # Load in the measured data
+    m_data = np.loadtxt('P4_measurements.txt', delimiter=',')
+
+    # Separate the data
+    m_x = m_data[:, 0]
+    m_y = m_data[:, 1]
+    m_z = m_data[:, 2]
```
araftery: you can do this in one line:
`m_x, m_y, m_z = m_data[:, :3]`

HW0/P4/P4.py
```diff
@@ -40,37 +116,145 @@
...
+        curr_vec = np.array([m_xk, m_yk, m_zk])
+
+        # "fix" the vector by application of inv_mat
+        fixed_curr_vec = np.dot(inv_mat, curr_vec)
+
+        # Extract the data into individual lists
+        m_x_fixed[kk] = fixed_curr_vec[0]
+        m_y_fixed[kk] = fixed_curr_vec[1]
+        m_z_fixed[kk] = fixed_curr_vec[2]
+
```
araftery: You can replace this loop and use `np.multiiply` instead, for example:
```
m_fixed = np.multiply(m_data, np.array([1. / rx, 1. / ry, 1. / rz]))
```

HW0/P4/P4.py
thereisnoquarter: Actually, you can't - they'll all point to the same place in memory, and modifying one of them will modify all of the others.

HW0/P5/P5.txt
```diff
@@ -1 +1,22 @@
...
+
+4) Using the procedure described in the previous question, it takes log_2(N) in parallel vs N in serial. This is plotted on P5.png, although we use log(N) rather than log_2 for simplicity - the trend is conserved.
+
+5) We do everything in the exact same way as before, except now we must first communicate the amount in each pile before adding. Clearly the communication can be done in parallel, so this simply modifies the constant for each "addition" from 1 to 2. Thus, we double the amount of time it takes, and the answer is 2 * log_2(256) = 16s. But actually, if we hand two bags to each person, we can avoid communication for the first layer and get an optimal result of 15s.
+
+6) We can write down a cost function for N bags and q workers as C(N, q) = q + (N/q - 1) + 2 * log_2(q). The first (q) term is the amount of time it takes to hand out the bags to the workers, it takes (N/q - 1) seconds for each worker to count their own bags, and then it takes 2 * log_2(q) to communicate and add the sums between the workers in parallel. Thus we have:
```
adisuissa: You can overlap communication and computation, and get a better result by doing that (don't need to wait until all workers receive all bags).

PR #157, opened 2015-09-11
Hw0b Finished

evanyao27:


HW0/P4/P4.py
```diff
@@ -40,37 +43,113 @@
...
     # Initial conditions for s0 and Sigma0
     # Compute the rest of sk using Eqs (2), (3), (4), and (5)
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    ax.plot(xs, ys, zs,
+            '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
```
abhishekmalali: Plot looks good. Something peculiar was use of assert statements to check shapes of the matrices. I wish I had done that. i had saved quite some time. Everything else looks good.

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
 # Your answers here.
+
+1. It would take 255 or 256 seconds (depending on whether or not you count
+adding the first value to 0 as a second or not) 
+
```
abhishekmalali: The answer to this is 255 seconds. Since it would make more sense to add the first two numbers directly rather than add it to zero.

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
...
+2. We have that 256 / 8 = 32, so in just 32 seconds, each of the 8 workers 
+will have a number that represents the number of pennies they counted
+indivudually. Then, assuming that communication takes no time, in 1 second we 
+could take the 8 numbers and make them into 4, by assigning 4 people 2 numbers 
+to add each. Then, in another second, we could turn 4 numbers into 2, and 
+then 2 numbers into 1. Therefore, in total, this process takes 35 seconds. 
+
```
abhishekmalali: Again you assumed that it takes N seconds to count N bags. But it should be N-1 seconds to count N bags by my understand. Hence the answer to this should be 34 seconds. But the process you are using is perfectly right.

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
...
+4. As N grows large, we see that the amount of time for one person to 
+count grows linearly, while if parallelized with infinite number of cashiers
+grows logarithmically. See graph, where the red line is non-parallelized 
+and the green line is. 
+
+5. If it takes time to commnicate, then it takes around 2 * log_2(N) seconds 
+because at each step, in addition to adding two numbers, each cashier will 
+need to communicate and receive the second number from someone else. 
+
```
abhishekmalali: Except for the -1 which you didn't account for everything here seems okay.

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
...
+sum. Then finally, it takes 1 second to communicate between the 2 people who 
+were counting, and then 1 final second to sum. This makes a total of 260 seconds. 
+
+If we pass out 2 bags at a time, we again designate 2 people to be the counters. 
+Then, for every second after the first, one of the 2 people will receive a number, while 
+the other person will perform a sum. This means that in 2 seconds, we would have 
+processed 4 bags. Then, in 128 seconds, when the bags are all passed out, we are 
+very close to being done. It takes 5 more seconds to process the last pair of bags. 
+Thus, it takes 133 seconds total. 
+
```
abhishekmalali: I got the same answers. Sounds good. Followed the same procedure as you.

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
...
+If we could choose how to pass out the bags, we could do the following: 
+
+Give each person half of the number of bags you have remaining. And then each person
+with > 2 bags will also give out half of their bags at each second. This is essentially 
+parallelizing the process of giving N/2 people N bags each. This takes log_2(N) time 
+because at each second, the maximum number of bags any person holds on to is cut in half. 
+Therefore, after log_2(N) seconds, we have finished dishing out all the bags. Now, in 
+another 2 * log_2(N) seconds, we will have counted everything. This makes a grand total of 
+3 * log_2(N) seconds.
```
abhishekmalali: I used the same technique as well, The answer to this was 23 seconds for me. Good to see that I am not the only one who got these answers.

HW0/P6/P6.txt
```diff
@@ -1 +1,16 @@
...
+Part A: 
+
+Everytime we run this code, we get that the threads take slightly different times. 
+This is true because we expect that all 4 threads finish at the same time, so then 
+4 new items will be executed. However, we see that some of these 4 threads, even 
+though they all should take 0.25 seconds, finish just a tad faster than others. This 
+means that we should not use waiting times to try to make code synchronous. For example, 
+if we tried to run processes A1, A2, A3 and B1, B2, B3, where all the A's run before the
+B's. We should not just put them into a list in that order because we should make no 
+time guarantees that all the A's will finish before any B's will.
```
abhishekmalali: The gist of your answer seems the same to what I have written. A few use cases where this might fail includes serial execution. Lets say we want to process a lengthy time series where the time values are not independent and are functions of the events that have previously occurred. Also if we think of the Kalman filtering process, we won't be able to run in on the multiprocessing module since the updates depend on whether the previous updates have been executed.

HW0/P6/P6.txt
```diff
@@ -1 +1,16 @@
...
+though they all should take 0.25 seconds, finish just a tad faster than others. This 
+means that we should not use waiting times to try to make code synchronous. For example, 
+if we tried to run processes A1, A2, A3 and B1, B2, B3, where all the A's run before the
+B's. We should not just put them into a list in that order because we should make no 
+time guarantees that all the A's will finish before any B's will. 
+
+Part B: 
+Yes, it is possible because there is a fixed overhead of using the multiprocessing 
+library. As we can see from the graph, it seems like at 10^-5 seconds, we get the same 
+speed from series and parallel.
```
abhishekmalali: Another thing what you would observe is the overhead slowly becomes negligible and as the graph shows, the parallel execution becomes nearly 4 times faster than the serial execution. And as mentioned, the problem uses 4 threads which means that the longer the delay, the more the parallel program achieves better results since the overhead becomes negligible.

HW0/P6/P6B.py
```diff
@@ -16,16 +16,25 @@ def burnTime(t):
...
+        for _ in range(N): 
+            burnTime(t)
+        serialTime = time.time() - start 
 
         # Compute the ratio of these times
-        # ratio.append(serialTime/parallelTime)
+        ratio.append(serialTime/parallelTime)
 
     # Plot the results
     plt.plot(wait_time, ratio, '-ob')
```
abhishekmalali: Plot behaviour seems good and similar to mine.

HW0/P4/P4.py
```diff
@@ -40,37 +43,113 @@
...
+    s = np.matrix([0., 0., 2, 15, 3.5, 4.0]).T
+
+    xs = [s.item((0,0))] 
+    ys = [s.item((1,0))] 
+    zs = [s.item((2,0))]
+
+    ss = [s] 
+    for k in range(1, K):
+        s_pred = predictS(s)
+        assert s_pred.shape == (6,1)
```
georgeqi94: Whoa, this is pretty neat. Didn't know you could do that!

HW0/P5/P5.txt
```diff
@@ -1 +1,54 @@
...
+
+3. Assuming that it doesn't take any time just to type a number into the 
+calculator, every second we will cut the number of numbers left to add in 
+half. Therefore, we will finish in log_2(n) steps, where n is the number of 
+bags. In our case, log_2(256) = 8. 
+
+4. As N grows large, we see that the amount of time for one person to 
+count grows linearly, while if parallelized with infinite number of cashiers
+grows logarithmically. See graph, where the red line is non-parallelized 
+and the green line is.
```
georgeqi94: Got a similar graph. I think it might be helpful to set your y axis to a log scale in order to see what's really going on there.

HW0/P5/P5.txt
georgeqi94: Also got 255 -- way I saw it was we needed 255 comparisons (i.e. additions).

HW0/P5/P5.txt
georgeqi94: I agree here -- I also got 34 seconds. Each second, you're essentially reducing the problem by 8 bags, so after 31 seconds you can reduce it to 8 bags.

HW0/P5/P5.txt
georgeqi94: Got all the same answers! Liked the way you concisely explained everything and generalized cases. :+1:

HW0/P6/P6.txt
georgeqi94: Same as what I got here! As you've noted, the fixed overhead is what slows it down, but once the wait time increases, the power of multiprocessing really kicks in (and the overhead is not as significant), making it up to 4x as fast.

HW0/P5/P5.txt
oyuchoi: good logarithmic solution

PR #158, opened 2015-09-11
Hw0b

JocelynF:


zezhouliu:
Agree with the previous comments, good job.  Also good job to @aidizhang  and @patrickday9 for the very insightful and substantive comments !

HW0/P6/P6.txt
```diff
@@ -1 +1,8 @@
-# Your answers here
+P6A
```
zezhouliu: You should not be seeing speed-ups of 16x when you are using 4 processes.  This should cap at 4x

PR #159, opened 2015-09-11
[Serguei Balanovih] Hw0b

sbalanovich:
Submission of HW0 Part b. Might make some changes to P5.txt if I find time.


HW0/P5/P5.txt
```diff
@@ -1 +1,15 @@
...
+If each cashier takes 2 bags, the first step of the parallelization loses the costs of communication and bag handoff and we have 1 + 3*7 = 1 + 21 = 22 seconds to compute the bags. This is in addition to the 128 seconds required by you to hand off 256 bags in pairs, resulting in a total of 128 + 22 = 150 seconds total, which is much better.
+With 4 bags per cashier, each cashier in the first step has to perform 3 seconds of computation, but there are now only 6 steps of computation, communication, and bag handoff, resulting in 3 + 3*6 = 3 + 18 = 21 seconds of total computation in addition to 64 seconds of initial bag handoff. This means that starting with 4 bags allows for a total runtime of 85 seconds.
+At x bags, we have x-1 seconds for the first layer and then ceil(log_2(256/x)) more stages at 3 seconds apiece. The initial cost of bag handoff is 256/x, so the total time is (x-1) + 256/x + ceil(log_2(256/x)) seconds. Minimizing this for x yields x = 16 with total time being 35 seconds for the entire computation (A quick Wolfram Alpha query was used for this).
```
TuanTrain: For the first parts of question 6, I created a way for the cashiers (processes) to count the bags while you were handing them out, instead of sitting idly, which saved some time of the grand log_2(N) communication and computation. Also, I figured that after calculating a number, the number can be passed instead of the bag, which would reduce your 3*log_2(256) to 2*log_2(256). 
For the final part, I allowed the passing of the bags to be done in the same way as computation (log_2(N)), so it would reduce to the 3*log_2(256) you had earlier. The trick here is that you're not handing out bags to more than one person at a time, but there are exponentially more people handing out bags each iteration if giving half of them out each time.

HW0/P6/P6.txt
```diff
@@ -1 +1,131 @@
...
+It is possible for parallel threads to take longer than the serial executions. In this graph, we see that this happens around and below the 10^-5 mark of computation time. Thus, when the actual wait time per action is small enough, the overhead of distributing and recombining resources actually slows the parallelized solution down relative to the serialized one. This makes sense if we again consider the bags of coins problem in P5. There, we noticed that when the compute time at the lowest level was only 1 second, the parallelized solution spent more time handing out the bags to the cashiers than it would have taken to simply make a serial computation. In that problem, we found that at longer computation times, with the overhead of multithreading reduced, the parallel solution proved much more effective. This exact behavior can be seen through the P6.png graph included here.
```
TuanTrain: I like your examples relating back to P5. It is most clear when thinking about having 4 bags, handing 2 out to be computed, while you compute your 2 bags. That would take 1 hand out, 1 parallel compute, 1 communication back, and 1 communication to add the two sums, while adding the 4 numbers by themselves takes 3 seconds. The overhead starts to really get in the way of computing. However, it seems to still be the best strategy to parallelize when there are a larger amount of bags, so since the tradeoff occurs roughly when computing and parallizing are equal speeds, we can conclude from this exercise that the overhead to run in parallel takes about 10^(-5) seconds

HW0/P5/P5.txt
```diff
@@ -1 +1,15 @@
...
+
+3. With an infinite number of employees but a finite number of bags (256), only a finite subset of employees (256 of them) would actually be useful. If each was assigned a bag and the same strategy as above was used to pair them off and compute the sums in parallel, only log_2(256) = 8 seconds would be required to count up all the bags.
+
+4. Based on the above result, log_2(N) seconds would be required to sum all the N bags by an infinite number of employees. Be sure to round the number up to account for the extra step required if the total number of bags is not an exact power of 2. So ceil(log_2(N)) seconds are required. This is reflected in the included graph.
+
+5. At every step of the way, before every computation is made, one instance of communication must take place. So every summation will now take 2 seconds (1 for communication and 1 for summation) such that the total time will now be 2*log_2(256) = 2*8 = 16 seconds to compute all bags.
```
onefishy: But if communication can happen only one on one, and each cashier needs two numbers to perform a sum. Shouldn't it take three seconds (two for sending the numbers and one for adding)?

HW0/P5/P5.txt
```diff
@@ -1 +1,15 @@
...
+
+6. If each cashier takes 1 bag from you, an extra 256 seconds are required for each handoff. Additionally, on top of the summation and communication costs, each operation now has a bag handoff cost. This means that the total time to sum all the bags will be 3*log_2(256) = 3*8 = 24 seconds in addition to the 256 required to collect the bags from you. This results in 256 + 24 = 280 total seconds.
+If each cashier takes 2 bags, the first step of the parallelization loses the costs of communication and bag handoff and we have 1 + 3*7 = 1 + 21 = 22 seconds to compute the bags. This is in addition to the 128 seconds required by you to hand off 256 bags in pairs, resulting in a total of 128 + 22 = 150 seconds total, which is much better.
```
onefishy: When you say the computation takes 1+3*7 seconds you assume that employees with bags at each stage compute their totals at the same time right? But the person who gets the first pair of bags can start adding them while you're handing out the second pair. Would this make a difference?

adisuissa: Indeed there's no need to wait until all cashiers receive their bags.

PR #160, opened 2015-09-11
HW0b

stephenshifang:


HW0/P4/P4.py
```diff
@@ -30,47 +30,106 @@
...
-    # A = ?
-    # a = ?
-    # s = ?
+    # Construct matrix A from blocks
+    sub_a = np.eye(3)
+    sub_b = np.eye(3) * dt
+    sub_c = np.zeros((3,3))
+    sub_d = np.eye(3) * (1-c*dt)
+    A = np.bmat([[sub_a, sub_b], [sub_c, sub_d]])
+    A = np.array(A)
```
dominedo: This is trivial, but the np.diag() function that you use below could be a way of achieving the same result with fewer lines of code.

HW0/P4/P4.py
```diff
@@ -30,47 +30,106 @@
...
+        '''Use Eq(2) to compute intermediate s'''
+        return np.dot(A,s) + a
+
+    def predictSig(A, B, Sig):
+        '''
+        Use Eq(3) to compute cov mat.
+        Returns an array.
+        '''
+        # Make sure inputs are matrices
+        A = np.asmatrix(A)
```
dominedo: I would suggest setting up (or converting) all your variables as matrices at the beginning of part 4, running all your calculations on matrices, and then just converting your final result back to an array for graphing purposes.  That way you can avoid having to convert back and forth at each iteration. It probably doesn't make a noticeable difference in run-time for just 120 steps, but it could help speed things up if you were calculating more time steps.

HW0/P5/P5.txt
```diff
@@ -1 +1,46 @@
...
+4.
+If we keep splitting the bags so that each cashier only needs to add 2 numbers during each iteration, total time needed = ceiling(log2(N)).
+
+5.
+Considering overhead associated with people communicating, 
+each split and count would need 1 second of adding and 1 second of communication overhead except for the final count.
+So total time needed = log2(256)*2-1 = 15 seconds
+
+6.
+If each cashier comes to me and takes 1 bag, then it will take 256 seconds to distribute the bags.
```
dominedo: I believe that you can obtain a faster result by allowing the cashiers to start summing up their totals while the first person is still distributing the bags.  That way, some of the log-2(N) steps can occur in parallel with the initial N distribution steps.  By my calculations, that brings the total for the first part down from 272 to 260 seconds (not a huge improvement, but everything helps!).

stephenshifang: That's a good point.

HW0/P4/P4.py
ytang14: I used np.eye as well, its not a big deal

HW0/P4/P4.py
ytang14: I was confused by the comment style. I personally liked to made a function depiction above each function and use the ''' ...''' comment style only for debugging

HW0/P5/P5.txt
ytang14: As stated by dominedo

adisuissa: Indeed, you can overlap communication and computation, and get a better result by doing that (don't need to wait until all workers receive all bags).

PR #161, opened 2015-09-12
Hw0b

Chongmoliu:


HW0/P5/P5.txt
```diff
@@ -1 +1,6 @@
 # Your answers here.
+1. 255 
+2. yes; There will be 8 employees each counting 32 bags at the same time. 34s(31+log2(8))
+3. 8s(2^8=256)
+5. 15s
+6. 260;133;yes, for example, firstly give 16 cashiers each 16 bags.
```
serenabooth: I feel like an explanation or mathematical expression for this. I came up with a slightly different answer.

HW0/P4/P4.py
```diff
@@ -49,28 +62,64 @@
...
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    S4=np.zeros([6,K])
+    S4[:,0]=s
+
+    sk=s
+    Sigmak=np.eye(6)
+
+    for i in xrange(1,K):
```
serenabooth: Code would be much nicer if these were functions.

HW0/P6/P6B.py
```diff
@@ -16,17 +25,29 @@ def burnTime(t):
...
+    
+    def parallel_Time(t):
+        start_time2 = time.time()
+        pool.map(burnTime, [t]*N)
+        end_time2=time.time()
+        parallelTime = end_time2- start_time2
+        return parallelTime      
 
+    # Compute the ratio of these times
+    ratio=[serial_Time(t)/parallel_Time(t) for t in  wait_time]
```
serenabooth: Nice! I kind of prefer this without the outer loop given in the spec.

haosutang:
Hi! I'm assigned to review your work, so I'll post some comments here.
Part 4): it may be better to break down some codes into functions, which offers a easier understandability. A small trick is np.transpose(A) can be written as A.T, may save you some typing.
Part 5) I got similar result as you have.
Part 6) Good work!

PR #164, opened 2015-09-12
Hw0b

TuanTrain:
HW0 complete


HW0/P4/P4.py
```diff
@@ -30,47 +30,116 @@
...
 
     #####################
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
     #####################
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '.g', label='Observed trajectory')
+    x_coords, y_coords, z_coords = np.loadtxt('P4_measurements.txt', delimiter=',',unpack=True)
```
lhoangcse: what does unpack do?

HW0/P4/P4.py
```diff
@@ -30,47 +30,116 @@
...
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-k', label='Blind trajectory')
+    # new matrix vectors
+    x_coords2 = [s.item((0,0))]
+    y_coords2 = [s.item((1,0))]
+    z_coords2 = [s.item((2,0))]
+
+    # iterate and update for x,y,z vals
+    prev = s
+    for _ in range(K-1):
```
lhoangcse: this is cool, didn't know you can use _ for dummy variables

HW0/P5/P5.txt
```diff
@@ -1 +1,48 @@
 # Your answers here.
+(let's say putting the first number in the calculator takes a second ie 0 + first number)
+
+1. 256 bags / 1 cashier = 256 calculations = 256 seconds
```
lhoangcse: I think this should be 255 since that's the number of pairs to add.

HW0/P5/P5.txt
```diff
@@ -1 +1,48 @@
...
+
+2. 256 bags / 8 cashiers = 32 calculations / cashier = 32 seconds
```
lhoangcse: Similarly here, I think it should be 31 seconds since each cashier has 32 bags to add, so the number of pairs is 32 - 1.

HW0/P5/P5.txt
```diff
@@ -1 +1,48 @@
...
+to the counter, 3: the counter adding the numbers, 4: the two counters communicating, 5: adding the two
+sums. In total, it will take 133 seconds by handing out two bags at a time.
+
+The best strategy is to parallelize giving out the bags, so you give out half your bags,
+so you and the other person have 128, then you both give half of your bags, so 
+four people have 64 bags, and so on, so it takes log_2(N) time to hand out the bags,
+plus the 2*log_2(N) for communication and counting, so the result is the counting is done
+in 3*log_2(N) seconds. (plus 1 if you include handing the last 2 bags to a cashier, since you
+can't count them  yourself). This way, it will take 24 seconds to count 256 bags. 
+(25 with the technicality).
```
lhoangcse: nice solution. I had the same approach as yours.

HW0/P6/P6.txt
```diff
@@ -1 +1,21 @@
...
+can change position arbitrarily, so it is impossible to tell exactly which process will
+start or end first if both are sent at basically the same time. This can affect how we
+program in parallel by making us consider these race conditions and making sure that
+all of the pieces that we need are in place before continuing, not just one of the process
+results we expect is the last. 
+One scenario where this is important is in the stock exchange, where two people can submit
+requests to buy or sell at the same time, but given the nature of computation, one person can
+get one price and the next person can get a larger or smaller price if the market has changed in
+that split second. Also, if there is a shortage of buyers or sellers, one person can have all of 
+their stock bought or sold before the other, though both sent in their request at the same time.
```
lhoangcse: Nice example, and I liked the fact that you mentioned race conditions.

HW0/P4/P4.py
```diff
@@ -30,47 +30,116 @@
...
+             '.g', label='Observed trajectory')
 
     #####################
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
-    # s = ?
```
drishi: Incredibly minor style point but I think we were supposed to uncomment the vars above and use those lines to initialize (aka what you did on the next few lines). Again, doesn't change anything, just may decrease the amount of unnecessary comments in the code.

HW0/P4/P4.py
drishi: Agreed, didn't know that either! Is _ just treated like any other char (ie is that the same as for i in range(K-1)) or is it different at all?

HW0/P4/P4.py
```diff
@@ -30,47 +30,116 @@
...
+        # add to list
+        x.append(s.item((0,0)))
+        y.append(s.item((1,0)))
+        z.append(s.item((2,0)))
+
+    ax.plot(x,y,z,
+            '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
```
drishi: Hmm did the legend not show up in the plot before? Should be a way to do that so it doesn't have to be included in the axis title.

HW0/P5/P5.txt
drishi: I agree, I believe it is 255 for the reason mentioned above.

HW0/P5/P5.txt
drishi: I got 34 seconds here, based on:
+16s to get 256 bags -> 128
+8s to get 128 bags -> 64
+4s to get 128 bags -> 32
+2s to get 32 bags -> 16
+1s to get 16 bags -> 8
+1s to get 8 bags -> 4
+1s to get 4 bags -> 2
+1s to get 2 bags -> answer
+
+= 34 seconds

HW0/P5/P5.txt
drishi: Had different answers from you on the ones indicated and on #5, where I thought it would take 7 extra seconds because in each round we will need to take an extra second to talk to another person (except the last) giving me 15 s. Nice strategy for Number 6!

HW0/P6/P6.txt
drishi: Good answer, agree with it! I think we were supposed to include sample output here right?

HW0/P6/P6B.py
```diff
@@ -16,16 +16,23 @@ def burnTime(t):
...
+        for _ in range(N):
+            burnTime(t)
+        serialTime = time.time()-start2
 
         # Compute the ratio of these times
-        # ratio.append(serialTime/parallelTime)
+        ratio.append(serialTime/parallelTime)
 
     # Plot the results
     plt.plot(wait_time, ratio, '-ob')
```
drishi: More or less exactly how I did it so looks all good to me!

evanyao27:
HI ANTUAN BYE ANTUAN 

This message will be deleted.

HW0/P5/P5.txt
oyuchoi: great solution

PR #165, opened 2015-09-12
Hw0b

patrickday9:


patrickday9:
Submitting HW0b.

HW0/P4/P4.py
```diff
@@ -30,47 +30,123 @@
...
+    s = np.zeros(shape=(6,K)) 
+    s[:,0] = s0
+    
     # Compute the rest of sk using Eq (1)
+    for j in xrange(K-1):
+        s[:,j+1] = np.dot(A, s[:,j]) + a
+
+    xb_coords = s[:,0:-1][0]
+    yb_coords = s[:,0:-1][1]
+    zb_coords = s[:,0:-1][2]
```
JocelynF: You could just do xb_coords = s[0 , :] to get the same result. Basically just taking the whole first row. Not a big deal, just a bit cleaner.

HW0/P4/P4.py
```diff
@@ -30,47 +30,123 @@
...
+        s_k_1 = updateS(Sig_k_1, Sig_tilde, s_tilde, measured_coord[j])
+        
+        # Save the answer
+        sk[:,j+1] = s_k_1
+        
+    x_kf_coords = sk[:,0:-1][0]
+    y_kf_coords = sk[:,0:-1][1]
+    z_kf_coords = sk[:,0:-1][2]    
+
+    ax.plot(x_kf_coords, y_kf_coords, z_kf_coords, '-r', label='Filtered trajectory')
```
xdai1091: Your code is pretty straightforward and easy-read. Well done

JocelynF: I agree. Very easy to follow.

HW0/P5/P5.txt
```diff
@@ -1 +1,41 @@
...
+((B/E)−1) + log_2(E)
+and plugging in the parameters for this problem, the answer is:
+(256/8−1) + log2(8) = 34 seconds
+
+3. Imagine the store is big. . . very big. As a matter of fact, the store has an infinite number of employees. How fast can the employees verify the sum?
+The store is very lucky to have an unlimited amount of employees. However, for the purpose of our problem, we will only need 256/2=128 employees in order to keep them all productive (as employees can only had 2 numbers at a time). So given that we have the maximum number of employees counting bags at the same rate of 2 bags per second, then we would arrive at the following:
+(256/128−1) + log_2(128) = 8 seconds
```
xdai1091: Great explanation.

HW0/P5/P5.txt
```diff
@@ -1 +1,41 @@
...
+
+In order to implement the best counting strategy, I started by splitting the number of bags evenly amongst the employees, or B/P = 256/8 = 32 bags per employee. Now the employees will have to count quite a bit of bags but once again this can happen in the parallel cascading manner that I previously discribed. This method minimizes the communication overhead and after some more T&E calculations, I arrived at the following algorithm: (B/E) - 1 + E which leads to 256/8 - 1 + 8  = 32 - 1 + 8 = 39 seconds.
```
xdai1091: In this question, I agree that we need to consider the fact that one employee can start summing first. But I don't quite agree with we need to split the number of bags evenly. The goal here is to make them finish counting at the same time, so if we hand "n" bags to the first person, we just need to "n-1" to the second one, then they could finish it at the same time. Does this make sense?

HW0/P6/P6.txt
```diff
@@ -1 +1,38 @@
...
+Bye Job 5
+Hi Job 8
+Hi Job 9
+Bye Job 6
+Bye Job 7
+Bye Job 8
+Bye Job 9
+[0, 1, 2, 3, 4, 5, 6, 7, 8, 9]
+
+The output from our program shows that when running tasks in parallel, our computer doesn't necessarily complete the tasks in the order that we would expect. One would expect for the program to wake 4 processes, then sleep those 4 processes before starting the next batch of processes. What happens in practice is that the program actually wakes new processes before closing the previous ones. This is worrisome when programming in parallel  as we must be concerned about data dependancies. Any process that is waiting on another process to complete can be potentially overwritten or completed out of order
```
xdai1091: Nice, I didn't know this before.

HW0/P5/P5.txt
JocelynF: I think the person can't count while also getting bags handed to them. If that were true you'd only need one person (for the one bag at a time problem) and it would take 256 +1 seconds. So if you handed "n" bags to the first person and "n-1" to the second person, the first person would finish counting when the second person finished receiving bags, but the second person would still have to count so they wouldn't finish at the same time.

HW0/P5/P5.txt
```diff
@@ -1 +1,41 @@
...
+
+In order to verify the 256 bags worth coins, given that it takes exactly 1 second to hand a bag of coins to a cashier we need to examine how long it takes to hand all of the bags to the employees. In the case of 8 employees working, it will take exactly 8 seconds to hand over the first round of bags, then an additional 8 seconds to hand of the next round and so on and so forth. The trick here is that one employee can start summing their bags as you hand of bags to the next employee. So to hand over 2 rounds of bags, and to allow time for all employees to add their bags, it will take 8+8+1=17 seconds. If we apply this logic to all bags, it takes 17 seconds for 2 rounds or 16 bags, it will take 16 rounds of 16 bags to complete the task or 16*16+1 = 257. Now for the communication costs (log_2(E)) and the additional time needed to add (log_2(E)) we add the 2 * log_2(E) factor and arrive at 257 + 2 * log_2(8) = 256+(2*3) = 263.
+
```
JocelynF: If I'm thinking about this problem correctly, the previous employees can be adding and communicating while the next employees are receiving bags so once the final bag is handed out the last employee takes one more second to add, one second to communicate with the others, and one second to add their total to theirs. So in the end it should take 256+3 seconds.

HW0/P5/P5.txt
```diff
@@ -1 +1,41 @@
...
+Now we can assume that it takes the same amount of time to hand of 2 bags as it does to hand off 1 bag. This means that employees can start summing their bags as I start to hand of bags to the next cashier which should result in reduced verification time. If we once again have 8 employees and 256 bags, I will hand 2 bags to the first employee and while I hand over the next 2 bags, the first employee will start summing. Now it will take B/2 time for each employee to count his or her bag. So for 256/2 = 128 seconds for each employee to count and since the employees are counting in a cascading manner, we must add 1 additional second to account for the time that the last employee needs to complete counting. Adding in our communication and employee constant of 2*log_2(E) arriving at the final answer of (256/2) + 1 + 2 * log2(8) = 128 + 2 + 2*3 = 135 seconds.
+
```
JocelynF: Again using the same idea I mentioned above, I think the previous employees can count and communicate while the next employees are receiving bags so I think after the last employee receives his bags it takes 7 seconds (if he can count by 2s) for him to count, 1 second to communicate, and 1 second to add. So it could be 128+9 = 137.

HW0/P5/P5.txt
```diff
@@ -1 +1,41 @@
 # Your answers here.
```
JocelynF: Really great explanations in this problem. Helped me understand where I had originally gone wrong in my HW0.

HW0/P6/P6.txt
```diff
@@ -1 +1,38 @@
...
+Imagine if we are trying to correctly sum the bags from P4 and determine which cashier combinations give the fastest results. If this is broken up amongst multiple processors and completed out of order, we could lose count and confuse our cashiers!
+
+
+Problem 6B
+Try to explain the trend you observe. Is it possible that a parallel program could take longer than it’s serial version? Under what conditions?
+
+In the Speedup versus function time plot, I see that the ratio significantly increases around 10^-4. This seems to be the sweet spot when the program really starts to take advantage of parallelism. After this quick increase in the rain there is a plateau around 10^-1. 
+
+Yes, it is possible as the ratio is smaller than one around 10^-5.5. It shows that parallel is not always better! I can imagine a scenario where there is a lot of data to move between processors and the computation is fairly simple. This would mean that it takes more time to move the data than to complete the compute task.
+
```
JocelynF: This problem looks good. Again, good explanations.

HW0/P5/P5.txt
patrickday9: Great, it's good to know that my long explanations were helpful... Thanks for the feedback!

HW0/P5/P5.txt
patrickday9: Interesting, now that I read through the problem again and think about it, it seems that you are right in that it will take more than 6 additional seconds for the last employee to add his/her bags. I think your answer may be correct, I hope that they release the solutions soon!

HW0/P5/P5.txt
patrickday9: Yes, good catch! I originally thought that the optimal way to attack this problem was to offload as many bags as possible to the employees but upon reading your explanation, I think you are correct. We want to keep employees busy at all times so no one is idle and in my "best strategy" many employees will be done while the last employee is still counting. This was a fun question, thanks for the helpful feedback!

HW0/P4/P4.py
adisuissa: I agree that the code is easy to follow.
It seems that we get different results in the predictSig function, but not highly important (might be the use of pinv instead of inv).

PR #166, opened 2015-09-12
Hw0b

rohitahuja:


HW0/P6/P6B.py
```diff
@@ -16,16 +17,27 @@ def burnTime(t):
...
 
     for t in wait_time:
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
+
+        # Serial
+        start = time.time()
+        pool.map(burnTime(t), range(N))
```
kevinkeyjkw: I may be wrong, but shouldn't the second parameter to map() be the values burnTime is operating on? So what I did was pool.map(burnTime,[t,t,t,t,t,t,t,t,t,t,t,t,t,t,t]) where t is the wait time, and I would do this for each wait time.

HW0/P5/P5.txt
```diff
@@ -1 +1,21 @@
...
+
+3) 8 seconds. We can sum the totals using a divide and conquer approach. In the first iteration, we have 128 cashiers summing 128 pairs (the 256 totals has 128 pairs), and this takes 1 second. In the second iteration, 64 cashiers sum the 64 resulting pairs, which takes another second. Following this pattern, it takes 8 seconds to sum the 256 totals.
+
+4) log_2 (N) seconds. Generalizing the strategy in question 3, summing the totals using divide and conquer effectively takes log base 2 of N operations since we can have an infinite number of cashiers reduce the number of totals by half in each iteration. Since each operation takes 1 second, this takes log_2 (N) seconds.
```
danfortunato: If N is not a power of 2, then it would actually take ceil(log2(N)) seconds.

HW0/P5/P5.txt
```diff
@@ -1 +1,21 @@
...
+
+Master strategy:
+23 seconds. If we are not fixed in the number of bags that we hand to a cashier in one second, then the best strategy would be as follows: In the first second, hand all the bags to a single cashier. In the second second, that cashier hands half of their bags to another cashier. In the third second, those cashiers hand half of their bags to two other cashiers. Continue this until each cashier involved has two bags each (overall, this takes 8 seconds). It takes 1 second to sum these 128 pairs. It takes two seconds to communicate and sum these 64 pairs. Following this pattern, we take 8 + 1 + 2*log_2(128) = 9 + 14 = 23 seconds.
```
danfortunato: Nice! I hadn't thought of changing the number of bags distributed at each iteration. Well done.

HW0/P6/P6B.py
danfortunato: @kevinkeyjkw That is what I did as well. The way it is currently written, `pool.map(burnTime(t), range(N))` actually computes `burnTime(t)` *once* before `pool.map` is called, effectively doing something like `pool.map(1.0, range(N))`, I believe. So I think your scaling plot is wrong, and it should level out around 4 (since there are 4 processes in your pool). Also, your `# Parallel` and `# Serial` comments are switched. :)

HW0/P5/P5.txt
adisuissa: Good job!

PR #170, opened 2015-09-12
Hw0b

dominedo:


HW0/P4/P4.py
```diff
@@ -40,38 +59,115 @@
...
-    # s = ?
+    # Initialize progation matrix
+    A = np.mat([
+        [1, 0, 0, dt, 0, 0],
+        [0, 1, 0, 0, dt, 0],
+        [0, 0, 1, 0, 0, dt],
+        [0, 0, 0, 1-c*dt, 0, 0],
+        [0, 0, 0, 0, 1-c*dt, 0],
+        [0, 0, 0, 0, 0, 1-c*dt]
+        ])
```
jack91an: Instead of hard coding, you could also construct two arrays and use the np.diag() function to insert the arrays into a zero matrix. This should help if we are recording more parameters and have a huge A matrix. (This comment also applies to later matrix definitions, but I think it's a trivial point)

HW0/P4/P4.py
```diff
@@ -2,6 +2,21 @@
...
+def predictSig(sigma):
+    return np.linalg.inv(A * sigma * A.transpose() + B * B.transpose())
+
+# Equation 4 from the instructions
+def updateSig(sigma_pred):
+    return np.linalg.inv(sigma_pred + C.transpose() * C)
+
+# Equation 5 from instructions
+def updateS(s_kal, s_obs, sigma_pred, sigma):
+    return (sigma * (sigma_pred * s_kal + C.transpose() * s_obs))
```
jack91an: I like the way you define all the functions before the body.

HW0/P5/P5.txt
```diff
@@ -1 +1,108 @@
...
+				d		d		d		d		d		d
+
+After second 5, a stationary set of operations develops.  Since all bags have to be distributed, at least 256 steps must be taken.  At the 256th second, the following occurs:
+
+256		257		258		259		260
+---		---		---		---		---
+s		c		s		c		s
+c		s		s
+d
+
```
jack91an: Nice observation of the stationary property after the fifth person! I myself was having problem coming up with the exact time needed and only calculated the theoretical lower bound of 258 seconds. I think there might be better summing/communication strategies that could take down another second or two.

HW0/P4/P4.py
dominedo: Thanks! I wasn't familiar with np.diag() but will have to remember that one for future problems.

HW0/P5/P5.txt
```diff
@@ -1 +1,108 @@
...
+Second 1: Distribute 128 bags to cashier #1.
+Second 2: Distribute 128 bags to cashier #2. Cashier #1 begins to sum up his bags.
+Second 3: Cashier #2 begins to sum up his bags.
+As per part 1, it will take each cashier 127 steps to sum up his bags.
+Second 128: Cashier #1 finishes summing his bags. He waits because no actions are available.
+Second 129: Cashier #2 finishes summing his bags.
+Second 130: Cashier #2 communicates his total to cashier #1.
+Second 131: Cashier #1 sums the two numbers.
+
```
jack91an: A better strategy indeed!
The strategy I proposed was to distribute 16 bags to 16 cashiers, so it will take much less time for each cashier to reduce their bags to 1, and then they start to communicate to each other (the whole process takes 30+ seconds). I think there's a fine balance between workers vs. communication overhead here, which is introduced in class in the past few days :)

HW0/P4/P4.py
jack91an: Thanks for the reply! Apologies for possibly spamming your inbox with notifications- made a few comments along the way and hopefully there’s some setting to turn the emails off.

HW0/P5/P5.txt
zhiqianseah: Your Qn 6, Scenario 1 solution is much more efficient than mine. I just left the cashiers standing there dumbly with the bags. kudos!

HW0/P5/P5.txt
zhiqianseah: Giving your two perspectives to consider.
My strategy was to give decreasing number of bags to cashiers at the back of the queue, so that the cashiers at the front can start summing bags while you are handling out bags. 
1st second: 22 bags to cashier 1
2nd second: 21 bags to cashier 2, cashier 1 starts counting (it will take him 21 seconds)
3rd second: 20 bags to cashier 3, cashier 2 starts counting (it will take him 20 seconds)
This method will take 33 seconds to finish (i think it is 2 seconds more optimal than jack's solution)

The solution that I got as a comment is much better:
1st second: 256 bags to cashier 1
2nd second: cashier 1 gives 128 bags to cashier 2
3rd second: cashier 1 and cashier 2 give 64 bags to cashier 3 and cashier 4
4th second: cashier 1, 2, 3, 4 give 32 bags to cashier 5, 6, 7, 8
...

Essentially you are outsourcing the bags distributions to the cashiers, who can distribute the bags out much more efficiently than you. This method takes only 22 seconds!

HW0/P4/P4.py
```diff
@@ -40,38 +59,115 @@
...
+    s_kal = np.array(s_kal)
+    
+    # Plot filtered trajectory
+    ax.plot(s_kal[0,:], s_kal[1,:], s_kal[2,:],
+            '-r', label='Filtered trajectory')
 
     # Show the plot
     ax.legend()
-    plt.show()
+    plt.show()
```
zhiqianseah: Unable to comment on the graph itself, so I will write it here. but I think it is better to have both the lone cashier and infinite cashier having the same y-axis values. The graph below gives the first impression that the infinite cashier takes a longer time than the lone cashier

HW0/P6/P6.txt
```diff
@@ -1 +1,41 @@
...
+
+One scenario where this type of behavior would be important is the case where subsequent jobs require inputs from earlier jobs. Such dependencies could result in bottlenecks, where processors are idle while waiting from other jobs to finish running. If not handled correctly, this situation could also result in later jobs using the wrong inputs (for example, if they are not able to detect that the earlier job has not finished running and use garbage values as input). This problem could be amplified in cases where larger jobs are unavoidable - for example, to minimize or offset the overall time associated with transferring a large dataset that is necessary for each job - as they could end up being run even more out of order.
```
zhiqianseah: I think that having the bottlenecks is inevitable if you want to ensure that later jobs is using the right values. This is the concept of Locks on shared data, so that only 1 thread can access a shared piece of data at any time. Bottlenecks are much better than incorrect input values!

HW0/P6/P6.txt
```diff
@@ -1 +1,41 @@
...
+
+This appears to be 'normal' behavior rather than an outlier; in fact, the code had to be run 15 times before getting an output where all 10 jobs ended in the same order that they had started. This implies that it is difficult to guarantee consistent runtimes across individual jobs, even for very simple jobs like this one, and that it is likely to be even harder with larger jobs.
+
+To test this, I increased the sleep time from 0.25 to 1.0.  As expected, this resulted in the jobs being even more out of order with each other: in a few cases, the second or third batches of jobs started running before all the jobs in the previous batch had finished running, and in once instance even the first batch of jobs started out of order. In fact, out of 30 total runs, the jobs did not once end in the same order that they had started. These results indicate that smaller job sizes are necessary when attempting to maintain consistent run times across jobs and/or ordered jobs.
```
zhiqianseah: Good experiment! However, I don't think one should make jobs smaller just to maintain consistent running order. A parallel program is not meant to be running in order! Instead, we just need to ensure that input values to subsequent functions is valid, as you have pointed out in the next paragraph

HW0/P6/P6B.py
```diff
@@ -14,18 +14,27 @@ def burnTime(t):
...
 
-    # Use a variety of wait times
-    ratio = []
-    wait_time = []
-
+    # Initialize list for serial vs parallel ratios
+    ratio = []    
+    
+    # Range of wait times between 10^-6 and 10^0
+    wait_time = [y*10**x for x in range(-6,0) for y in range (1,10)] + [1]
```
zhiqianseah: you can use logspace to get evenly spaced data points on a log graph

dominedo: Thanks!  I'll keep that in mind for future problems.

HW0/P5/P5.py
```diff
@@ -0,0 +1,28 @@
...
+    axis2 = axis1.twinx()
+    
+    axis1.plot(bags, lone_cashier, '--b', label='Lone cashier')
+    axis2.plot(bags, inf_cashiers, '-r', label='Infinite cashiers')
+    
+    axis1.set_xlabel('Number of bags')
+    axis1.set_ylabel(ylabel='Lone cashier time (seconds)', color='blue')
+    axis2.set_ylabel(ylabel='Infinite cashier time (seconds)', color='red')
+    
+    plt.show()
```
adisuissa: Using the 2 Y-axis graph method makes the graph harder to read (although a bit more precise). I guess the original intention was to compare the two methods, thus a single Y-axis is better.

PR #180, opened 2015-09-12
Finished HW0

xdai1091:
Completed for all parts of HW0


HW0/P4/P4.py
```diff
@@ -1,6 +1,27 @@
 import numpy as np
 import matplotlib.pyplot as plt
 from mpl_toolkits.mplot3d import Axes3D
+from numpy.linalg import inv
+
+
+# Function to calculate predictS
+def predictS(s_p):
```
kandluis: For all these functions, I see the rational behind not passing in the A, a, etc. parameters (it makes the signature much messier).

However, you're relying on the fact that these functions will be called only after the above are defined appropriately. It seems like asking for error/problems. I prefer to encapsulate everything as tightly as I can :)

HW0/P4/P4.py
```diff
@@ -30,47 +51,105 @@
     # Load true trajectory and plot it
     # Normally, this data wouldn't be available in the real world
     #####################
-
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '--b', label='True trajectory')
+    # Read the file
+    trueFile = open("P4_trajectory.txt","r"); 
+    x_coords, y_coords, z_coords = np.loadtxt(trueFile,delimiter =',',usecols = (0,1,2),unpack = True);
```
kandluis: Very clever! I did not know about the usecols parameter. It looks super useful.

HW0/P4/P4.py
```diff
@@ -30,47 +51,105 @@
...
     # Part 4:
     # Use the Kalman filter for prediction
     #####################
-
-    # B = ?
-    # C = ?
+    # Create the B Matrix
+    B = np.identity(6);
+    b = np.matrix([[bx,by,bz,bvx,bvy,bvz]]);
+    B = b * B;
```
kandluis: Did you try looking into np.diag() function :)

HW0/P4/P4.py
```diff
@@ -30,47 +51,105 @@
...
+
+    # Create the C Matrix ( 3 * 6)
+    C = np.zeros(shape = (3,6));
```
kandluis: I like this way of filling the C matrix. I hard coded my values because I didn't think about doing this.

HW0/P5/P5.txt
```diff
@@ -1 +1,18 @@
...
+
+Q2: (34 secs) We have 8 people. Everyone takes 32 nags and in 31 seconds, all bags will be counted. Then it will take three seconds to add them up.
+
+Q3: (8 secs). If we have infinite people, let assume everyone takes two bags, then everyone takes two numbers to add up. It will take 8 seconds to finish counting. 
+
+Q4: For multiple persons, the trends would equals to log2(bags_numbers) while the single person equals to n-1
+
+Q5: (15 Seconds), in this case, if two people talk, we will add 1 sec. Therefore, 7 seconds will be added. 
+
+Q6: (272 s, 270s) If we gave bags to 256 people, it will take 256 s for handling, 8 secs to communicate each other and calculating. The total will be 272 secs. The same thing for two bags each person, it will take 270 s.
```
kandluis: I agree with your logic (handing, calculating), but I think you're not considering the fact that the communication can happen at the same time as we hand out the bags (ie, give the first four bags out, then after two seconds, those four people have summed to a value). Does that make sense?

HW0/P6/P6B.py
```diff
@@ -16,18 +17,33 @@ def burnTime(t):
...
+    print wait_time;
 
     for t in wait_time:
         # Compute jobs serially and in parallel
         # Use time.time() to compute the elapsed time for each
-        serialTime = 1
-        parallelTime = 1
+        # Compute serial time
+        start = time.time();
+        for i in range(0,16):
```
kandluis: I would use N instead of 16 here -- we've already defined it, so no need to have random "magic" constants.

xdai1091: Thanks! Useful

HW0/P4/P4.py
adisuissa: It seems like B is a vector and not a matrix. This has an impact on the computation of the kalman filter solution.

HW0/P5/P5.py
```diff
@@ -0,0 +1,17 @@
...
+import numpy as np
+
+x = np.linspace(0.5,300,1000);
+
+y1 = [math.log(i,2) for i in x] 
+y2 = [i-1 for i in x] 
+
+p1, = plt.plot(x,y1,'b');
+p2, = plt.plot(x,y2,'r');
+plt.ylim([0,10]);
```
adisuissa: You can probably improve the graph by not limiting the Y axis (the idea is to show the general behavior of the methods, and not specific values).

HW0/P5/P5.txt
adisuissa: Indeed you can overlap communication and computation, and get a better result by doing that (don't need to wait until all workers receive all bags).

HW0/P6/P6.txt
```diff
@@ -1 +1,4 @@
-# Your answers here
+# Xingchi Dai
+Q1: One thing I noticed is the order of threads has been changed. The thread started first might not end first, although we have the same sleeping time. This will affect our parallel programming in a way we cannot predict the result, and will make one CPU idle. A very obvious scenario will be if multi-threads are related, since the order is not sure, one thread may wait for another or even the whole program will mess up.
```
adisuissa: Processes instead of threads, but the same reasons apply here.

PR #188, opened 2015-09-18
Hw0b

nicodri:


nicodri:
Thanks for your review, the np.linalg.pinv() is effectively a good way to prevent a singular issue in the code. I forgot to save my plot, I've just commit it.

HW0/P4/P4.py
```diff
@@ -31,46 +32,77 @@
...
+
     # Compute the rest of sk using Eqs (2), (3), (4), and (5)
 
-    # ax.plot(x_coords, y_coords, z_coords,
-    #         '-r', label='Filtered trajectory')
+    def predictS(sk):
+        return A * sk + a
+
+    def predictSig(sigk):
+        return np.linalg.pinv(A * sigk * np.transpose(A) + B * np.transpose(B))
```
adisuissa: It seems that we get different results in the predictSig function, but not highly important (might be the use of pinv instead of inv).

HW0/P5/P5.txt
```diff
@@ -1 +1,51 @@
...
+then it keeps going until 128 secondes.
+129: c1 computes (c1: 1, c2:1, c3:0)
+130: c1 to c2 (c1: 0, c2:2, c3:0)
+131sec: c2 computes
+
+Total time here is 131s
+257: c1 to c2 (c1: 0, c2:2, c3:0)
+
+
+
```
adisuissa: What about a better strategy (if there are x bags that can be passed)?

PR #189, opened 2015-09-18
Hw0b

x75093:
HW0b Submission


HW0/P4/P4.py
```diff
@@ -29,46 +29,94 @@
     #
     # Load true trajectory and plot it
     # Normally, this data wouldn't be available in the real world
+
+    s_true = np.loadtxt("P4_trajectory.txt", delimiter = ',')
+
+    x_coords, y_coords, z_coords = s_true[:,0], s_true[:,1], s_true[:, 2]
+
     #####################
```
nicodri: Naming the coords seem not necessary here I think as you are only using this coords once

HW0/P4/P4.py
```diff
@@ -29,46 +29,94 @@
...
+
+    #plt.show()
 
     #####################
     # Part 2:
     #
     # Read the observation array and plot it (Part 2)
+    s_measure = np.loadtxt("P4_measurements.txt", delimiter = ',')
+
+    x_measure, y_measure, z_measure = s_measure[:,0]/rx, s_measure[:,1]/ry, s_measure[:, 2]/rz
```
nicodri: Same comment here :)

HW0/P4/P4.py
```diff
@@ -29,46 +29,94 @@
...
     #####################
     # Part 3:
     # Use the initial conditions and propagation matrix for prediction
     #####################
 
-    # A = ?
-    # a = ?
-    # s = ?
+    A = np.array([[1,0,0,dt,0,0],[0,1,0,0,dt,0],[0,0,1,0,0,dt],
+[0,0,0,1-c*dt,0,0],[0,0,0,0,1-c*dt,0],[0,0,0,0,0,1-c*dt]])
```
nicodri: If you are using Sublime, it may be helpful to install the package 'autopep8' to automatize the corrections related to the rules from pep8. It helps have a cleaner and more spread code.

HW0/P4/P4.py
```diff
@@ -29,46 +29,94 @@
...
+    file_ext = 'P4_measurements.txt'
+    m = np.loadtxt(file_ext, delimiter = ',')
+
+    for i in range(K-1):
+        stil_temp = predictS(sk1[:,i])
+        sigtil_temp = predictSig(sig0)
+        sig0 = sigtil_temp #update sig0 
+        sigk1_temp = updateSig(sigtil_temp)
+        s_temp = updateS(sigk1_temp, sigtil_temp, stil_temp, m[i])
+        sk1[:,i+1] = s_temp
```
nicodri: I think it may be wiser also here to minimize the number of local variable, you only need sk1 for the next iteration. s_temp and sigk1_temp are overloading your code. Try to figure out how many times you we'll need a variable: if only one, no need to store it, otherwise it may be useful to store it.

HW0/P5/P5.txt
```diff
@@ -1 +1,55 @@
...
+
+	For a two bag handoff, assuming 8 employees and 256 bags, it will first take (N / 2) seconds to count each round of bag handoff. We must add 2 seconds onto the final handoff number in order to account for the last employee to count his bag after the last round of handoffs. This leads us to (N / 2) + 1 seconds, plus the time to combine numbers which is 2 * log2(2). The complete calculation is (N / 2) + 2 * log2(employees) + 1 = 128 + 1 + 2 * 3 = 135 seconds. 
+
+	For a better strategy, assuming 8 employees and 256 bags, we should evenly distribute the bags amongst the employees, at 32 bags per employee. As we hand out the bags, employees will begin counting in parallel to the handout operation. This allows us to form the following general calculation: ( N / employees) - 1 + employees, or in this specfic case, (256 / 8) - 1 + 8 = 39 seconds.  
+
+
```
nicodri: We have a similar reasoning for this part so I guess we should be on the right path :)

HW0/README
```diff
@@ -0,0 +1 @@
+My name is PATRICK KUIPER
```
nicodri: Nice Scientist Name :)

HW0/P5/P5.txt
zezhouliu: For the last part, the thing is it takes a "turn"/second to pass the elements out.  So it you wouldn't be able to pass out all 32 to different cashiers in the same turn.

PR #196, opened 2015-09-24
Hw1

nmboffi:


thereisnoquarter:
finished problem 3

HW1/P2/P2b.py
```diff
@@ -1,3 +1,40 @@
...
+
+    # Get the i, j pixel arrays
+    xs = sc.parallelize(range(0, NX), 10)
+    ys = sc.parallelize(range(0, NY), 10)
+
+    # Take the cartesian product to get the pixel grid
+    # Now we partition into 100 partitions, and to determine where each element
+    # of the RDD goes, we simply assign a random number between 1 and 100 (inclusive)
+    pixels = xs.cartesian(ys).partitionBy(100, lambda p: randint(1, 101) )
+
```
andrewreece: Nice solution.  You might also consider [`.repartition(numPartitions)`](https://spark.apache.org/docs/1.2.0/programming-guide.html), which randomly assigns elements to `N = numPartitions` partitions.  Also, careful with `randint` - your upper bound here is 101, which is one too many.  You used the [`random` module's version](https://docs.python.org/2/library/random.html#random.randint
) of this function, which is upper bound inclusive, so your range includes 101.  ([Numpy's randint](http://docs.scipy.org/doc/numpy/reference/generated/numpy.random.randint.html), on the other hand, is upper bound exclusive.)

HW1/P3/P3.py
```diff
@@ -0,0 +1,27 @@
...
+    sorted_pair_words = pair_words.map(lambda (x, y): (''.join(sorted(x)), y))
+
+    # Reduce by key and merge lists
+    seqs_and_word_list = sorted_pair_words.reduceByKey(lambda x,y: x+y)
+
+    # Now add in the number of words
+    proper_format = seqs_and_word_list.map(lambda (x, y): (x, len(y), y))
+
+    # And lets find the max
+    max_finder = proper_format.map(lambda (x, y, z): (y, x, z)).max()
```
andrewreece: `takeOrdered()` is an alternative to `map().max()` here

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,163 @@
...
+        color = 'WHITE'
+    elif 'GRAY' in color_list:
+        # People for the next round
+        # Also means we SHOULDN'T stop yet
+        color = 'GRAY'
+        acc.add(1)
+    else:
+        # Still haven't gotten to this guy
+        color = 'BLACK'
+
```
andrewreece: The color scheme approach is really creative!

HW1/P6/P6.py
```diff
@@ -0,0 +1,219 @@
...
+    # findall returns a list, so...
+    return matches
+
+def filter_words(word):
+    """Function that uses regular expressions to filter out words that are made of
+    entirely capital letters, or entirely capital letters followed by a period, as per
+    the problem statement.
+    Note that the way we matched words, \w+, has already filtered out words made of only numbers.
+    """
+
```
andrewreece: Native Python string methods `.isupper()` and `.isdigit()` save having to go through regex...but either way works

HW1/P2/P2.txt
```diff
@@ -1 +1,16 @@
...
+take, for example, the case where we partition in even squares that tile the plane or by columns or rows - some partitions
+will necessarily be forced to do more work.
+
+On the other hand, it is clear that choosing a random partitioning strategy will lead to the highest initial shuffle cost. This
+is due to the fact that, no matter what partitioning scheme Spark chooses by default, there simply cannot be any relation
+between the final partitioning scheme and the original partitionig scheme. It is likely for this reason that we see that, despite
+the fact that the work looks significantly more spread out in the random case (see the plot - as expected, the work is spread out
+in what appears to be, to a very good approximation, a normal distribution), the computation takes an equal amount of time.
+
+I am unsure how to explain the spike of work at the tail of the distribution, but it consistently appears despite the rest of the structure remaining constant.
```
araftery: While it wouldn't eliminate the spike, you could make the histogram a little prettier by reducing the bin size.

HW1/P4/P4.py
```diff
@@ -0,0 +1,99 @@
...
+    # Now we access every single edge that appears in the graph
+    all_edges = issues_and_chars.flatMap(get_all_individual_pairs)
+
+    # Now each key is a character, and it appears many times -
+    # once for each edge. Consolidate all of these into an adjacency
+    # list representation of the graph.
+    # We are on 4 cores so we give it 16 partitions. We will try to preserve this partitioning
+    # as we progress. We did not worry about the partitioning scheme of the previous RDD's
+    # because they were so transient.
+    marvel_graph = all_edges.groupByKey().partitionBy(16)
```
araftery: Good job setting a sensible partition size and preserving it throughout the implementation! Reduces shuffle costs and definitely speeds up the code!

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # Zero off our accumulator
+    accum = sc.accumulator(0)
+
+    # Cache our graph to avoid unnecessary computation due to laziness 
+    dist_graph = dist_graph.cache()
+
+    # Arbitrarily set the target distance to a large number
+    # Also give us an array to hold the target path
+    target_d = 100000
```
araftery: Could use `np.inf` here.

HW1/P6/P6.py
araftery: I think some regex is necessary, since, for example:

```
In [1]: "TEST!".isupper()
Out[1]: True
```

This would imply the string should be excluded, but you actually want to keep it since it is not a word of entirely capital letters nor a word of entirely capital letters followed by a period.

HW1/P6/P6.py
```diff
@@ -0,0 +1,219 @@
...
+    poss_words = next_three[1]
+    
+    # And choose accordingly...
+    word_3 = weighted_choice(poss_words)
+    
+    # Add our words to the phrase...
+    phrase.append(word_1)
+    phrase.append(word_2)
+    phrase.append(word_3)
+
```
araftery: Why not just skip this part, set `next_pair` to a random pair of words, and let the loop run?

HW1/P6/P6.py
```diff
@@ -0,0 +1,219 @@
...
+    word_3 = weighted_choice(poss_words)
+    
+    # Add our words to the phrase...
+    phrase.append(word_1)
+    phrase.append(word_2)
+    phrase.append(word_3)
+    
+    # Make sure we stop at some point
+    curr_num_words += 3
+
```
araftery: Why not just skip this part, set `next_pair` to a random pair of words, and let the loop run?

HW1/P6/P6.py
```diff
@@ -0,0 +1,219 @@
...
+    
+    # And we are done!
+    return phrase
+
+def weighted_choice(choices):
+    """ Inspired by http://stackoverflow.com/questions/3679694/a-weighted-version-of-random-choice
+    Given a list of tuples, [(x1, c1), (x2, c2), ...]
+    Returns some xn from the list chosen randomly but weighted by the counts cn.
+    """
+
```
araftery: Another option here is to use `zip` to get the counts and then divide the counts by the length of the list, which gives you the probabilities you need.

HW1/P2/P2.txt
oyuchoi, line 16: Maybe random partitioning might result in a partially random tail?  There is a way that you can get ~same performance by choosing a deterministic procedure

PR #198, opened 2015-09-28
Hw1

ytang14:


HW1/P2/P2.txt
```diff
@@ -1 +1,7 @@
 # Your discussion here
+The result of P2a is reasonable. It failed to produce a very handy partition. Because for the 2000*2000 pixels, some of the pixel's greyscale value was calculated and drew many times, while some of them might never been touched. So it is a bad idea to partition them in the default way.
```
btweinstein: I don't think that the pixel values were "calculated and drew" many times...some pixels just required many more iterations than others and were drawn once. Every pixel was touched, or else we would not have created the mandelbrot set! I agree that it is a bad idea to partition in the default way.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,14 @@
...
+sc = pyspark.SparkContext(appName="Spark1")
+my_rdd = sc.parallelize(xrange(2000),10)
+rdd = my_rdd.cartesian(my_rdd)
+ans = rdd.map(lambda x: [x,mandelbrot((x[1]/500.0-2),(x[0]/500.0-2))])
+draw_image(ans)
 
-# Your code here
+ans1 = sum_values_for_partitions(ans)
+
+plt.hist(ans1.collect())
```
btweinstein: You should really make this histogram on a log-log scale. If you do that, you will see that work per partition varies over 3 orders of magnitude. You can't see that at all right now...it looks like all partitions complete O(10^7) iterations which is ok. A linear histogram does not  illustrate the disparity appropriately in my opinion.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,21 @@
...
+#y = sc.parallelize(xrange(2000),10)
+#rdd = x.cartesian(x).repartition(100)
+#rdd = x.cartesion(x).partitionBy(lambda x:randint(1,101))
+ans = rdd.map(lambda x: [x,mandelbrot((x[0]/500.0-2),(x[1]/500.0-2))])
+draw_image(ans)
 
-# Your code here
+ans1 = sum_values_for_partitions(ans)
+
+plt.hist(ans1.collect())
```
btweinstein: I can't read the x axis on your histogram...what are the values there?

HW1/P2/P2.txt
```diff
@@ -1 +1,7 @@
...
+
+I used a random shuffle in P2b. Because hence the pixels will be randomly shuffled into different partitions. Hence the average numbers of calculations will be approximately the same. I think this is a good way to partition because of the randomness in making choices. Visually the P2b histogram also produced a histogram similar to binomial distribution.
```
btweinstein: I agree, I think this is a pretty good way to repartition.

HW1/P2/P2.txt
```diff
@@ -1 +1,7 @@
...
+
+
+The tradoff of this way is that shuffling takes time. The algorithm has a running time of O(n). Hence for large files, this might not be a good way in memory management and processing time.
```
btweinstein: Agreed.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,21 @@
...
+from random import randint
+sc = pyspark.SparkContext(appName="Spark1")
+x_ = range(2000)
+random.shuffle(x_)
+x = sc.parallelize(x_,10)
+rdd = x.cartesian(x)
+#y = sc.parallelize(xrange(2000),10)
+#rdd = x.cartesian(x).repartition(100)
+#rdd = x.cartesion(x).partitionBy(lambda x:randint(1,101))
+ans = rdd.map(lambda x: [x,mandelbrot((x[0]/500.0-2),(x[1]/500.0-2))])
```
btweinstein: You could just use rdd.repartition to randomly assign jobs to different partitions. Your way works, but rdd.repartition is one line!

HW1/P3/P3.txt
```diff
@@ -1 +1 @@
-# Your discussion here
+(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])
```
btweinstein: This is what I got too. Good!

HW1/P3/P3.py
```diff
@@ -0,0 +1,50 @@
...
+	value = [word]
+	return (key,value)
+
+# This is the function maps to reduceByKey, add up all the anagrams in one list
+def reduce_mapper(item_1, item_2):
+	return item_1 + item_2
+
+rawRDD = sc.textFile('EOWL_words.txt')
+
+pairedRDD = rawRDD.map(K_V_mapper) # each item in rawRDD is something like (hashCode, word)
```
btweinstein: Although this is probably fine, it seems complicated. Is making your own hash faster or something?

HW1/P3/P3.py
```diff
@@ -0,0 +1,50 @@
...
+resultRDD = reducedRDD.map(resultmapper).sortBy(lambda x:-x[1]) # firstly put the items into correct form. Then sort 
+# it in descending order
+
+maxLength = resultRDD.take(1)[0][1]
+maxAnagrams = resultRDD.filter(lambda x : x[1] == maxLength)
+
+print maxAnagrams.collect()
+
+answer = resultRDD.take(1)[0]
+f = open('P3.txt', 'w')
```
btweinstein: When I am using open/close statements in python, I always use them in a 

with f as open('blah.txt', 'rb'):
      etc. etc. etc.

HW1/P4/P4.txt
```diff
@@ -1 +1,51 @@
...
+After 1 turn, the number of new visited nodes is 8
+After 2 turn, the number of new visited nodes is 0
+
+
+The diameter of CAPTAIN AMERICA is 3. Which means that if we start running bfs on CAPTAIN AMERICA, it will traverse a connected component of a graph in 3 hops.
+As known that CAPTAIN AMERICAN is on of the most characters in Marvel's comic, so it has a lot of connections. And hence the connected component which CAPTAIN AMERICAN lies in is expected to be a large one. And the the program's result correctly depicts this fact. It has 6408 nodes, comparing with other two connected components, which are 7 and 9, respectively. 
+
+If a character has a distance that is not defined in such a search, it means that this character is not within the same connected_components of the character which the search began with.
+
+-------------END-------------
```
btweinstein: I got the same answers as you. Great!

HW1/P4/P4.py
```diff
@@ -0,0 +1,64 @@
...
+# ouput----a tuple in the required format
+def preprocess_mapper(x):
+    x = x.split('","')   
+    if len(x) == 2:
+        character = x[0][1:]
+        comic = x[1][:-1]
+        return (comic, character)
+    return (-1, -1)
+
+movie_characters_RDD = c.map(preprocess_mapper).filter(lambda (k, v): v != -1).groupByKey().mapValues(lambda x: list(x))
```
btweinstein: The code might be easier to read if you split this over multiple lines. But I think I get it.

HW1/P5/P5.txt
```diff
@@ -1 +1,25 @@
...
+'September_16' is a neighbor of 'Harvard_University', 'Mickey_Rourke' is a neighbor of 'September_16', and 'Kevin_Bacon' is a neighbor of 'Mickey_Rourke'.
+
+
+
+For connected components:
+My script is correctly working on P4's small dataset but not on the large dataset of P5.
+
+My algorithm is:
+1, make two different graphes - asymmetric and symmetric
+2, bfs on one note, find a cc, then change to another node which is not in that cc
```
btweinstein: I don't think the scaling of this algorithm will allow it to work on large datasets. Since you use BFS repeatedly, I think the algorithm will converge in roughly O(N). You can write different algorithms that converge on O(D) or O(log[D]) where D is the diameter of the graph.

HW1/P5/P5.txt
```diff
@@ -1 +1,25 @@
 # Your discussion here
+---------Shortest Path----------------
+
+'Kevin_Bacon'--------->'John_Lithgow'--------->Harvard_University(arrived!)
```
btweinstein: Nice I found similar paths

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,90 @@
...
+
+    # After turn i+1's reduce, we update(/find) all nodes whose distance is (i+1) from origin
+
+    # When we assume diameter is limited, there's nothing special and we just update whole graph turn by turn
+    # When assumption is relaxed, if we update distance of 0 new node in turn i, we know we should stop at here
+    # We use numOfVisitedNodes (type sc.accumulator) to represent all nodes whose distances have been updated in present turn
+    # We use presentRDD.filter(lambda (k, (f, a, d)): d == turnId + 1).count() to get the num of new visited nodes in turn [turnId]
+    # here (k, (f, a, d)) is key-value pair of RDD in each turn, which is the short version of (key, (flag, adjacentNodes, distance))
+    log = ''
+    if isDiameterLimited:
```
btweinstein: Do you avoid shuffles in your code via partitioning? I can't tell

HW1/P5/cc.py
```diff
@@ -0,0 +1,132 @@
...
+max_cid = -1
+
+links_RDD = links.map(links_mapper).partitionBy(4).cache()
+links_intersect, links_union = pre_process(links_RDD)
+
+sym_graph_RDD = links_union.map(dataStruCons).cache()
+revRDD = links_RDD.flatMap(reverse_edges).reduceByKey(lambda x,y: x.union(y)).partitionBy(4)
+asym_graph_RDD = links_RDD.join(revRDD).map(lambda (k,v): (k, (False, (set(v[0]).intersection(v[1])),max_distance,max_cid)))
+
+def bfs_one_node(origin_index, _rdd):
```
btweinstein: Yeah I don't think this will ever work on the amazon dataset...there are too many connected components. In the directed case, there are 10s of thousands of unconnected pieces...so the script would take more than 10,000 iterations to complete, which is basically impossible. I will copy what I did.

I have implemented an algorithm that will compute the number of connected components
in O(D) iterations; I don't think it converges in O(log(D)), but maybe I somehow got O(log(D)) unknowingly. The algorithm works in a couple of steps:

1. Assign each node in the graph a unique integer ID
2. Pass the ID of each parent to their children
3. If the ID passed to the child is smaller than the child's current ID, the child adopts the new ID
4. Goto 2, repeat until the number of unique ID's converges to a constant value.

The maximum amount of time it will take for one node to see the effect of another node is O(D) as the maximum number of nodes between two nodes is D and information propagates along one node per iteration.

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
+import findspark
+findspark.init()
+import pyspark
+import time
+import re
+from random import random
+from bisect import bisect
```
btweinstein: Oh cool, I'm not familiar with the bisect package; what does that do?

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
...
+start = time.time()
+sc = pyspark.SparkContext(appName='test')
+
+def mapFunc(word):
```
btweinstein: It may be helpful to define these functions in the middle of your code when they are needed. Defining them at the beginning makes it difficult to understand when you will use them.

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
...
+	if len(word) == 0:
+		return False
+	return (not word[:-1].isupper()) if word[-1] == '.' else  not word.isupper() and not word.isdigit()
+
+textRDD = sc.textFile('pg100.txt',8)
+rawRDD = textRDD.flatMap(mapFunc)
+wordsRDD = rawRDD.filter(filterFunc) # Now wordsRDD contains all the desired words
+
+firstWords, secondWords, thirdWords = wordsRDD.zipWithIndex().map(lambda (k, v): (v, k)), wordsRDD.zipWithIndex().map(lambda (k, v):(v-1,k)).filter(lambda (k, v): k >= 0), wordsRDD.zipWithIndex().map(lambda (k, v): (v-2, k)).filter(lambda (k, v):k >= 0)
+threeWordSet = firstWords.join(secondWords).join(thirdWords).map(lambda (k, v): v)
```
btweinstein: I don't think you partitioned your RDD's. The join will likely have an extremely high cost consequently. You can use "partitionBy" to avoid that.

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
...
+
+def weighted_choice(choices):
+    values, weights = zip(*choices)
+    total = 0
+    cum_weights = []
+    for w in weights:
+        total += w
+        cum_weights.append(total)
+    x = random() * total
+    i = bisect(cum_weights, x)
```
btweinstein: Alternatively, you can use numpy.random.choice which draws an element from an array based on weights.

HW1/P6/P6.txt
```diff
@@ -1 +1,20 @@
...
+all this- It wounds thine honour else, That thou no care; it shall lose my parson, my priest, my Sir
+
+hangers till then. [Aside] That will be merely a madness; and, tell you, lords, that Clifford fled with the Where
+
+cousin. must say is held That valour is the abstract and record my woes. When heaven doth weep, doth not
+
+a Whitsun morris-dance; For, my good lord; therefore mistrust me not. Hark! the kings and queens of France. Were but
+
+that miracle. Yet, in a coronal of gold without flowers. Exeunt, first passing over the new court? There's no jesting;
+
```
btweinstein: Looks good!

HW1/P2/P2b.py
```diff
@@ -1,3 +1,21 @@
 from P2 import *
+import findspark
+findspark.init()
+import pyspark
+import random
+from random import randint
+sc = pyspark.SparkContext(appName="Spark1")
+x_ = range(2000)
+random.shuffle(x_)
```
xinyanhan: I used the same random shuffle method, nice bell curve!

HW1/P3/P3.txt
xinyanhan: If you take order and print the first few entries, you might find that there are actually 3 entries with 11 anagrams :)

HW1/P4/P4.txt
xinyanhan: Your code is clear, nice comments, got the same answers too!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,90 @@
...
+    # first type simply consists of itself (for we don't want to lose the adjacentNodes information during each turn)
+    # second type will be send only by those nodes that are new visited nodes in the previous turn. We are going to update distances of these nodes' unvisited neighbors
+
+    def mapFunc((key, (flag, AdjacentNodes, distance))):
+        result = []
+        if distance == turnId:
+            for otherKey in AdjacentNodes:
+                result.append((otherKey, (True, [], distance+1)))
+        result.append((key, (False, AdjacentNodes, distance)))
+        return result
```
xinyanhan: Nice way of setting true/false to indicate if a node has been visited or not.

HW1/P6/P6.py
xinyanhan: It might be easier to read if you split up the chain of map() functions into several lines.

PR #201, opened 2015-10-01
HW1

btweinstein:
Leaving for a wedding tomorrow, so I am handing the homework in early. Enjoy!


HW1/P2/P2a.py
```diff
@@ -1,3 +1,61 @@
...
+
+sc = ps.SparkContext(conf=config)
+
+# Do the computation
+
+num_pixels = 2000
+rows = sc.range(num_pixels, numSlices=10)
+cols = sc.range(num_pixels, numSlices=10)
+
+indices = rows.cartesian(cols)
```
sdaulton: Since you use cartesian and rows and columns are identical, you really only need to use sc.range() once to generate rows and then do:

indices = rows.cartesian(rows)

HW1/P2/P2a.py
```diff
@@ -1,3 +1,61 @@
...
+
+def mandelbrot_wrapper(row, col):
```
sdaulton: Nice wrapper

HW1/P2/P2b.py
```diff
@@ -1,3 +1,91 @@
...
+    x = col/(num_pixels/4.) - 2.
+    y = row/(num_pixels/4.) - 2.
+
+    return ((row, col), P2.mandelbrot(x, y))
+
+########### Different from part A: load balancing! ########
+num_partitions=100
+
+# We create a mask of the most expensive computations. These are computations
+# that take about 3 orders of magnitude longer to run than the fastest points!
```
sdaulton: Nice Idea

HW1/P3/P3.py
```diff
@@ -0,0 +1,41 @@
...
+config = ps.SparkConf()
+config = config.setAppName('anagram_solver')
+sc = ps.SparkContext(conf=config)
+
+### Parsing the data efficiently! ###
+wlist = sc.textFile('EOWL_words.txt', use_unicode=True)
+
+def get_anagrams_key(input_str):
+    alphabet = string.ascii_lowercase
+    key = [0 for z in range(len(alphabet))]
```
sdaulton: Interesting way of doing this.  It would use less space, and may be simpler just use sorted letter sequences instead of arrays of character counts that are length 26.

HW1/P3/P3.py
```diff
@@ -0,0 +1,41 @@
...
+    for count, cur_letter in enumerate(alphabet):
+        key[count] = input_str.count(cur_letter)
```
sdaulton: This is linear time for each string and you do if for every string ==> O(N^2).  There are more efficient ways.  See comment above

HW1/P3/P3.py
```diff
@@ -0,0 +1,41 @@
...
+    key = tuple(key)
+    return (key, input_str)
+
+key_string_rdd = wlist.map(get_anagrams_key)
+grouped_by_anagram_rdd = key_string_rdd.groupByKey()
```
sdaulton: as a rule of thumb, you probably want to avoid groupByKey wherever possible

HW1/P3/P3.py
```diff
@@ -0,0 +1,41 @@
...
+
+def get_into_final_form(x):
+    words = list(x[1])
+    num_anagrams = len(words)
+    sorted_key = ''.join(sorted(words[0]))
+    return (sorted_key, num_anagrams, words)
+
+final_form_rdd = grouped_by_anagram_rdd.map(get_into_final_form)
+final_form_result = final_form_rdd.collect()
```
sdaulton: takeOrdered() is a nice way to combine these last 3 lines into 1 line

HW1/P4/P4.py
```diff
@@ -0,0 +1,76 @@
...
+    dat = raw_line.split('"')
+    character_name = dat[1]
+    issue = dat[3]
+    return (issue, character_name)
+
+issue_character_rdd = marvel_raw_data.map(get_issue_character)
+# Put everyone in a group if they are in the same comic and get nodes from that!
+issue_groups_rdd = issue_character_rdd.groupByKey()
+
+def get_links_from_groups(x):
```
sdaulton: This is quite a costly operation, which I think could be parallelized using only rdds

HW1/P4/P4.py
```diff
@@ -0,0 +1,76 @@
...
+# There are a bunch of duplicates; we want to get rid of them
+def cleanup_links(x):
+    linked_to = list(x[1])
+    unique_links = tuple(set(linked_to))
+    return (x[0], unique_links)
+
+network_rdd = character_and_links_rdd.map(cleanup_links).cache() # Completed network representation
+# Our network has the form (Name, [All individuals that the node links to]). Note that
+# in this representation (A, [B]) does not imply (B, [A]), i.e. this is an innately directed
+# representation, but we initialize the network in such a way that the links are bi-directional.
```
sdaulton: Nice comments

HW1/P4/P4.py
```diff
@@ -0,0 +1,76 @@
...
+
+# We now do the breadth-first search, BFS
+from HW1.network_commands import BFS
+
+#### Captain America ####
+searcher = BFS(sc, 'CAPTAIN AMERICA', network_rdd)
+searcher.run_until_converged() # I could run 10 iterations, but the solution converges faster than that
```
sdaulton: Why not make it so it runs 10 iterations or stops if a path is found first?

HW1/P5/P5.txt
```diff
@@ -1 +1,160 @@
-# Your discussion here
+Note that helper functions are stored in the "network_commands" python module in the directory above this one.
+I wanted to place all of my network code in the same place.
+
+###### Shortest path #######
+
+The shortest path between 'Harvard_University' and 'Kevin_Bacon' is of length 3. One example is
```
sdaulton: The shortest path from Harvard University to Kevin Bacon has length 3, but the shortest path from Kevin Bacon to Harvard University has length 2...

HW1/P6/P6.py
```diff
@@ -0,0 +1,106 @@
...
+        words_and_weights = shake_phrases_rdd.lookup(cur_key)[0]
+    return sentence
+
+phrase_list = []
+for i in range(10):
+    phrase_list.append(get_phrase(shake_phrases_sorted))
+
+for cur_phrase in phrase_list:
+    print cur_phrase
+    print
```
sdaulton: Nice comments, clean code

HW1/network_commands.py
```diff
@@ -0,0 +1,253 @@
...
+        def get_smaller_index(a, b):
+            if a < b:
+                return a
+            return b
+
+        children_indices = connected_rdd.flatMap(get_children_index)
+        child_with_smallest_index = children_indices.reduceByKey(get_smaller_index, num_partitions) # This is partitioned
+        joined_rdd= connected_rdd.join(child_with_smallest_index) # copartitioned!
+        connected_rdd = joined_rdd.map(lambda x: (x[0], (x[1][0][0], x[1][1])), preservesPartitioning=True) # Should preserve partitioning...
+        return connected_rdd.cache()
```
sdaulton: Nice object-oriented code.  But is this the best way to code a couple variations of bfs?

HW1/P2/P2a.py
btweinstein: Yes good point, thanks. I think I did it twice for my sanity, i.e. the case they were different.

HW1/P3/P3.py
btweinstein: Agreed. In hindsight, I think I would have gone with the sorted letter sequences! For some reason, it made more sense as I was programming just to count each letter.

HW1/P3/P3.py
btweinstein: Agreed, but is there an easier way in this case? We have to group things eventually.

HW1/P3/P3.py
btweinstein: Yes I agree.

HW1/P4/P4.py
btweinstein: Shouldn't it just scale as O(N)? Doesn't seem that bad to me. There might be a way to do this in parallel though, that's true.

HW1/P4/P4.py
btweinstein: That's what run_until_converged does, I believe.

HW1/P5/P5.txt
btweinstein: I don't understand what you mean, sorry. I guess it depends on how you define a path. If you define the path length as excluding the start and end node, sure. I excluded only the start node.

HW1/P6/P6.py
btweinstein: Thanks! :)

HW1/network_commands.py
btweinstein: I wholeheartedly believe that it is. Whenever I have to program something complex, or even marginally complex, classes simplify our lives as we do not have to deal with 50 interlocking functions. Instead, we can just think about objects and how they are supposed to behave; this is more intuitive to me.

HW1/P5/P5.txt
sdaulton: I believe the question asks for the shortest path from Kevin Bacon to Harvard University and the shortest path from Harvard University to Kevin Bacon.  Since this is a directed graph, those paths are not necessarily the same length.  Here you have found a shortest path from Harvard University to Kevin Bacon (which has length 3).  But you haven't stated the length of the shortest path from Kevin Bacon to Harvard University (which has length 2)

btweinstein: Ah crap you are correct, I forgot to do that. :( Sorry for the confusion. This would be a very simple fix in my code, I would just change the source and destination nodes. Good find!

PR #204, opened 2015-10-04
HW1 submission

LeonhardFS:


HW1/P3/.ipynb_checkpoints/Problem 3-checkpoint.ipynb
```diff
@@ -0,0 +1,232 @@
+{
```
jonathanhfriedman: This is a very minor complaint, but uploading IPython notebooks makes it a bit difficult to look at code by scrolling down the "Files changed" tab, since the notebooks are so long.

HW1/P6/P6.py
```diff
@@ -0,0 +1,142 @@
...
+	if not os.path.isfile('Shakespeare.txt'):
+		# the texts from shakespeare are contained in lines 182 - 124367
+		# create the file Shakespeare.txt using the line range from above
+		# if all is ok, this should return True, True
+		with open(filename, 'r') as f:
+		    lines = f.readlines()
+		    lines = lines[181:124367]
+		    with open('Shakespeare.txt', 'w') as f2:
+		        f2.writelines(lines)
+		    print f2.closed
```
jonathanhfriedman: This seems like a debugging thing that should be removed from final code, since "with open" automatically closes the file. Maybe change it to an assert instead, so it doesn't print output? This is a very minor nitpick, though.

PR #205, opened 2015-10-04
Hw1

quezacot:
Completed HW1.


HW1/P3/P3.txt
```diff
@@ -1 +1,2 @@
 # Your discussion here
+('aerst', 11, ['arets', 'aster', 'rates', 'reast', 'resat', 'stare', 'stear', 'strae', 'tares', 'tears', 'teras'])
```
Sdeppen: I got the same answer!

HW1/P2/P2.txt
```diff
@@ -1 +1,2 @@
 # Your discussion here
+The histograms show a significant improvement of work balance among partitions. The computation can complete faster because the balanced work loading reduced the running time of partition needs the longest time. However, the repartition needs extra efforts to partition the tasks evenly. The trade off is this overhead for the repartition. This overhead needs to be shorter than the improvement of running time to worth doing the repartition.
```
Sdeppen: What kinds of extra efforts?

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+    sc = SparkContext("local", appName="Spark1")
+    sc.setLogLevel('WARN')
+    
+    sidePart = 10
+    numPart = sidePart*sidePart
+    
+    xisrdd = sc.parallelize(xrange(0,2000)) #no partition    
+    cartrdd = xisrdd.cartesian(xisrdd) #generate coordinate
+    
+    balance = cartrdd.partitionBy(numPart)
```
Sdeppen: This seems to work pretty well.

HW1/P6/P6.txt
```diff
@@ -1 +1,12 @@
...
+['And,', 'looking', 'on', 'his', 'golden', 'blood,', 'And', 'let', 'us', 'forth,', 'So', 'that', 'this', 'is', 'the', 'sky,', 'And', 'now', "I'll", 'go']
+['devilish-', "is't", 'not', 'strange?', 'Do', 'you', 'desire?', '[To', 'ANTONIO]', "I'll", 'be', "suppos'd", 'upon', 'a', "slipp'ry", 'place', 'Makes', 'nice', 'of', 'no']
+['yet.', 'Re-enter', 'O', 'madam,', 'my', 'old', 'master', 'must', 'be', "lin'd,", 'So', 'must', 'slender', 'Rosalinde.', 'They', 'that', 'least', 'lend', 'it', 'to']
+['so', 'ever.', 'There', 'be', 'four', 'of', 'us!', 'Though', 'now', 'we', "hous'd", 'him', 'in', 'this', 'city', 'he', 'Hath', 'widowed', 'and', 'unchilded']
+['grant', 'is', 'the', 'baseness', 'of', 'our', 'fears,', 'until', 'our', 'city', 'did.', 'Only', 'myself', 'stood', 'out;', 'For', 'our', 'affairs', 'from', 'England']
+['Messengers,', 'Watchmen,', 'etc.', '<<THIS', '1990-1993', 'LIBRARY,', 'INC.,', '(1)', 'ONLY,', '(2)', 'MEMBERSHIP.>>', 'Flourish.', 'Enter', 'EDWARD,', 'GLOUCESTER,', 'HASTINGS,', 'STANLEY,', 'and', 'others', 'Thus']
```
Sdeppen: It seems some of your words did not get filtered out properly...

HW1/P6/P6.py
```diff
@@ -0,0 +1,48 @@
...
+            return False
+    return True
+
+if __name__ == '__main__':
+    sc = SparkContext("local", appName="Spark1")
+    sc.setLogLevel('WARN')
+    
+    txtfile = sc.textFile('pg100.txt', 32, use_unicode=False) #load text
+    #txtfile = sc.textFile('test.txt', use_unicode=False) #load text
+    words = txtfile.flatMap(lambda line: line.split(" ")).filter(parsefilter)
```
Sdeppen: When you split the lines, I think you'd want to split them without quotes.

HW1/P2/P2.txt
quezacot: repartition requires shuffling.

HW1/P3/P3.txt
newtestsubject: Actually, there are 3 different letter combinations that all have 11 acronyms. I am not sure if we were supposed to identify all 3 all just one.

('aerst', 11, ['arets', 'aster', 'rates', 'reast', 'resat', 'stare', 'stear', 'strae', 'tares', 'tears', 'teras'])
 ('aelst', 11, ['least', 'leats', 'salet', 'slate', 'stale', 'steal', 'stela', 'taels', 'tales', 'teals', 'tesla']) 
 ('aeprs', 11, ['asper', 'pares', 'parse', 'pears', 'prase', 'presa', 'rapes', 'reaps', 'spaer', 'spare', 'spear'])

HW1/P2/P2.txt
newtestsubject: I think that it is important to think about a scenario when shuffling would slow down the computation. This would be the case when for example the list is very large and we at the same time have many processors and therefore divide the data in many small partitions. In this case the small partitions would be processed relatively quickly, but shuffling the list/RDD would take a relatively long time. It such case it might be better not to shuffle.

HW1/P6/P6.txt
newtestsubject: Yes, words that contain only capitalized letters should have been filtered out.

HW1/P2/P2b.py
oyuchoi, line 23: The performance looks much much tighter (1.95-2.2), but I'm curious why this makes it bimodel on the extremes ?

HW1/P5/P5.txt
```diff
@@ -1 +1,46 @@
 # Your discussion here
+
+Part1 shortest paths
+
+I ran two times and got two results:
```
oyuchoi, line 5: Why does running it 2x give 2 different results (rather than the same result or all the results)?

HW1/P6/P6.txt
oyuchoi, line 10: @Sdeppen @newtestsubject have a point

PR #207, opened 2015-10-04
Hw1

Kai-Square:
Submission


HW1/P2/P2b.py
```diff
@@ -1,3 +1,30 @@
...
+    
+    I = sc.parallelize(range(pixel + 1))
+    J = sc.parallelize(range(pixel + 1))
+
+    IJ = I.cartesian(J)
+    IJ = IJ.partitionBy(100) # this is different from part a
+    
+    IJK = IJ.map(lambda ij: (ij, mandelbrot(* IJtoXY(ij[0], ij[1], pixel))))
+    numpart = sum_values_for_partitions(IJK)
+    plt.hist(np.log10(np.array(numpart.collect()) + 1), bins=np.linspace(0, 8, 17))
```
foo-bar-baz-qux: I think it would be good to see a more detailed spread of the distribution here without the log scale.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,30 @@
...
+    conf = SparkConf().setAppName('KaiSquare')
+    sc = SparkContext(conf = conf)
+
+    pixel = 2000
+    
+    I = sc.parallelize(range(pixel + 1))
+    J = sc.parallelize(range(pixel + 1))
+
+    IJ = I.cartesian(J)
+    IJ = IJ.partitionBy(100) # this is different from part a
```
foo-bar-baz-qux: I'm not sure whether this is guaranteed to be random all the time though. It might be that cartesian repartitions in a way  that puts contiguous blocks of the image to be generated together, potentially creating hotspots. Possibly some call to repartition() might be needed for randomness all the time.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,17 @@
+# P4 bfs method
+
+def ssbfs(charname, charadjlist):
+    sc = charadjlist.context
+    result = sc.parallelize([(charname, 0)])  # the rdd that keep track with the visited notes
+    accum = sc.accumulator(0)  # seemingly playful
+    while True:
+        curvalue = accum.value  # the depth
+        waiting = result.filter(lambda kv: kv[1] == curvalue)  # the nodes of deepest level
```
foo-bar-baz-qux: Since we had to use the accumulator, it could also be used here to get the size of the filtered set (same result though.)

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,74 @@
...
+
+def shortestpath(nodestart, nodeend, directedlist):
+    sc = directedlist.context
+    directedlist.cache()
+    visited = set([nodestart])  # visited nodes
+    current = set([nodestart])  # nodes of the current level
+    result = sc.parallelize([(nodestart, None)])  # track the path, child-parent pairs
+
+    while True:
+        if nodeend in current:  # if found
```
foo-bar-baz-qux: Another option is to keep the whole path list in the value (provided sufficient memory.)

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,74 @@
...
+                    break
+            return len(path) - 1, path[::-1]
+
+        if len(current) == 0:  # if no node in current level
+            break  # end the bfs
+
+        waitinglist = directedlist.filter(lambda kv: kv[0] in current)  # reach out from current level
+        waitinglist = waitinglist.flatMapValues(lambda v: v)
+        waitinglist = waitinglist.map(lambda kv: (kv[1], kv[0]))
+        nextlist = waitinglist.filter(lambda kv: not kv[0] in visited)  # new nodes we found
```
foo-bar-baz-qux: Could this also be done with a flag in each element of the RDD to specify whether the node was unseen, about to be processed, or processed? This way it should also save on the expense of calls to collect() and then reinsertion.

HW1/P2/P2b.py
jeewon-hwang: Wow, different approach. I just used random partition, and did not think about repartitioning after cartesian product. But I also wonder as "foo-bar-baz-qux" mentioned, there might be some chunks with similar values causing high iterations.

PR #208, opened 2015-10-04
Hw1

k-ye:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+    return calcBase(ij[0])
+
+if __name__ == '__main__':
+    imageIJ = []
+    for i in xrange(TOTAL_PIXELS):
+        for j in xrange(TOTAL_PIXELS):
+            imageIJ.append((i, j))
+
+    sc = pyspark.SparkContext(appName='YK-P2a')
+    data = sc.parallelize(imageIJ, 100).partitionBy(100)
```
xchang4: Great solution!

HW1/P4/P4.txt
```diff
@@ -1 +1,14 @@
 # Your discussion here
+
+The representation of my graph is an adjacency list for each character. Each entry in the RDD is of the following format:
+  (characterName, [a list of adjacent characterNames])
+
+Character Name    |   Number of nodes touched    | Diameter
+-----------------------------------------------------------
+CAPTAIN AMERICA   |               6408           |    3
```
xchang4: I think you did not strip the white spaces in the character name. If you did that, the answer would be 6403 instead of 6408.

HW1/P2/P2.txt
```diff
@@ -1 +1,3 @@
 # Your discussion here
+
+Fromt the two histogram plots we can see that the load of workers in P2b is much more balanced than that of P2a, as the total iteration time for each partition is narrowed in a small range. However, the transform of repartition will result in shuffle, which causes much more overhead in computation time. So this is the tradeoff.
```
jpbm: The most elegant solution for load balancing I saw was someone who took the computational grid and assigned individual pixels randomly to partitions, so that each partition wound up with a roughly equal number of computationally heavy and light points. This randomization had to be done before converting the grid to RDD.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,34 @@
...
+
+def calcY(ij):
+    return calcBase(ij[0])
+
+if __name__ == '__main__':
+    imageIJ = []
+    for i in xrange(TOTAL_PIXELS):
+        for j in xrange(TOTAL_PIXELS):
+            imageIJ.append((i, j))
+
```
jpbm: Was creating the computational grid in this way a bottleneck in terms of computation time? I feel like it could become an issue because you're doing nested 'for' loops. 

I did it like this, using "map":
q = range(0,4000)
X = sc.parallelize(q,100)
coordinates = X.flatMap(lambda x: [(x,y) for y in q])

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+    return calcBase(ij[1])
+
+def calcY(ij):
+    return calcBase(ij[0])
+
+if __name__ == '__main__':
+    imageIJ = []
+    for i in xrange(TOTAL_PIXELS):
+        for j in xrange(TOTAL_PIXELS):
+            imageIJ.append((i, j))
```
adisuissa, line 20: This should have been done in spark using sc.range and cartesian.

HW1/P3/P3.py
```diff
@@ -0,0 +1,24 @@
...
+  sortedWords = wlist.map(lambda x: (''.join(sorted(x)), set([x])))
+  mergedWords = sortedWords.reduceByKey(lambda s1, s2: s1 | s2)
+  finalRdd = mergedWords.map(lambda x: (x[0], len(x[1]),list(x[1])))
+  finalRdd.cache()
+
+  maxlen = 0
+  anagram = None
+  maxItems = finalRdd.takeOrdered(20, lambda x: -x[1])
+  # print maxItems
+  for item in maxItems:
```
adisuissa, line 16: Unclear why do you need this for loop. Instead you can change the value of takeOrdered to 1 (instead of 20).

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,102 @@
+from pyspark import AccumulatorParam
+
+# Accumulator used for BFS. Basically the algorithm adds all the visited
+# nodes in one iteration into this accumulator instance.
+class SetAccm(AccumulatorParam):
```
adisuissa, line 5: Using the accumulator to store a set of all the nodes and their distances, makes it a very large accumulator, which is not very efficient.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,102 @@
...
+
+    while True:
+        # Initialize the accumulator in each iteration
+        tmpFoundAccm = sc.accumulator(set(), SetAccm())
+        # Those entries in frontier queue will be searched
+        toSearch = graph.filter(lambda (k, _): k in frontier)
+        # Add all the adjacent nodes into the accumulator
+        toSearch.foreach(lambda (k, adj): tmpFoundAccm.add(set(adj)))
+
+        frontier = tmpFoundAccm.value - closed
```
adisuissa, line 27: This is done in python, but should have been done in spark (using joins, subtractByKey, and unions).

PR #210, opened 2015-10-04
Hw1

vivjay30:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,26 @@
...
+array_1 = [i for i in range(2000)]
+array_2 = [i for i in range(2000)]
+
+new_array = []
+for i in array_1:
+	for j in array_2:
+		new_array.append((j,i))
+
+intermediate = sc.parallelize(new_array)
+intermediate = intermediate.partitionBy(100, lambda x : random.randint(0,99)).map(lambda x: ((x[0], x[1]), mandelbrot(x[0]/500.0 - 2, x[1]/500.0 - 2)))
```
fpeng01: The uniform random distribution way works well. Nice strategy.

HW1/P3/P3.py
```diff
@@ -0,0 +1,10 @@
+from pyspark import SparkContext
+sc = SparkContext("local", "Simple App")
+words = sc.textFile("wlist.txt")
+
+rdd = words.map(lambda x : (''.join(sorted(x)), x)) \
+		   .groupByKey() \
+		   .map(lambda (key,values) : (key, len(list(values)), list(values))) \
+		   .sortBy(lambda (x1, x2, x3) : x2, ascending = False)
+
```
fpeng01: Clear style with one function call each line, and also condensed coding. The script is much more shorter than mine.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,87 @@
...
+	distances = sc.parallelize([(startNode, 0)])
+	for i in range(10):
+		# For each node, keep the node in the RDD and add it's neighbors. Take the min distance for ones that have been visited twice
+		distances = distances.flatMap(lambda (node, distance): [(node, distance)] + [(newNode, distance + 1) for newNode in adjacencyList[node]] if distance == i else [(node, distance)]) \
+							 .groupByKey() \
+							 .map(lambda (node, distances) : (node, min(list(distances))))
+		print len(distances.collect())
+	# Could also return distances.collect()
+	return distances.count()
+
```
fpeng01: Nice implement. I used an algorithm that is similar to the traditional bfs. I guess it works because data for p4 is not very large. I should have figured out an algorithm like yours

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,55 @@
...
+sc = SparkContext("local", "Simple App")
+
+[2729536, 3554817, 3554854, 2152782]
+"National_Lampoon's_Animal_House"
+'National_Lampoon_(magazine)'
+
+# Testing the algorithm
+graph = sc.parallelize([(1,(0, None, [2,3])), (2, (float("inf"), None, [5,3])), (3, (float("inf"), None, [5,1,10])), (10, (float("inf"), None, [2])), (5, (float("inf"), None, []))])
+print search_bfs_new(1,10, graph, sc)
+
```
fpeng01: Good idea of starting by testing with a path known.

HW1/P5/P5_helper.py
```diff
@@ -0,0 +1,82 @@
...
+		if distance == frontier_distance:
+			print "Found one"
+			return [(new_node, (distance + 1, node, [])) for new_node in neighbors]
+		else:
+			return [(node, (distance, parent, neighbors))]
+	return helper
+def combine_nodes((node, distances)):
+	ret = min(list(distances), key = lambda (distance, parent, neighbors) : distance)
+	neighbors = max(list(distances), key = lambda (distance, parent, neighbors) : len(neighbors))
+	return (node, (ret[0], ret[1], neighbors[2]))
```
fpeng01: I guess this should be the most common and efficient way of building the graph and then finding path. Is this the "bfs" for paralleling computing?

HW1/P6/P6.py
```diff
@@ -0,0 +1,41 @@
+source = sc.textFile("s3://Harvard-CS205/Shakespeare/Shakespeare.txt")
+source = source.flatMap(lambda x : x.split())
+from operator import add 
+def validWord(word):
+	if word.isdigit() or word.isupper() or (word[-1] == "." and word[:-1].isupper()):
+		return False
+	return True
+
```
fpeng01: Nice work with filtering the text.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,26 @@
...
+from pyspark import SparkContext
 
-# Your code here
+sc = SparkContext("local", "Simple App")
+
+array_1 = [i for i in range(2000)]
+array_2 = [i for i in range(2000)]
+
+new_array = []
+for i in array_1:
```
haosutang: In Problem 2a, using Cartesian may be a good alternative to initialize this large coordinate array. Otherwise, "new_array" will not be parallelized at the beginning but rather stored in memory.

HW1/P3/P3.py
haosutang: Problem 3 is good!

HW1/P4/P4_bfs.py
haosutang: Problem 4: The results looks fine. A lot of effort in getting a nice BFS method. However, the collect() before the search can be costly.

HW1/P5/P5_helper.py
```diff
@@ -0,0 +1,82 @@
...
+
+def search_bfs_new(startNode, goalNode, adjacencyList, sc):
```
haosutang: Problem 5: Your helper file search_bfs_new method is a little tard to follow. I think you had the same idea of finding the path with an rdd storing the node, distance, parent node and neighbors.

HW1/P6/P6.py
```diff
@@ -0,0 +1,41 @@
+source = sc.textFile("s3://Harvard-CS205/Shakespeare/Shakespeare.txt")
```
haosutang: Problem 6: Fairly straightforward. Looks good.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,26 @@
 from P2 import *
+from pyspark import SparkContext
 
-# Your code here
+sc = SparkContext("local", "Simple App")
+
+array_1 = [i for i in range(2000)]
```
adisuissa, line 6: I think you can use array_1 = range(2000), and it will have the same output.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,26 @@
...
-# Your code here
+sc = SparkContext("local", "Simple App")
+
+array_1 = [i for i in range(2000)]
+array_2 = [i for i in range(2000)]
+
+new_array = []
+for i in array_1:
+	for j in array_2:
+		new_array.append((j,i))
```
adisuissa, line 13: You create the ranges and the cartesian product in python, but you should have done it in spark instead.

HW1/P2/P2b.py
adisuissa, line 16: The randint range should probably be (0, 100), as it uses the [0, 100) range.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,87 @@
+from pyspark.accumulators import AccumulatorParam
+
+class BFSAccumulatorParam(AccumulatorParam):
```
adisuissa, line 3: This makes the accumulators to be very large, and inefficient.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,87 @@
...
+	# While we still have unvisited nodes
+	while not distances.isEmpty():
+		# A helper function to add ot the accumulator
+		def f(listvalue):
+			accum.add({listvalue[0]:listvalue[1]})
+
+		distances.foreach(f)
+		# Add all the elements in the rdd to to the seen set
+		current = distances.collect()
+		seen = sc.broadcast(seen.value.union(set([node[0] for node in current])))
```
adisuissa, line 38: You collect and then broadcast the BFS closed list (nodes already seen) to all nodes.
This is very costly, and could be avoided by using joins, subtractByKey, and unions.

HW1/P4/P4.py
```diff
@@ -0,0 +1,39 @@
...
+		 .flatMap(lambda (book, characters): itertools.permutations(list(characters), 2)) \
+		 .groupByKey() \
+		 .map(lambda (key, values) : (key, set(values))) \
+		 .collect()
+
+
+# Turn rdd into dict
+myAdjacencyList = {}
+for item in data:
+	myAdjacencyList[item[0]] = item[1]
```
adisuissa, line 34: This makes the adjacency list to be a python construct instead of a spark rdd.
The next step is basically emitting this to all workers, which has a non-negligible overhead.

HW1/P5/P5_helper.py
```diff
@@ -0,0 +1,82 @@
+from pyspark.accumulators import AccumulatorParam
+class BFSAccumulatorParam(AccumulatorParam):
```
adisuissa, line 2: As in P4, this uses a large dataset for the accumulator, resulting in a non-negligible overhead.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,55 @@
...
+	while True:
+		if page_names[i-1] == input:
+			print i
+			break
+		i += 1
+
+def get_name(page_names, index):
+	return page_names[index-1]
+
+Harvard = find_index(my_list, "Harvard_University") #2152782
```
adisuissa, line 43: Where is my_list defined?

HW1/P5/P5_helper.py
```diff
@@ -0,0 +1,82 @@
...
+		new_distances.foreach(f)
+		# If we found the end
+		if goalNode in accum.value:
+			found = 1
+			break
+		# If we didn't explore any new nodes
+		if len(accum.value) == prev_len:
+			break
+		# For each node, keep the node in the RDD and add it's neighbors. Take the min distance for ones that have been visited twice
+		adjacencyList = adjacencyList.flatMap(expand_node(frontier_distance))
```
adisuissa, line 64: You could have used joins and unions to do the same thing.
Also separating the neighbors from the distance and parent might be useful (to reduce the amount of data that is being moved around).

HW1/P6/P6.py
```diff
@@ -0,0 +1,41 @@
...
+			  .grouByKey() \
+			  .map(lambda (key, values) : (key, len(values))) \
+			  .map(lambda ((w1, w2, w3), count) : ((w1, w2), (w3, count)))\
+			  .grouByKey()\
+			  .map(lambda (k, v) : (k, list(v)))
+
+# Print out 10 sentences of 20 words
+ret = []
+for i in range(10):
+	start = fifth.takeSample(True, 1)
```
adisuissa, line 28: Where is fifth initialized?
Does it supposed to be final?

HW1/P6/P6.py
```diff
@@ -0,0 +1,41 @@
...
+	current_key = start[0][0]
+	sentence = []
+	for i in range(20):
+		sentence.append(current_key[0])
+		values = fifth.lookup(current_key)[0]
+		best = max(values, key = lambda (word, count) : count)
```
adisuissa, line 34: It seems like you are choosing the word with the highest count.
You should have chosen the word randomly according to the number of appearances (see numpy's random.choice).

PR #213, opened 2015-10-04
HW1

averyfaller:


HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,35 @@
...
+        i = itr_acc.value
+        new_nodes = nodes.filter(lambda node: filterer(node[1] + 1)).join(graph).flatMap(lambda node: node[1][1]).distinct().map(lambda x: (x, i))
+        # Cogroup the input nodes with the new nodes, taking only the first iteration of each character as its value
+        nodes = nodes.cogroup(new_nodes).map(lambda x: getCogrouper(x)).partitionBy(8).cache()
+        node_count = nodes.count()
+        
+        # print "NODES PARTITIONS ", nodes.partitioner.numPartitions
+        # print "GRAPH PARTITIONS ", graph.partitioner.numPartitions
+        # Short-circuit if we didn't add any new nodes this iteration
+        if node_count == total_acc.value:
```
jonathanhfriedman: An arguably more Spark-y approach to using accumulators is to add to your accumulator if/when a new path is encountered, and end the search if the accumulator's value is the same before and after an iteration.

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,93 @@
+from pyspark import SparkContext, SparkConf
```
jonathanhfriedman: Much respect for doing this problem!

averyfaller: Thanks!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,35 @@
+def filterer(i):
+    def filt(x):
+        return x==itr_acc.value
+    return filt
```
jacobbaron: I'm a little confused by this. Why do you need this nested function structure? You're basically just doing a filter to only add nodes from the present iteration right? Can't you just do that with a lambda function?

HW1/P4/P4_bfs.py
jacobbaron: I agree with Jonathan -- as it is, it doesn't seem like this accumulator does anything parallel. I think for this problem the intention for the accumulator was to use it to replace count() as a check to see if you have finished searching. In this case, count() is really the same operation, but the accumulator has somewhat more flexibility in general, allowing you to access certain things within the data without collect().

HW1/P5/P5_connected_components.py
jacobbaron: Were you able to get any results with your code? I believe I did something pretty similar to your code, and it mapped out some of the space pretty well, but it got hung up by the fact that there were a bunch of pages (~few thousands) that only link to 1 or two other pages, so it would have taken like 8 hours to finish.

HW1/P2/P2.txt
```diff
@@ -1 +1,6 @@
...
+
+P2b_hist.png has a much smaller spread of times (1.94-2.12 million as opposed to 0 to 9 million ).  Also, the shape of the graph is much prettier, with a nice bell-curve shape.  Most of the tasks fall in the middle of the range, with just a few falling outside.  I tried partitioning based on columns, rows and a combination of the two, but randomly assigning the tasks appeared to work best at spreading out the pixels that take the longest to compute amongst all of the partitions.
+
+Although the random partitioning strategy is faster overall than the default partitioning, it does add shuffling time to the runtime.  And while in this instance it is still faster to use the random partitioning strategy, there will be instances where the shuffling time counter-balances the gains made by re-partitioning.  It is important to pay attention to this in case the cost of shuffling becomes more than the time it saves.
```
oyuchoi, line 6: Your graphs' x-axis are hard to read!

HW1/P4/P4_bfs.py
oyuchoi, line 30: Yes, you should be using an accumulator for part 4!

PR #214, opened 2015-10-04
Hw1

chengchengqin:


HW1/P4/P4.txt
```diff
@@ -1 +1,18 @@
-# Your discussion here
+1. Description:
+Firstly, I set up a RDD with the key of each comic issue, the value of the characters that appear in that comic issue. Through several transformations, I set up a final graph, which is a paired RDD, with the key of each character, and the value of all the characters who are in the same comic issue.
+
+2. Graph diameter:
+We start at distance 0 which can be seen as the center of the graph. Every time we move to the neighbours with distance of 1. Since the diameter is 10, meaning that we can move 5 times. Therefore, the maximum number of steps we can search is 5.
+
```
kevinkeyjkw: Hi chengcheng, since the diameter of the graph is the greatest distance between any two vertices in a graph. https://en.wikipedia.org/wiki/Distance_(graph_theory), I think the worst case scenario would be 10, not 5.

HW1/P6/P6.py
```diff
@@ -0,0 +1,30 @@
...
+words=lines.flatMap(lambda x: x.split())
+
+words_filter=words.filter(lambda x: re.match(r"^(\d+|[A-Z]+\.*)$",x) is None)
+
+RDD=words_filter.zipWithIndex()
+RDD1=RDD.map(lambda x: (x[1],x[0]))
+RDD2=RDD1.map(lambda x: (x[0]-1,x[1]))
+RDD3=RDD1.map(lambda x: (x[0]-2,x[1]))
+RDD1_RDD2=RDD1.join(RDD2)
+together=RDD1_RDD2.join(RDD3)
```
kevinkeyjkw: Subtracting 1 and 2 from each index and then joining is a great idea!

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,25 @@
+from P5_bfs import *
```
zezhouliu: For problems like this one, it would be helpful to have comments to give insight on your strategy

PR #216, opened 2015-10-05
Hw1

haosutang:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,28 @@
...
+import matplotlib.pyplot as plt
+
+sc = pyspark.SparkContext(appName='myAppName')
+
+i = np.arange(1,2001)
+j = np.arange(1,2001)
+
+random.shuffle(i)
+random.shuffle(j)
+
```
nicodri: Why don't you do the shuffle inside Spark? For example, you can call a partitionBy with a shuffle on the index on the rdd intensity. Then you won't have any issue if the data set is too large. And also that way the work is even more balanced.

HW1/P3/P3.txt
```diff
@@ -1 +1,42 @@
...
+   u'pares',
+   u'parse',
+   u'pears',
+   u'prase',
+   u'presa',
+   u'rapes',
+   u'reaps',
+   u'spaer',
+   u'spare',
+   u'spear']),
```
nicodri: Nice job. I only print the first top anagram, I didn't realize other had also 11 elements!

HW1/P4/P4.py
```diff
@@ -0,0 +1,83 @@
...
+
+	return graph
+
+# Get Graph method 2:
+def getGraph_cartesian(data):
+	'''Use Cartesian method to return a transformed graph RDD of the form (chara, [neighbors])
+	Complexity is O(#character^2)
+	Better to use when characters are few while dataset is large.
+	'''
+	data = data.map(lambda l: l.split('"')).map(lambda l: [w for w in l if w!='' and w!=',']).map(lambda l: (l[0], l[1]))
```
nicodri: I got the same idea but you can make it with only one map.

HW1/P4/P4.py
```diff
@@ -0,0 +1,83 @@
...
+
+	# create graph: list of (chara, [neighbors])
+	data_matrix = data2.cartesian(data2)	#creates n by n array for finding neighbors
+	data3 = data_matrix.map(lambda l: (l[0][0], (l[0][1], l[1][0] if intersection(l[0][1],l[1][1]) and l[0][0]!=l[1][0] else '')))	#find neighbors
+	data3 = data3.filter(lambda l: l[1][1]!='')	# clean up matrix
+	data3 = data3.map(lambda l: (l[0], l[1][1]))
+	graph = data3.groupByKey().map(lambda l: (l[0], list(l[1])))
+
+	return graph
+
```
nicodri: Your method provides the right result. But as you wrote it may be long to run for large data set. Maybe you should try to build your graph of (car, [car_neighbors]) in three steps:
* compute the grah (issue, char)
* compute the rdd of the edges of the graph: (char A, char B)
* Group By key the edges graph and aggregate the list of neighbors as value

To gain in efficiency, don't forget to cache your final graph as you will use it often in your BFS.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,23 @@
...
+        
+        # get the neighbors of those in the tocheck list and put them into a neighbors list
+        neighbors = graph.filter(lambda l: l[0] in tocheck).map(lambda l:l[1]).reduce(lambda a,b: a+[i for i in b if i not in a])
+
+        # clean up and put it into checked list and rerun this for unsearched characters.
+        unique_neighbors = [i for i in neighbors if (i not in checked and i not in tocheck and i!=character)]
+        checked = checked + tocheck
+        tocheck = unique_neighbors
+        # diameter +=1
+        accum.add(1)
```
nicodri: Your accumulator could be a simple int here, you are not really using it inside Spark right?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,23 @@
...
+    # diameter_max = 10
+    # diameter = 0
+    
+    tocheck = [character]
+    checked = []
+
+    while len(tocheck)!=0: # and diameter<diameter_max:
+        
+        # get the neighbors of those in the tocheck list and put them into a neighbors list
+        neighbors = graph.filter(lambda l: l[0] in tocheck).map(lambda l:l[1]).reduce(lambda a,b: a+[i for i in b if i not in a])
```
nicodri: The function provides the right results but I am worried about its execution time as you are using an action (reduce) for each stage to extract the neighbors. Maybe a version without any action which returns a rdd may be more efficient/

HW1/P5/P5_bfs_v3.py
```diff
@@ -0,0 +1,132 @@
...
+    
+
+shortest_path_id = []
+
+for i in path:
+	shortest_path_id += [page_names.filter(lambda (K, V): K==i).collect()]
+
+print '********************PATH**********************'
+print list(reversed(shortest_path_id))
+print '**********************************************'
```
nicodri: We have the same results, except your code seems slower. Also I got the general idea of your approach but your code is long and hard to follow.

HW1/P6/P6.py
```diff
@@ -0,0 +1,67 @@
...
+
+for key in start_words:
+    sentence[key] = key[0] +' '+ key[1]
+    newkey = key
+    for size in range(18):# generate a 20-word size sentence
+        try:
+            candidates = word_connections.filter(lambda l: l[0]==newkey).collect()# here collect only one element
+            candidates = dict(candidates[0][1])
+            inverse = [(v, k) for k, v in candidates.items()]
+            best_candidate =  max(inverse)[1]#get the highest frequency, if the numbers are the same, just use the last one.
```
nicodri: You should have used here a probabilistic model, i.e use the frequency of each neighbor as its probability to appear and sample one each time. You can biased the model to make the highest frequency appear more often but still we need to have a randomization.
Otherwise your sentence will be the same for each call as the next word is found in a deterministic manner.

HW1/P2/P2b.py
stephenshifang: I agree with nicodri. I used partitionBy as well.

HW1/P6/P6.py
stephenshifang: I agree. I'd calculate the relative frequencies of the next word and use np.random.choice to generate the next word given the empirical distribution. Depending on the situation, this deterministic approach might work as well however.

PR #217, opened 2015-10-05
Hw1

zelong430:


HW1/P4/P4.py
```diff
@@ -0,0 +1,75 @@
...
+graph_dist = graph.map(lambda x: (x[0], ( 999999 , x[1]   )))
+
+
+#generate txt file
+f = open('P4.txt','w')
+
+#write result of "CAPTAIN AMERICA"
+output = bfs(graph_dist, 'CAPTAIN AMERICA', 6, sc )
+result = output[0].cache()
+num_iter = output[1] - 1
```
apetschek: I like how you not only kept track of total nodes connected to Captain America, but that you also kept track of the number of nodes by specific distance (0, 1, 2).

HW1/P2/P2b.py
```diff
@@ -1,3 +1,30 @@
...
+sc = pyspark.SparkContext()
+i = range(2000)
+j = range(2000)
+
+random.shuffle(i)
+random.shuffle(j)
+
+rdd_i = sc.parallelize(i, 10)
+rdd_j = sc.parallelize(j, 10)
+
```
apetschek: For randomization, i think it is also possible to randomize the partitions you create on the original data instead of randomizing the points prior to parallelizing the data. The method .repartition() does this.

HW1/P3/P3.txt
```diff
@@ -1 +1,3 @@
-# Your discussion here
+(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])
+(u'aelst', 11, [u'least', u'leats', u'salet', u'slate', u'stale', u'steal', u'stela', u'taels', u'tales', u'teals', u'tesla'])
+(u'aeprs', 11, [u'asper', u'pares', u'parse', u'pears', u'prase', u'presa', u'rapes', u'reaps', u'spaer', u'spare', u'spear'])
```
apetschek: I got the same ones!

HW1/P6/P6.py
```diff
@@ -0,0 +1,86 @@
...
+#########################################################################################
+#Shift word RDD
+#Generate RDD with items containing 3 consecutive words
+zip_index = words_filtered.zipWithIndex()
+no_shift = zip_index.map(lambda x: (x[1], ( 0, x[0])))
+left_shift1 = no_shift.map(lambda x: (x[0]-1, (1,x[1][1])))
+left_shift2 = left_shift1.map(lambda x: (x[0] - 1, (2, x[1][1])  ))
+collection_rdd = sc.union([no_shift, left_shift1, left_shift2])
+phrase_rdd = collection_rdd.groupByKey().map(lambda (x,y): (x,[i for i in y])).filter(lambda x: len(x[1]) == 3)
+
```
apetschek: this was a very clever method for getting the triplets of words! i had thought about doing three separate zipWithIndex() calls, but I like how you did it here with a single zipWithIndex() and then creating the left shift, no shift, right shift RDDs and combining them together.

HW1/P2/P2b.py
dominedo: I hadn't thought about randomizing the data before creating the RDD.  It probably doesn't make a difference here seeing as the data isn't particularly large, but do you have an idea of whether it is faster to randomize pre- or post-RDD?

HW1/P3/P3.py
```diff
@@ -0,0 +1,29 @@
...
+group_rdd = anagram_rdd.groupByKey()
+list_rdd = group_rdd.map(lambda x: (x[0], [i for i in x[1]]) )
+result_rdd = list_rdd.map(lambda x : (x[0], len(x[1]), x[1]))
+
+
+#find the largest number of anagrams
+#and write to tx file
+def compare(x):
+	return x[1]
+max = result_rdd.top(3, key = compare)
```
dominedo: Nice! I used .takeOrdered(), but this looks like a more intuitive approach to sorting lists in descending order.

HW1/P4/P4.txt
```diff
@@ -1 +1,32 @@
-# Your discussion here
+Result of CAPTAIN AMERICA
+Num of nodes has distance 0 from CAPTAIN AMERICA: 1906
+Num of nodes has distance 1 from CAPTAIN AMERICA: 4463
+Num of nodes has distance 2 from CAPTAIN AMERICA: 38
+CAPTAIN AMERICA touched 6407 nodes in total, using 3 iterations
```
dominedo: I like how you kept track of the number of nodes touched at each iteration.

HW1/P4/P4.txt
```diff
@@ -1 +1,32 @@
...
+
+Result of MISS THING/MARY
+Num of nodes has distance 0 from MISS THING/MARY: 6
+MISS THING/MARY touched 6 nodes in total, using 1 iterations
+
+
+Result of ORWELL
+Num of nodes has distance 0 from ORWELL: 8
+ORWELL touched 8 nodes in total, using 1 iterations
+
```
dominedo: Are you including the original node as one of the 'touched' nodes? I have one more node than you, but it looks like this is the only reason for the difference.

HW1/P6/P6.py
```diff
@@ -0,0 +1,86 @@
...
+words_filtered = words_no_cap.filter(lambda x: not capt_letter_period(x)).filter(lambda x: not x=='')
+#########################################################################################
+#Shift word RDD
+#Generate RDD with items containing 3 consecutive words
+zip_index = words_filtered.zipWithIndex()
+no_shift = zip_index.map(lambda x: (x[1], ( 0, x[0])))
+left_shift1 = no_shift.map(lambda x: (x[0]-1, (1,x[1][1])))
+left_shift2 = left_shift1.map(lambda x: (x[0] - 1, (2, x[1][1])  ))
+collection_rdd = sc.union([no_shift, left_shift1, left_shift2])
```
dominedo: Nice! I used the same approach.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,128 @@
...
+	num_of_iter = 0
+	num_node_not_touch_start = 0
+	g = g.map(lambda x: (init_start_node(x, v ) ) )
+	flag = True
+	for i in range(max_iter):
+		#num_node_already_touch = num_of_node
+		num_of_iter = num_of_iter + 1
+		g = g.flatMap(lambda x: update(x, num_of_iter - 1, end_v))
+		g = g.reduceByKey(lambda x,y : reconstruct(x,y))
+		#num_new_touch = g.filter(lambda x: x[1][0] == num_of_iter).count()
```
dominedo: It looks like you didn't use .cache() inside your loops - did this ever cause you trouble? I'm curious, because I found that I needed to cache RDDs in order for them to retain their values across iterations.

HW1/P3/P3.txt
zezhouliu: It is nice that you provided more than just one of the anagrams!

HW1/P4/P4.txt
zezhouliu: Yes, I think this is the case (+1 to each of these answers)

PR #218, opened 2015-10-05
Hw1 - Thibaut Perol

tperol:
Completed HW1 - Thibaut Perol


HW1/P2/P2a.py
```diff
@@ -1,3 +1,71 @@
...
+# shut down the previous spark context
+#sc.stop() 
+sc = pyspark.SparkContext(appName="myAppName")
+import numpy as np
+import matplotlib.pyplot as plt 
+import matplotlib.cm as cm
+import pdb
 
-# Your code here
+def mandelbrot(x, y):
```
vivjay30: You don't need to copy all these helper functions into P2a.py, instead you can just import them using from P2 import *

HW1/P3/P3.py
```diff
@@ -0,0 +1,47 @@
...
+wlist = wlist.map(lambda r: r.lower())
+
+# sort all the words alphabetically, check if they are in the dictionary
+anagram = wlist.map(lambda r: ( ''.join( sorted(r) ), r)).groupByKey().sortByKey()
+
+# create the requested structure
+myrdd = anagram.map(lambda r: (r[0], len(r[1]), r[1] ))
+
+# Get the largest valid number of anagrams
+myanagrams = myrdd.takeOrdered(1 ,key=lambda x: -x[1] )
```
vivjay30: There are actually multiple anagrams with that maximum number of permutations so you should take more than 1 to make sure you are aware of all maximum anagrams

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,59 @@
...
+         # Find the nodes at a distance d and creates (nodes,1)
+         # where nodes are the nodes at a distance d
+         node_d = nodes.join(node_d).map(lambda  x: (x[0], x[1][0] )).flatMapValues(lambda x: x).map(lambda g: (g[1],1)).groupByKey()
+         node_d = node_d.map(lambda x: (x[0],1))
+         
+         # update allchar rdd by subtracting the characters just found
+         allchar = allchar.subtractByKey(node_d)
+
+         # This is the number of characters not touched by the search
+         num_char_new = len(allchar.collect())
```
vivjay30: You really don't want to be calling collect. It's very slow. There are ways to run the entire BFS using an accumulator, or better RDD structures without using collect() on every iteration

HW1/P6/P6.py
```diff
@@ -0,0 +1,112 @@
...
+    elif phrase_number==7:
+        word1 = 'You' ; word2 = 'are'
+    elif phrase_number==8:
+        word1 = 'They' ; word2 = 'are'
+    elif phrase_number==9:
+        word1 = 'There' ; word2 = 'is'
+    
+    number_words = 2
+    phrase = [word1, word2]
+    while number_words <= 19:
```
vivjay30: This section of the code could be cleaner if you had a single rdd with pairs of words as keys so you only had to do a single lookup on the pair of words, not two lookups with one lookup for each word

HW1/P2/P2a.py
```diff
@@ -1,3 +1,71 @@
...
+    return (i,j),count
+
+pi = sc.parallelize(range(2000),10)
+pj = sc.parallelize(range(2000),10)
+
+# Create the tuples K =  (i,j)
+K = pi.cartesian(pj)
+
+# Now apply the function mantelbrot to the pixel
+rdd = K.map(lambda p : Pixel_values(p[0], p[1]))
```
draj36: Nice use of map function.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,97 @@
...
+
+# Here we can sort by distance to the point that requires the most iterations
+#K = pi.cartesian(pj).sortBy(lambda x: math.sqrt( (x[0]-1000.0)**2 + (x[1]-1000.0)**2  ), ascending = True)
+#K = K.map(lambda x :(math.sqrt( (x[0]-1000.0)**2 + (x[1]-1000.0)**2  ),(x[0],x[1]) ))
+
+# Here we redistribute randomly the K tuplets 
+K = pi.cartesian(pj).sortBy(lambda x: random.randint(0,1) , ascending = True)
+
+# Now check the distance to the point that required the most iteration
+K = K.partitionBy(100).map(lambda x :(math.sqrt( (x[0]-1000.0)**2 + (x[1]-1000.0)**2  ),(x[0],x[1]) ))
```
draj36: Partitioned according to pixel distance from max compute point (1000,1000). Interesting

HW1/P4/P4.py
```diff
@@ -0,0 +1,42 @@
...
+
+wlist = sc.textFile('P4_data.txt',  use_unicode=True)
+
+# Convert into (keys, values) with key= comic book  values = characters in the comic book
+wlist = wlist.map(lambda x : x.split('"') )
+comlist = wlist.map(lambda x : (x[3],x[1]) ).groupByKey()
+
+# Create a rdd with keys = characters and values = neighbords of the key/character
+newcom = comlist.values().map(lambda x: [char for char in x] )
+newcom = newcom.map(lambda x: list(it.permutations(x,2)) ).map(lambda x: (1,list([i for i in x]))  ).flatMapValues(lambda x : x).values().groupByKey().sortByKey()
```
draj36: Maybe split up operations to make it easier to read.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,59 @@
...
+
+	# now we gonna iterate until d = 10
+	while d<10:
+         # Find the nodes at a distance d and creates (nodes,1)
+         # where nodes are the nodes at a distance d
+         node_d = nodes.join(node_d).map(lambda  x: (x[0], x[1][0] )).flatMapValues(lambda x: x).map(lambda g: (g[1],1)).groupByKey()
+         node_d = node_d.map(lambda x: (x[0],1))
+         
+         # update allchar rdd by subtracting the characters just found
+         allchar = allchar.subtractByKey(node_d)
```
draj36: Good use of subtractByKey to update untouched nodes.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,139 @@
...
+	assert len(Harvard_University) == 1
+	Harvard_University = Harvard_University[0][0]  # extract node id
+
+
+	# run the algorithm to find the distance
+	distance = find_distance(neighbors,
+        	page_names,
+        	iterations=10) # this is the max diameter
+
+	print distance
```
draj36: Distance is found by computing number of BFS iterations necessary to find node. Path is not returned though.

HW1/P6/P6.py
```diff
@@ -0,0 +1,112 @@
...
+for phrase_number in range(0,10):
+    
+    print 'phrase number: ', phrase_number
+    print  'final: ', final
+    
+    if phrase_number==0:
+        word1 = 'He' ; word2 = 'can'
+    elif phrase_number==1:
+        word1 = 'While' ; word2 = 'he'
+    elif phrase_number==2:
```
draj36: Hard coded first 2 words. Could also do random initializations.

HW1/P6/P6.py
```diff
@@ -0,0 +1,112 @@
...
+        # Collect the rdd with indices for word 1 when followed by word 2 (this takes forever)
+        indices_list = indices.keys().collect()
+    
+
+        # Filter the rdd with (index, word) from Shakespeare
+        sentences = index_words.filter(lambda (ind3, w3): ind3-2 in indices_list).map(lambda (k,v):(v,k))
+        sentences = sentences.groupByKey().map(lambda (k,v): (k, len(v) )).sortByKey()
+
+        # Now take the most popular 3rd word
+        word3 = sentences.takeOrdered(1, lambda x: -x[1])[0][0]
```
draj36: Third word should be randomized according to its relative frequency (to other third words) conditional on the first 2 words. If I'm not mistaken, takes most popular instead.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,97 @@
...
+#plt.xscale('log')
+plt.xlim([1e6,3*1e6])
+plt.xlabel('Number of iterations', fontsize = 20 )
+plt.ylabel('Number of partitions',fontsize = 20)
+plt.grid(True)
+plt.gcf().set_size_inches(8, 8)
+plt.savefig('P2a hist.png', box = 'tight')
+plt.tick_params(axis='x', labelsize=20)
+plt.tick_params(axis='y', labelsize=20)
+plt.savefig('P2b_hist.png')
```
zezhouliu: Looks good, but maybe increase the granularity of your histograms?  Having each bar be 500k apart (~16.66% in your histogram) is a large spread per bar

HW1/P6/P6.py
zezhouliu: Agree with all the above points by @draj36  and @vivjay30

PR #219, opened 2015-10-05
HW1

foo-bar-baz-qux:


HW1/P2/P2.txt
```diff
@@ -1 +1,23 @@
...
+
+# Trade-offs
+
+While there were better results in the optimised version, there were some tradeoffs:
+- The input array was generated on the driver and then parallelized, meaning the driver stores 2000^2 elements.
+Not a problem at this dimension but potentially problematic for larger images. Using cartesian functions within
+the RDD means there is shuffling, but less strain on the driver.
+
+- Shuffling the input array takes some time, especially for very large inputs. At 2000^2 elements, this is
+unlikely to cause much issue, but at even larger sizes it might be problematic.
```
dominedo: Did you consider doing this the other way round, i.e. creating the RDD and then shuffling it, instead of shuffling the input array before creating the RDD? I took this approach and found that the shuffle was relatively fast (though I don't know exactly how it would compare to yours).

HW1/P2/P2a.py
```diff
@@ -1,3 +1,24 @@
...
+effort = sum_values_for_partitions( rdd ).collect()
+print 'Time =', time.time() - start_t
+print effort
+
+print effort
+plt.figure()
+plt.xlabel("Iterations")
+plt.ylabel("Number of partitions")
+plt.hist(effort)
+plt.show()
```
dominedo: It's interesting that your graph for this part is bimodal - most of the ones that I have seen only have one peak. Do you think this is related to how you generated your input arrays?

HW1/P3/P3.py
```diff
@@ -0,0 +1,9 @@
+# Q3
+wlist = sc.textFile('EOWL_words.txt')
+
+maxAnagrams = wlist.map(lambda x: (''.join(sorted(x)), [x]))\
+                    .reduceByKey(lambda x, y: x+y)\
+                    .map( lambda (s, v): (s, len(v), v) )\
+                    .reduce( lambda x, y: x if x[1] > y[1] else y )
+
+print maxAnagrams
```
dominedo: Wow - just one line of code! Very impressive!

HW1/P4/P4.txt
```diff
@@ -1 +1,69 @@
-# Your discussion here
+# Assumptions
+I've tried to clean the data set somewhat by removing whitespace and making sure there are no duplicate
+or self-edges. This leads to a slightly lower number of total nodes (and hence fewer touched nodes.)
+
```
dominedo: I did this too, and my results match yours.

HW1/P4/P4.py
```diff
@@ -0,0 +1,51 @@
...
+        c1 = chars[i]
+        if c1 not in adjList:
+            adjList[c1] = set()
+        for j in xrange(i+1, len(chars)):
+            c2 = chars[j]
+            if c2 not in adjList:
+                adjList[c2] = set()
+            adjList[c1].add(c2)
+            adjList[c2].add(c1)
+
```
dominedo: Did you consider also processing the CSV file and creating the adjacency list with RDDs, instead of just using them in your BFS function?

HW1/P6/P6.py
```diff
@@ -0,0 +1,77 @@
...
+    print k, v
+    return ((k[0], k[1]), [(k[2],v)])
+
+print 'Cleaning input...'
+
+allWords = []
+with open('pg100.txt','r') as f:
+    for line in f:
+        allWords += clean(line)
+
```
dominedo: Same comment as above: did you consider using RDDs for this part of the code as well?

HW1/P6/P6.py
```diff
@@ -0,0 +1,77 @@
...
+for i in xrange(10):
+    phrase = []
+    first = RDD.takeSample(True, 1, None)[0]
+    phrase += list(first[0])
+    print 'Starting with', phrase
+    for j in xrange(18):
+        look = RDD.filter(lambda ( (k), v ): (k == tuple(phrase[-2:]))).take(1)
+
+        if len(look) != 1:
+            print "ERROR: no results returned, selecting random word"
```
dominedo: Nice! I appreciate the error checking.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,192 @@
...
+ret1 = resolvePath(page_names, kb2hu)
+print 'Kevin Bacon => Harvard University'
+print ret1
+
+hu2kb = getPath(links, hu_idx, kb_idx)
+ret2 = resolvePath( page_names, hu2kb )
+print 'Harvard University => Kevin Bacon'
+print ret2
+
+
```
dominedo: Thanks for taking the time to include comments - it makes your code very readable. :+1:

HW1/P4/P4.py
foo-bar-baz-qux: I didn't think about this initially, but didn't have much time left to implement it when this point was mentioned on Piazza. I think it's probably more robust to handle this with RDDs if you have very large graphs, even though it's likely slower.

HW1/P2/P2a.py
foo-bar-baz-qux: Hmmm interesting, yeah it's probably to do with the partitioning call I make after doing the cartesian. I've seen people paralellize with 10 partitions and then after the cartesian function will have 100 partitions. I've just gone with the default partitioning and then made 100 partitions after that so that might explain the difference.

HW1/P2/P2.txt
foo-bar-baz-qux: This is a good point. I only realised there was a repartition command after I code-reviewed someone else's submission; I think it's more robust to shuffle the RDD since then you don't have to worry about whether the driver will run out of memory.

HW1/P3/P3.py
LeonhardFS, line 9: Indeed :+1:

HW1/P4/P4.py
LeonhardFS, line 45: Did you consider encapsulating those routines into one class or functions? It is easier to read something like

adjList = readAdjListFromCSV(...) 

as going through a lot of lines of code because the function itself describes precisely what you are doing. And you can reuse the code later!

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,158 @@
...
+        retDist = a[0]
+    else:
+        retDist = b[0]
+    retC = max(a[2], b[2]) # Pick the higher color
+    return (retDist, retAdj, retC)
+
+
+# Input is graphRDD
+# Returns (numComponents, maxCompSize)
+def findComponents(graphRDD):
```
LeonhardFS, line 42: Did you try this code and how fast it runs? Like how many steps were actually doable with it?

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,192 @@
...
+INF = 999999999
+# issues = {}
+# charLookup = {}
+# revLookup = []
+# adjList = {}
+# heroCount = 0
+numPart = 24
+updates = sc.accumulator(0)
+
+# # Read CSV file
```
LeonhardFS, line 22: please remove uncommented code before final submission ;)

HW1/P2/P2b.py
```diff
@@ -1,3 +1,26 @@
...
-# Your code here
+from pyspark import SparkContext
+import time
+import random
+
+sc = SparkContext(appName="P2", pyFiles=[])
+
+in_list = [( (i, j), (i, j) ) for i in xrange(2000) for j in xrange(2000)]
+random.shuffle(in_list)
+
```
adisuissa, line 11: You could have created the ranges in spark (using sc.range), and then repartition (instead of shuffle).

HW1/P3/P3.py
```diff
@@ -0,0 +1,9 @@
+# Q3
+wlist = sc.textFile('EOWL_words.txt')
+
+maxAnagrams = wlist.map(lambda x: (''.join(sorted(x)), [x]))\
+                    .reduceByKey(lambda x, y: x+y)\
+                    .map( lambda (s, v): (s, len(v), v) )\
+                    .reduce( lambda x, y: x if x[1] > y[1] else y )
```
adisuissa, line 7: Nice use of the last reduce to find the one with the best value.

HW1/P4/P4.py
adisuissa, line 45: Indeed the adjacency list is in a python construct, instead of a spark one.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,52 @@
+# References:
+# http://hadooptutorial.wikispaces.com/Iterative+MapReduce+and+Counters
+# http://www.slideshare.net/jhammerb/lec5-pagerank
+# The CSV and Spark manual pages
+
+INF = 999999999
```
adisuissa, line 6: Should probably be a negative value to make the code more general.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,52 @@
...
+
+# Map: emit neighbour nodes if GRAY, otherwise
+# just emit the same thing if WHITE or BLACK
+def processNode((k,v)):
+    curDist, adj, col = v
+    ret = []
+    if col == 1: # Emit all neighbouring nodes and cur
+        for nb in adj:
```
adisuissa, line 14: You should have avoided doing this in python, but use spark's functions (such as join, subtractByKey and union) instead.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,192 @@
...
+#             adjList[c1].add(c2)
+#             adjList[c2].add(c1)
+
+# Map: emit neighbour nodes if GRAY, otherwise
+# just emit the same thing if WHITE or BLACK
+def processNode((k,v)):
+    curDist, adj, col, prevP = v
+    ret = []
+    if col == 1: # Emit all neighbouring nodes and cur
+        for nb in adj:
```
adisuissa, line 55: Same as in P4.

HW1/P6/P6.py
```diff
@@ -0,0 +1,77 @@
...
+        allWords += clean(line)
+
+print 'Creating RDD...'
+
+rddList = []
+numP = 2
+forW = 3
+
+for i in xrange(0, len(allWords)-forW):
+    rddList.append( tuple(allWords[i:i+forW]) )
```
adisuissa, line 37: Creating the words should have been done in spark, and not in python.

HW1/P6/P6.py
```diff
@@ -0,0 +1,77 @@
...
+def randPick(words):
+    norm = 0.
+    r = random.uniform(0.,1.)
+    for k,v in words:
+        norm += v
+    for k,v in words:
+        r -= float(v) / norm
+        if r < 0:
+            return k
+    return "ERROR"
```
adisuissa, line 56: You could use numpy's random.choice to clarify the code.

PR #220, opened 2015-10-05
Hw1

Elena-Zhao:


HW1/P3/P3.py
```diff
@@ -0,0 +1,28 @@
...
+wlist = sc.parallelize(lines, 20)
+
+#Return RDD in the form (SortedLetterSequence, NumberOfValidAnagrams)
+sequence_num_RDD = wlist.map(lambda word: (''.join(sorted(list(word))),1)).reduceByKey(lambda x,y: x+y)
+
+#Return RDD in the form (SortedLetterSequence1, [Word1a, Word2a, ...])
+sequence_words_RDD = wlist.map(lambda word: (''.join(sorted(list(word))),''+ word)).groupByKey().map(lambda x : (x[0], list(x[1])))
+
+#Join the two RDDs above and map to designated format
+result = sequence_num_RDD.join(sequence_words_RDD).map(lambda x: (x[0], x[1][0], x[1][1]))
```
tobytoyuito: I think you don't need to use another RDD to calculate NumberOfValidAnagrams because you already have a list for each sequence. You can use len(x[1]) as NumberOfValidAnagrams.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,114 @@
...
+    # initial state
+
+    MC_Graph_bfs = MC_Graph.map(lambda (x, y): (x, (y, 0, False)) 
+                                if (x == init_node) 
+                                else (x, (y, float("inf"), False)))
+
+    print MC_Graph_bfs.filter(lambda (x, y): x == init_node).collect()
+    # Convert MC_Graph_bfs to adictionary and broadcast it every worker
+    # This is used for lookup during each update
+    tmp = dict(MC_Graph_bfs.collect())
```
tobytoyuito: I think you can comment this out because you didn't use this dictionary any more.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,114 @@
...
+        # other nodes (nodes used for update must already have their min distance set)
+        #
+        # The update is done by calling update_graph graph function (see below), and reduce 
+        # the result to get minimum distance for each single node 
+        #########################################################################
+
+        updated_nodes = MC_Graph_bfs.filter(lambda x: 
+            x[1][2] == False).flatMap(
+            partial(update_graph, 
+                mc_graph = dict(MC_Graph_bfs.collect()), accum = accum)).groupByKey().mapValues(reduce_nodes)
```
tobytoyuito: I think it will be faster if you don't use collect(). You can use accumulator to help you remember the visited nodes.

HW1/P6/P6.py
```diff
@@ -0,0 +1,60 @@
...
+
+def generateLine(word_count):
+    init_key = resultRDD.keys().takeSample(True, 1)[0] 
+    phrase = init_key[0] + ' ' + init_key[1]
+    for i in xrange(word_count - 2):
+        try:
+            new_word = resultRDD.map(lambda x: x).lookup(init_key)[0][0][0]
+        except: 
+            init_key = resultRDD.keys().takeSample(True, 1)[0]
+            new_word = resultRDD.map(lambda x: x).lookup(init_key)[0][0][0]
```
tobytoyuito: This is good to prevent error if a bigram does not exist. PS: It is interesting because you were not randomly generating bigram out of nowhere. You were generating from the original text. Only the last bigram was possible to raise this error.

HW1/P6/P6.py
```diff
@@ -0,0 +1,60 @@
...
+        not x.isdigit()).filter(lambda x:
+        not x.isupper()).filter(lambda x:
+        x != '')
+
+    wlist_2 = wlist.collect()
+
+    word_seq = []
+
+    for i in xrange(len(wlist_2)-2):
+        word_seq.append(((wlist_2[i], wlist_2[i+1]),wlist_2[i+2]))
```
tobytoyuito: I did this too. Someone use rdd to do it in parallel.

HW1/P4/P4.py
```diff
@@ -0,0 +1,55 @@
...
+findspark.find()
+findspark.init('/usr/local/opt/apache-spark/libexec')
+import pyspark
+from P4_bfs import *
+
+from functools import partial
+import time
+
+# Helper function, used to generate a list of (key, value) tuples given a node set
+def generatePairs(node_set):
```
scheng95: could you have used cartesian, and then removed self-loops?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,114 @@
...
+                                else (x, (y, float("inf"), False)))
+
+    print MC_Graph_bfs.filter(lambda (x, y): x == init_node).collect()
+    # Convert MC_Graph_bfs to adictionary and broadcast it every worker
+    # This is used for lookup during each update
+    tmp = dict(MC_Graph_bfs.collect())
+    #broadcast_graph = sc.broadcast(tmp)
+
+    # Accumulator set to 0
+    accum = sc.accumulator(0)
```
scheng95: there's not really an advantage to using accumulators like this, you might as well have just had a normal count variable

PR #221, opened 2015-10-05
Hw1

kevinkeyjkw:


HW1/P2/P2.txt
```diff
@@ -0,0 +1,9 @@
+# Your discussion here
+In the default partitioning strategy, almost 90% of the partitions were doing near 0 calculations. The other 10% were
```
ankitvgupta: Nice.

HW1/P2/P2.txt
```diff
@@ -0,0 +1,9 @@
...
+doing most of the work, while the rest were just sitting there wasting their processing power. 
+
+For my partitioning strategy, I randomized my x,y coordinates before taking the cartesian. Thus my matrix would contain
+all the points in some random order. This randomization created a more evenly distributed load across the executors as 
+can be seen from P2b.png. Most of the executors have a medium-sized load, thus there is not much wasted processing power.
```
ankitvgupta: Cool, that makes sense.

HW1/P2/P2b.py
```diff
@@ -0,0 +1,16 @@
...
+sc=SparkContext()
+xrand=range(2000)
+yrand=range(2000)
+random.shuffle(xrand)
+random.shuffle(yrand)
+x = sc.parallelize(xrand,10)
+y = sc.parallelize(yrand,10)
+xy = x.cartesian(y)
+result = xy.map(lambda t:[t,mandelbrot((t[0]/500.0)-2,(t[1]/500.0)-2)])
+draw_image(result)
```
ankitvgupta: I would definitely recommend you add more comments to your code. Granted, in this case, your code was pretty simple, it may have been harder to understand long-term.

HW1/P3/P3.py
```diff
@@ -0,0 +1,11 @@
+from pyspark import SparkContext
+import pdb
+
+sc = SparkContext()
+textFile = sc.textFile("EOWL_words.txt")
+#Sort the letters of each word and make that key, and the original word the value
+sortedTf = textFile.map(lambda word: (''.join(sorted(word)),word))
+#Group words that are the same anagrams together with the count
+v = sortedTf.groupByKey().map(lambda x: (len(list(x[1])),x[0],list(x[1]))).reduce(lambda a,b:max(a,b))
```
ankitvgupta: If you have a KV, it is a bit cleaner to do something like .map(lambda (k, v):
instead of .map(lambda x:
and then having to access x[0] and x[1] separately.

HW1/P4/P4.py
```diff
@@ -0,0 +1,29 @@
...
+import matplotlib.pyplot as plt
+import itertools
+import pdb
+
+from P4_bfs import bfs
+
+sc = SparkContext()
+textFile = sc.textFile('source.csv',20)
+
+#startNode = 'CAPTAIN AMERICA'
```
ankitvgupta: style nit - remove commented code.

HW1/P4/P4.py
```diff
@@ -0,0 +1,29 @@
...
+startNode = 'ORWELL'
+graph = (
+	textFile
+	.map(lambda x:(x.split('"')[3],x.split('"')[1]))
+	.groupByKey()
+	.map(lambda t:(t[0],list(t[1])))
+	.flatMap(lambda f:[(y,tuple([x for x in f[1] if x != y])) for y in f[1]])
+	.groupByKey().map(lambda f:(f[0],10000,frozenset([inner for outer in f[1] for inner in outer]),False))
+	.map(lambda x:(x[0],0,x[2],True) if x[0] == startNode else x)
+	.cache()
```
ankitvgupta: Looks like your code works, but it is extremely hard to follow. If your argument to a lambda function is a tuple, consider representing that as lambda (t1, t2, t3, ...) instead of lambda x and then having to access x[0] and x[1]. It makes code way more readable!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,36 @@
+def extraArgs(neighbors,distance):
+	"""If a node is in neighbors, and it hasn't been visited, return it. Else return"""
+	def updateNode(n):
```
ankitvgupta: Maybe I'm just not familiar with this style, but what is the purpose of defining this function inside the other?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,36 @@
...
+				return (n[0],distance+1,n[2],True)
+			else:
+				return (n[0],distance+1,n[2],n[3])
+		else:
+			return n
+	return updateNode
+
+def f(neighbors):
+	return lambda y:y[3]==False and y[0] in neighbors
+
```
ankitvgupta: Definitely need a comment explaining what this does? What does the 3rd index mean?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,36 @@
...
+def bfs(graph,startNode):	
+	print 'Start Node:',startNode
+	#greyNodes are the nodes currently being visited. Get the starting node and its neighbors
+	greyNodes= set(graph.filter(lambda x:x[0]==startNode).collect())
+	for i in range(4):
+		#Store neighbors of currently being visted nodes
+		neighbors = set([])
+		print 'Nodes touch on iteration',i,len(greyNodes)
+		#Add the neighbors of each currently being visited node's neighbors to greyNodes
+		for x in greyNodes.copy():
```
ankitvgupta: Why is the .copy() needed?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,36 @@
...
+	return updateNode
+
+def f(neighbors):
+	return lambda y:y[3]==False and y[0] in neighbors
+
+def bfs(graph,startNode):	
+	print 'Start Node:',startNode
+	#greyNodes are the nodes currently being visited. Get the starting node and its neighbors
+	greyNodes= set(graph.filter(lambda x:x[0]==startNode).collect())
+	for i in range(4):
```
ankitvgupta: Why are you only doing this for 4 iterations?

HW1/P4/P4_bfs.py
ankitvgupta: why is this copy needed?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,36 @@
...
+		#Store neighbors of currently being visted nodes
+		neighbors = set([])
+		print 'Nodes touch on iteration',i,len(greyNodes)
+		#Add the neighbors of each currently being visited node's neighbors to greyNodes
+		for x in greyNodes.copy():
+			neighbors.update(x[2])
+			greyNodes.remove(x)
+		#Retrieve the nodes in neighbors
+		greyNodes.update(graph.filter(f(neighbors)).collect())
```
ankitvgupta: You definitely want to avoid this collect() call. Think about how you can use RDD joins and accumulators to do this in parallel!

HW1/P5/P5.txt
```diff
@@ -0,0 +1,9 @@
+# Your discussion here
+For Kevin to Harvard, I was able to run bfs on my local machine in 2 minutes. But I am only able to run bfs from Harvard to Kevin on AWS.
+Using 5 instances, I was able to get the result in around 15 minutes. I don't think my code is parallized enough, there are a few collects
```
ankitvgupta: Yes, I think the collect is definitely one of the issues. Try to think about how to approach this program in a parallel way, not just in a serial way.

HW1/P3/P3.py
kevinkeyjkw: You're right, that would be much clearer to read. Thanks for the tip, I'll definitely remember to use that next time.

HW1/P4/P4_bfs.py
kevinkeyjkw: I was having a bug where the the current value of i was not being passed in when calling:
graph = graph.map(updateNode(neighbors,i))
 Instead, it would pass in the value of i only when I performed an action such as filter. (I think this has to do with spark's lazy evaluation)
So extraArgs allows me to fix this

HW1/P4/P4_bfs.py
kevinkeyjkw: Sorry for my lack of comments. y[3] tells me if the node has been visited or not. y[0] is name of node.
This function returns true if the node has not been visited and is in neighbors.

HW1/P4/P4_bfs.py
kevinkeyjkw: Should be 10. I changed it after I saw that at most 4 iterations were made.

HW1/P4/P4_bfs.py
kevinkeyjkw: Inside the loop, I'm changing the values of greyNodes. I get "Set changed size during iteration" error if I don't use copy.

HW1/P2/P2b.py
zezhouliu: Yes, I sometimes found it difficult to understand your code at a high level

HW1/P4/P4_bfs.py
zezhouliu: An accumulator will help you break out early!

zezhouliu:
Great comments @ankitvgupta

PR #223, opened 2015-10-05
Hw1

tobytoyuito:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,27 @@
...
+
+if __name__ == '__main__':
+    sc = pyspark.SparkContext(appName="Spark 2")
+    sc.setLogLevel('WARN') 
+
+    x = np.arange(1, 2001)
+    x = sc.parallelize(x)
+
+    pixels = x.cartesian(x)
+    pixels = pixels.partitionBy(100)
```
kevinkeyjkw: I'm not understanding how you distributed the load. Was it through partitionBy?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,23 @@
...
+            return v1
+    visited = sc.accumulator(set(), SetAccumulatorParam())
+    #visited = set()
+    not_visited = set([start])
+    iteration = 0
+    
+    while(len(not_visited)>0):
+        iteration += 1
+        pre = set(list(visited.value))
+        rdd.filter(lambda (K, V): K in not_visited).foreach(lambda (K, V): visited.add(V))
```
kevinkeyjkw: Great way to use accumulators! I didn't know until now.

HW1/P6/P6.py
```diff
@@ -0,0 +1,57 @@
...
+        if word.upper()==word: return False
+        tmp = word[:-1]
+        if len(word)>0 and word[-1]=="." and tmp.upper()==tmp: return False
+        return True
+
+    return [word for word in words if good_word(word)]
+
+words = lines.flatMap(filter_words).collect()
+print "making trigrams..."
+trigrams = sc.parallelize([[(words[i], words[i+1], words[i+2]), 1.] for i in range(len(words)-2)])
```
kevinkeyjkw: I used for loop as well. But I saw someone else zip each word with an index, subtract 1 from each index to get 2nd RDD, subtract 1 again from each index to get 3rd RDD. Then join three RDDs to get and RDD of (word1,word2,word3). For loops would work for a small data set like this, but join would probably be better for larger ones.

HW1/P2/P2b.py
Peilin-D: I'm also not clear about this. I tried partitionBy, but it didn't work well for me. I used repartition (random shuffle) instead and got good result.

oyuchoi, line 17: Your P2b distribution looks like it has a very tight bound, which is good!  The graph is a little deceptive since its bimodal, but its definitely much closer than the original.

HW1/P5/P5.txt
```diff
@@ -1 +1,5 @@
-# Your discussion here
```
oyuchoi, line 1: Did you get any other results? Or did it only find the first one?

PR #227, opened 2015-10-05
Hw1

Chongmoliu:


HW1/P2/P2.txt
```diff
@@ -0,0 +1,16 @@
...
+From figure P2a_hist, we found that jobs by using default partition are significantly unbalanced. Most partitions are distributed to the less iterations, very few partitions are for large iterations and most excution time are spent on the parts requiring more iterations. 
+
+In order to rebalance the task, we firstly try to use partitionBy(numPartitions=100, partitionFunc=lambda x: random.randint(0, 100)) to resuffle the data in pixel RDD randomly. From P2b1_hist, we found that jobs are better distributed for 200000-2500000 iterations, but there is still a few distributions distributed to large iterations which could be slow tasks and cause the computation to have poor parallelism. 
+
+After discussing with other classmates, we try to reshuffle the data before using parallelize method, which means:
+
+x=range(2000)
+random.shuffle(x)
+ii=sc.parallelize(x, numSlices=10). 
+
```
andrewreece: Nice idea to shuffle before parallelizing!  For your 1b implementation, consider `repartition()` instead of writing your own `partitionFunc`, as they both randomize elements to partitions.

HW1/P3/P3.py
```diff
@@ -0,0 +1,43 @@
...
+# In[17]:
+
+Jumble_RDD=newlist.map(lambda x: (x[0], len(list(x[1])), list(x[1])))
+
+
+# In[19]:
+
+max(Jumble_RDD.collect(), key=lambda x: x[1])
+
+
```
andrewreece: Commenting here is a bit sparse.  Nice that you included all of the words with the most anagrams.

HW1/P5/P5.txt
```diff
@@ -0,0 +1,9 @@
+# Your discussion here
+
+Finding Directed Paths:
+From “Kevin Bacon” to “Harvard University”, one of the shortest path is 
+[u'Kevin_Bacon', u'Marisa_Silver', u'Harvard_University'](takes only a few seconds)
+
+For the opposite direction, it'll take too much time. 
+
```
andrewreece: Calls to `collect` within your loop are the main cause of long compute times here.  Instead, consider keeping a running list of nodes you've already touched, which you can use to filter the main graph on each iteration.  Then, call `collect` only once when finished.

HW1/P2/P2.txt
annawhitney: Yeah, this is clever - did you benchmark the relative performance of shuffling using random.shuffle prior to parallelizing against a Spark shuffle? I'd certainly expect it to be faster this way, but I'd be interested in seeing the numbers.

HW1/P3/P3.py
```diff
@@ -0,0 +1,43 @@
...
+import pyspark
+from pyspark import SparkContext
+sc =SparkContext()
+
+
+# In[2]:
+
+import numpy as np
+import matplotlib.pyplot as plt 
+import matplotlib.cm as cm
```
annawhitney: You're not actually using matplotlib in P3, no? Not that it makes a huge difference, but removing imports that you're not using makes your code a bit more readable (and might save some time).

HW1/P4/P4.py
```diff
@@ -0,0 +1,67 @@
...
+# In[4]:
+
+import ast
+import itertools
+
+
+# In[5]:
+
+#convert the initial RDD (name, issue) into a RDD (issue, {all characters in the issue})
+series_RDD = sc.textFile('source.csv').map(lambda x: ast.literal_eval(x.encode('utf-8').decode('ascii', 'ignore'))[::-1]).groupByKey().mapValues(lambda x: set(x))
```
annawhitney: I like this way of parsing each line of the file with ast.literal_eval - saves you from actually mucking around with string parsing.

HW1/P4/P4.txt
```diff
@@ -0,0 +1,14 @@
+# Your discussion here
+
+1. for the graph representation, firstly convert the initial RDD into a RDD (issue, {all characters in the issue}). Then, convert the new RDD into a RDD (node, {neighbor1, neighbor2, ...}) whose values' type are all set. 
+
+2. the number of touched nodes:(including the start node itself)
+"CAPTAIN AMERICA": 6048
+"MISS THING/MARY": 7
+"ORWELL":9
+
+If a node doesn't have a defined distance, there will be no path between the start node and this node, which means the node's character has never appeared in the same comic issues as the character of the start node or of its neighbors. For example, the touched nodes for "CAPTAIN AMERICA" is 6048 and the total nodes number is 6444. Thus, only a few characters have never been in the same issues as "CAPTAIN AMERICA" or as its neighbors.
```
annawhitney: Not just its neighbors, though, right? A node without a defined distance/no path between the nodes means that the characters has never appeared in the same issue as Captain America, any neighbors, neighbors of those neighbors, and so on indefinitely - in graph theoretical terms, it means that node is on a different connected component (shades of P5...)

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,68 @@
...
+# In[ ]:
+def BFS(G, T, start, goal): #G:graph RDD, (char1, [char2, char3, ...])    
+    queue = sc.parallelize([start])
+    prev ={}# prev[nd] = previous node in the path before nd
+    while not queue.isEmpty():
+        q=set(queue.collect())
+        neighbors =G.filter(lambda x: x[0] in q).flatMapValues(lambda x:x).filter(lambda x: not(x[1] in set(prev.keys()))).map(lambda x: (x[1],x[0])).reduceByKey(lambda x,y: x).distinct()
+        prev.update(dict(neighbors.collect())) 
+        #print prev
+        if goal in set(neighbors.keys().collect()):
```
annawhitney: By calling `collect` every iteration (multiple times, no less), you're basically defeating the purpose of doing this in Spark, because you're pulling from an RDD to a dict or a set every iteration.

HW1/P6/P6.py
```diff
@@ -0,0 +1,82 @@
...
+# In[18]:
+
+#combine the three RDD
+Triple_W_RDD=W1_RDD.union(W2_RDD).union(W3_RDD)
+
+
+# In[19]:
+
+from collections import Counter
+Markov_RDD=Triple_W_RDD.reduceByKey(lambda x, y: x + ' '+y).mapValues(lambda x: Counter(x.split()).items())
```
annawhitney: It's not clear to me what you're using `Counter` for here. Is there a way you could do this with RDD transformations rather than going into the `collections` module?

HW1/P4/P4.py
zezhouliu: Good use of groupByKey

HW1/P3/P3.py
```diff
@@ -0,0 +1,43 @@
...
+
+# In[3]:
+
+wlist = sc.textFile('EOWL_words.txt')
+
+
+# In[7]:
+
+#convert to a new RDD in the form: (SortedLetterSequence, [Word1a, Word2a, ...])
+newlist=wlist.keyBy(lambda x: ''.join(sorted(x))).groupByKey().sortByKey()
```
zezhouliu: Good!

HW1/P4/P4.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+# In[6]:
+
+#convert series_RDD into a RDD (node, {neighbor1, neighbor2, ...})
+neighbors_RDD=series_RDD.flatMap(lambda x: ((i, tuple(x[1].difference(set([i])))) for i in x[1])).reduceByKey(lambda x, y: x+y).mapValues(lambda x: set(x))
+
+
+# In[27]:
+
```
zezhouliu: Using an accumulator makes this more efficient!

HW1/P4/P4.txt
zezhouliu: @annawhitney is correct that not showing up as a touched node implies that it's not connected (meaning not associated with) any of the comics that originate from Captain America

HW1/P5/P5.txt
zezhouliu: Agree with this comment.  @annawhitney added a comment directly in your P5.py file about calling collect.

HW1/P4/P4.txt
```diff
@@ -0,0 +1,14 @@
+# Your discussion here
+
+1. for the graph representation, firstly convert the initial RDD into a RDD (issue, {all characters in the issue}). Then, convert the new RDD into a RDD (node, {neighbor1, neighbor2, ...}) whose values' type are all set. 
+
+2. the number of touched nodes:(including the start node itself)
+"CAPTAIN AMERICA": 6048
```
zezhouliu: 6408?

PR #228, opened 2015-10-05
finished homework 1

zhiqianseah:
finished homework 1


HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,72 @@
...
+	counter = 0
+	while (accum.value != 0): 
+
+		#we only need to work with those records that hasn't been reached yet
+		filtered_rdd = bfs_rdd.filter(lambda (x,(y,z, remove)): remove == 0)
+		child_expanded_rdd = filtered_rdd.flatMap(lambda (x,(y,z, remove)): check_and_spawn( (x,(y,z, remove)), counter))
+		#print child_expanded_rdd.count()
+
+		#reset accumulator
+		accum = sc.accumulator(0)
```
rohitahuja: It would also be fine to check if your accumulator changed since the previous iteration

HW1/P6/P6.py
```diff
@@ -0,0 +1,91 @@
...
+#result = first_three_words_4.filter(lambda ((x,y), z): (x == 'Now') & (y == 'is'))
+print result
+
+
+#----Now, to generate the Markov chain!-------------------------------
+
+for counter in range(0,10):
+	final_word_list = []
+	#take a random sample. NOTE: i am working with the list of 3rd words BEFORE they are counted
+	#as it is easier to take a random choice from there
```
rohitahuja: I believe the random choice should be weighted based on the counts of the words, since intuitively, that's how a Shakespearean sentence should be constructed (a word that Shakespeare uses 100 times shouldn't have the same probability of getting chosen as a word used 2 times). A PDF of the word counts should suffice.

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,160 @@
...
+results_index = bfs(rdd_intersect2, sc)
+
+print results_index
+print >> f1, results_index
+
+print max(results_index)
+print >> f1, max(results_index)
+
+print time.time() - start
+print >> f1, time.time() - start
```
rohitahuja: Interesting BFS method. I use a similar method to expand nodes and then reduce them.

HW1/P2/P2.txt
```diff
@@ -1 +1,4 @@
 # Your discussion here
+From the graph in P2a, it is clear that the number of mandelbrot iterations varies distinctly based on its coordinates. To balance the amount of work across all the partitions, my strategy is to randomly shuffle the list of input coordinates. With such a large number of data points (4000000) randomly sorted into a small number of partitions (100), it is likely that each partition has a similar percentage of coordinates with high versus low iterations. From the graph in P2b, the work is balanced really well, with total iterations in the range of [1960000, 2100000], a spread of only 250000. In contrast, the total iterations in P2a_hist covers the a wider range of [0,9000000]
```
rohitahuja: I also randomly shuffled the input coordinates. One way that I had also thought about after submission was to uniformly map the high iteration points to partitions, but not sure if the preprocessing cost of that is a bottleneck

HW1/P6/P6.py
zhiqianseah: Hi Robit, it is hard to see from the code alone, but I did take a weighted random sample of the 3rd word. The RDD that I use to sample contains a list with repeated entries (ie ['a', 'b', 'a', 'c', 'c', 'b']), hence a simple sample from this list is already a weighted sample. Honestly, the Counter version just makes things more complicated, so I revert to a previous RDD for doing the markov chain.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,39 @@
...
+mandel_image = img.map(lambda (i,j): ((i,j), mandelbrot(((j/500.0)-2), ((i/500.0)-2)))) 
+
+#sum of values in partitions
+sum_data = sum_values_for_partitions(mandel_image)
+plt.hist(sum_data.collect())
+plt.show()
+
+#uncomment this to draw the image
+#draw_image(mandel_image)
+
```
xdai1091: The plot looks good to me :)

HW1/P4/P4.py
```diff
@@ -0,0 +1,63 @@
...
+#--------------------------------------------------------------------------
+
+start = time.time()
+#Toggle for the different superheroes to start with
+#results = bfs(vertices3, "ORWELL", sc)
+#results = bfs(vertices3, "MISS THING/MARY", sc)
+results = bfs(vertices3, "CAPTAIN AMERICA", sc)
+print time.time()-start
+
+print results
```
xdai1091: easy to follow!

PR #230, opened 2015-10-05
Hw1 Submissions

hyunkwang:
I have done with HW1 except the connected components in Problem 5.


HW1/P2/P2.txt
```diff
@@ -1 +1,10 @@
 # Your discussion here
+For this problem 2, I generated bar plots with partitions for X axis and corresonding efforts for Y axis instead of histograms mentioned in the HW1 manual so that I can easily see how much each effort is distributed roughly the same on each partition.
```
matheuscfernandes: I am not understanding what you mean. From what I see, you did plot the histograms, but the difference is what you are plotting in the y-axis?

HW1/P2/P2.txt
```diff
@@ -1 +1,10 @@
...
+
+I named them P2a_bar.png and P2b_bar.png. 
+
+[Discussions]
+  As you can see P2a_bar, each effort for partitions are not equally distributed because of the intrinsic property of mandelbrot function. This function does more computations only for when the absolute value of z (= norm of z) is smaller than 2 (See conditions for the while loop). So, for partitions which have pixels whose norms are less than 2, much more efforts need to be made, but not for others. 
+  After the RDD repartitioned (See P2b_bar.png), each effort of each partition is roughly the same, which means this system is load balanced. The reason is that repartitioning helps distributing "stragglers", or pixels whose norms are smaller than 2, roughly equally over the entire partitions. So, when computation times compared, the repartitioned one takes much faster than the original one.
+ But, in order to repartition the task, tasks need to be reshuffled, which is quite a bit cost. For this problem, the cost is not noticable, but if we deal with much bigger data, the shuffling could be a bottleneck for performance.
+
```
matheuscfernandes: That is a great explanation. Pretty much exactly my reasoning. +1

HW1/P2/P2a.py
```diff
@@ -1,3 +1,33 @@
+# Initialize SC context
+import findspark
+findspark.init()
+import pyspark
+sc = pyspark.SparkContext(appName="Spark 1")
```
matheuscfernandes: Just for future info, you can get the total number of cpu's and pass it along to spark.  

`config = ps.SparkConf()
config = config.setMaster('local[' + str(2*mp.cpu_count()) + ']')
config = config.setAppName('P2b')

sc = ps.SparkContext(conf=config)`

HW1/P2/P2a.py
```diff
@@ -1,3 +1,33 @@
...
+  # Plot a histrogram (Also plot a bar)
+  ## For histrogram
+  #plt.hist(seq_of_efforts)
+  #plt.savefig("P2a_hist.png")
+  # For Bar plot
+  ind = np.arange(len(seq_of_efforts))
+  plt.bar(ind+0.2,seq_of_efforts,0.6)
+  plt.savefig("P2a_bar.png")
+
+
```
matheuscfernandes: Pretty much exactly what I did.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,35 @@
...
 # Your code here
+if __name__ == '__main__':
+  # Set up Initial RDDs with 100 partitions
+  rdd = sc.parallelize(range(2000),10)
+  rdd1 = sc.parallelize(range(2000),10)
+  rdd2 = rdd1.cartesian(rdd)
+
+  # RDD repartitioned using partitionBy
+  rdd2_new = rdd2.partitionBy(100)
+
```
matheuscfernandes: I think you could have even done a better distribution if you had assigned it randomly rather than by the order of partitioning that partitionBy gives you.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+import pyspark
+sc = pyspark.SparkContext(appName="Spark 1")
+
+# Your code here
+if __name__ == '__main__':
+  # Set up Initial RDDs
+  wlist = sc.textFile("EOWL_words.txt",100)
+  rdd = wlist.map(lambda word: "".join(sorted(word)))
+  # Find dictionary "sequence" : count
+  count_result = rdd.countByValue()
```
matheuscfernandes: That's good. I didn't know you could do that.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+  # Find dictionary "sequence" : list of anagrams
+  list_of_result = wlist.groupBy(lambda word: "".join(sorted(word))).collectAsMap()
+  # Map rdd to (SortedLetterSequence1, NumberOfValidAnagrams1, [Word1a, Word2a, ...])
+  rdd_final = rdd.map(lambda seq: (seq,count_result[seq],list(list_of_result[seq])))
+  
+  print max(rdd_final.collect(),key=lambda item:item[1])
```
matheuscfernandes: I think it would be nice to have these sorted so that we all had the same order. It just made sense to me to alphabetically sort, so before this point I 'sorteByKey' the RDD. Nothing is wrong with your's just comparing yours to mine. We did not have the same word  because of that, but the count was 11 for mine too.

HW1/P3/P3.txt
```diff
@@ -1 +1,3 @@
 # Your discussion here
+A Single entry from the RDD
+(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])
```
matheuscfernandes: I had 11 too, but I have a different word because of sorting it before.

HW1/P4/P4.txt
```diff
@@ -1 +1,27 @@
...
+2) Root Node : MISS THING/MARY 
+The number of touched nodes : 6
+The number of untouched nodes : 6438
+Diameter : 1
+
+3) Root Node : ORWELL
+The number of touched nodes : 8
+The number of untouched nodes : 6436
+Diameter : 1
+
```
matheuscfernandes: I got the same thing for the final numbers. Only I accounted for the character itself (Node0).

HW1/P4/P4.txt
```diff
@@ -1 +1,27 @@
...
+Questions
+Q1. For the purposes of this problem, you may assume that the diameter of this graph is ten (10). What does this diameter imply about the maximum number of steps in which your search must be executed?
+Answer : diameter is the maximum distance from a root node. As diameter gets larger, the time that needs to track leaf nodes is more increased. In other words, the complexity of the graph computations is increased. 
+
+Q2. What does it mean if a character does not have a defined distance during the search?
+Answer : That means a character is not connected to a root node through all nodes. it cannot search the character from root node.
```
matheuscfernandes: That also means that they may have their internal loop. For example they are two different authors for the comics. They may have their own little loop that doesnt connect to any of the other characters.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,44 @@
...
+        break
+      else:
+        # Create a list of neighbors of nodes at a current level
+        rdd_queue = rdd_queue.map(lambda (node, (dist, neighbors)): neighbors)
+        neighbors_at_level = rdd_queue.reduce(lambda x,y: list(set(x+y)))
+        # Update distances of the neighbors 
+        rdd_graph = rdd_graph.map(update_dist(level, neighbors_at_level))
+        # Go to next level
+        accum.add(1)
+        level += 1
```
matheuscfernandes: I dont understand why you are using both the accumulator and the level variable? Shouldn't both have the same functionality?

HW1/P5/P5.txt
```diff
@@ -1 +1,14 @@
 # Your discussion here
+
+1. Short Path
+1) ”Kevin_Bacon" -> "Harvard University"
+Using Labels : 2729536 -> 2578703 -> 2152782
+Using Paget Titles : Kevin_Bacon -> John_Lithgow -> Harvard_University
+
+2) "Harvard University" -> "Kevin Bacon"
+Using Labels : 2152782 -> 4530382 -> 44868 -> 2729536
+Using Paget Titles : Harvard_University -> September_8 -> 1958 -> Kevin_Bacon
```
matheuscfernandes: I got the same number of steps, but I did not get the same path. I think it is OK though to have multiple different answers to this as there may be different paths one can take.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,85 @@
...
+      # flatMap graph to RDD 
+      rdd_graph = rdd_graph.flatMap(lambda (node, (dist, parent, neighbors)) : update_child_links(node, dist, parent, neighbors, level))
+      rdd_graph = rdd_graph.reduceByKey(lambda x, y: update_graph(x,y))
+
+      # if distance of end_node is not infinite number
+      if rdd_graph.map(lambda x: x).lookup(end_node)[0][0] != float('inf'):
+        break
+      level += 1
+
+    level += 1
```
matheuscfernandes: Why did you remove your accumulator from the past problem. It is supposedly significantly faster than doing the adding to the variable yourself.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,85 @@
...
+
+    # Convert short path using labels to one using page titles
+    pages_for_short_path = []
+    short_path = short_path[::-1]
+    for i in range(len(short_path)):
+      page = rdd_pages.map(lambda x : x).lookup(short_path[i])[0]
+      pages_for_short_path.append(page)
+
+    # Path (Using Labels, Using Page Titles)
+    return (" -> ".join(short_path), " -> ".join(pages_for_short_path))
```
matheuscfernandes: Looks great :+1:

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
...
+  print random_phrases[0]
+  print random_phrases[1]
+  print random_phrases[2]
+  print random_phrases[3]
+  print random_phrases[4]
+  print random_phrases[5]
+  print random_phrases[6]
+  print random_phrases[7]
+  print random_phrases[8]
+  print random_phrases[9]
```
matheuscfernandes: Very good! :100:

HW1/P6/P6.txt
```diff
@@ -1 +1,12 @@
...
+[01] think no. Enter Cassio and some have greatness thrust upon them.' was one, sir, that you have a good soldier
+[02] that huge spirit now is the very same; the very same; the very same; the very same; the very same;
+[03] ebbs, his flows, as if he be not so much as to say the truth, so did you, Bardolph. You
+[04] thankfully. In himself, 'tis much; In you, which is the very same; the very same; the very same; the very
+[05] him pight to do with the and the rest of the world, And rouse from sleep that fell poison which
+[06] We'll draw cuts for the love of God, very religiously; and, as you are a gentleman of the world, And
+[07] 'The sleeves curiously cut.' Ay, there's the point, sir. Marry is it; the very same; the very same; the very
+[08] Could fetch your money. Come, come, you are a gentleman of the world, And rouse from sleep that fell poison
+[09] up afloat, Whilst he upon whose weal depends and rests The lives of those that have not seen the day
+[10] too strict a paradox, Striving to better, oft we pity enemies. Why, then, let the world is not so much
```
matheuscfernandes: Yours are funnier than mine. Haha. :)

HW1/P2/P2.txt
ytang14: This is also what I got. Nice reasoning, especially on the shuffling part. I tried to think about the trade off- after which threshold the shuffling cost became more imperative than the unequally partition cost?

HW1/P2/P2a.py
ytang14: Thank you  matheuscfernandes! I didn't notice that when I wrote the code

HW1/P2/P2a.py
ytang14: Same as mine too

HW1/P2/P2b.py
ytang14: you should just use random to get a better result, because the data set is very large. I produced pretty much a normal distribution.

HW1/P3/P3.txt
ytang14: I got the same answer as you did. I guess there are three words that have 11 anagrams. So any of them should be correct.

HW1/P4/P4.txt
ytang14: I got the same answers

HW1/P4/P4.txt
ytang14: Agreed. matheuscfernandes gives a point that I did not even considered about

HW1/P4/P4_bfs.py
ytang14: I used a counting variable similar to your 'level' variable and I did not used accumulator. However, the instructor said we are required to use accumulator, so I only used the add function of an accumulator..

HW1/P5/P5.txt
```diff
@@ -1 +1,14 @@
...
+
+2. Connected Components
+I couldn't finish it because it needs to do lots of experiments with AWS while we are not allowed to run
+AWS instances for a long time because of free credits.
```
ytang14: Me too, I think my algorithm is correct, but the AWS produced way too much more problems than I expected

HW1/P6/P6.py
ytang14: This is correct. Just out of curiosity, how long does it take for your code to run?

HW1/P2/P2b.py
```diff
@@ -1,3 +1,35 @@
...
+findspark.init()
+import pyspark
+sc = pyspark.SparkContext(appName="Spark 1")
 from P2 import *
 
 # Your code here
+if __name__ == '__main__':
+  # Set up Initial RDDs with 100 partitions
+  rdd = sc.parallelize(range(2000),10)
+  rdd1 = sc.parallelize(range(2000),10)
```
adisuissa, line 12: You should use sc.range instead of creating the ranges and then parallelizing them.

HW1/P4/P4.py
```diff
@@ -0,0 +1,25 @@
...
+
+# Your code here
+if __name__ == '__main__':
+  # Set up Initial RDDs
+  MC_data = sc.textFile("data.txt",100)
+  rdd_data = MC_data.map(lambda dataset: (dataset.split("\",\"")[0].split("\"")[1],dataset.split("\",\"")[1].split("\"")[0]))
+  # rdd_graph will be (K, V) RDD with K = node and V = [connecting nodes]
+  rdd_graph = rdd_data.groupBy(lambda data: data[1])\
+            .map(lambda x: [cha for cha, comic in list(x[1])])\
+            .flatMap(lambda cha_list: [(item,list(set(cha_list)-set([item]))) for item in cha_list])\
```
adisuissa: Instead of moving the data to python and processing it there, you should just use spark's functions (joins, subtractByKey and unions).

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,44 @@
+def update_dist(level, neighbors_at_level):
+    '''
+    Update Distances of all nodes given a current level and neighbors at next level
+    '''
+    return lambda (node, (dist, neighbors)): (node, (min(dist,level+1), neighbors)) if node in neighbors_at_level else (node, (dist, neighbors))
```
adisuissa, line 5: You could have used spark's functions, such as join, subtractByKey, and union, to do this (instead of python).
You could also update the accumulator here (to count the number of nodes with new distance values).

HW1/P5/P5_bfs.py
adisuissa, line 56: Actually, it won't do much of a difference as the accumulator is used just for storing the number of iterations.
If the accumulator was used to count the number of nodes with a new distance value (during a map or a reduce phase), it would have been useful.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,85 @@
...
+    '''
+    Given a (K, V) RDD with K = node and V = [connecting nodes],
+    find the shortest paths from a source node to all other nodes in a graph.
+    '''
+    # rdd_graph => (node, (distance, parent, neighbors)) RDD
+    rdd_graph = rdd_graph.map(lambda (node, neighbors): (node, (0, None, neighbors)) if (node == start_node) else (node, (float('inf'), None, neighbors)))
+    level = 0
+    while True:
+      # flatMap graph to RDD 
+      rdd_graph = rdd_graph.flatMap(lambda (node, (dist, parent, neighbors)) : update_child_links(node, dist, parent, neighbors, level))
```
adisuissa, line 48: Again, could have used joins and other rdd functions to get the same behavior, and possibly decrease the amount of data that is being stored in the RDD.

HW1/P6/P6_func.py
```diff
@@ -0,0 +1,17 @@
...
+  Generate New Random Phrases with 20 words given randomly-chosed Word1_Word2
+  '''
+  Phrases = list(Word1_Word2)
+  for i in range(18):
+    if i == 0:
+      New_Word1_Word2 = Word1_Word2
+    # Look up [(Word3, Count3)] for (Word1, Word2)
+    Word3_Count3 = rdd.map(lambda x : x).lookup(New_Word1_Word2)[0]
+    Word3_Count3.sort(key=lambda (word3, count3): count3)
+    Chosen_Word3_Count3 = Word3_Count3[-1]
```
adisuissa, line 12: You should have chosen the 3rd word using to a biased distribution (according to the number of appearances).

PR #231, opened 2015-10-05
Hw1

isadoranun:


HW1/P2/P2a.py
```diff
@@ -1,3 +1,16 @@
 from P2 import *
 
-# Your code here
+import findspark
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext()
```
matheuscfernandes: Just for future info, you can get the total number of cpu's and pass it along to spark.

`config = ps.SparkConf()
config = config.setMaster('local[' + str(2*mp.cpu_count()) + ']')
config = config.setAppName('P2b')

sc = ps.SparkContext(conf=config)`

HW1/P2/P2a.py
```diff
@@ -1,3 +1,16 @@
...
+
+tuples = []
+for i in xrange(2000):
+    for j in xrange(2000):
+        tuples.append((i,j))
```
matheuscfernandes: I think that instead of doing this for loop, they wanted you to use RDDs and the .cartesian() function for creating these tuples. Here is what I did instead:

`row = sc.parallelize(range(0,2000),10)
column = sc.parallelize(range(0,2000),10)

# this creates all possible combinations of rows and columns
tuples = row.cartesian(column)`

HW1/P2/P2b.py
```diff
@@ -1,3 +1,9 @@
 from P2 import *
 
-# Your code here
+import findspark
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext()
+
+partitions2 = sc.parallelize(tuples).map(lambda xy: (xy, mandelbrot((xy[1]/500.0)-2,(xy[0]/500.0)-2))).partitionBy(100, lambda xy: random.randrange(0, 100, 1))
```
matheuscfernandes: I did the same thing by using the random integers. However, I divided it from 1 to 101, because I did not think there was a 0. When I had the 0, it would give me an error every once in a while.

HW1/P3/P3.py
```diff
@@ -0,0 +1,13 @@
...
+import pyspark
+sc = pyspark.SparkContext()
+
+wlist = sc.textFile('./EOWL_words.txt')
+
+wlist2 = wlist.map(lambda line: (''.join(sorted(line)), (line)))
+
+wlist3 = wlist2.groupByKey().mapValues(lambda x: [i for i in x ]).map(lambda y: (y[0],len(y[1]),y[1]))
+
+wlist3.takeOrdered(1,lambda x: -x[1])
```
matheuscfernandes: Looks great! :100: I have the same number, only I got a different word because I sorted in alphabetical order prior to taking it.

HW1/P3/P3.txt
```diff
@@ -1 +1,15 @@
...
+   u'aster',
+   u'rates',
+   u'reast',
+   u'resat',
+   u'stare',
+   u'stear',
+   u'strae',
+   u'tares',
+   u'tears',
+   u'teras'])]
```
matheuscfernandes: I had the same number, but not the same word.

HW1/P4/P4.txt
```diff
@@ -1 +1,16 @@
...
+
+If the diameter of the graph is ten, it means that the maximum number of iterations that BFS can do is 10 since in each iteration we search for the nodes that are at “i” distance from the starting node.
+
+“Count the number of nodes touched in each of the above searches and place these counts in your writeup. What does it mean if a character does not have a defined distance during the search?”
+
+When applying the BFS code to the created graph, we obtain the following results. The number of nodes touched for:
+- Captain America is 6408
+- Orwell is 9
+- Miss thing/Mary is 7
+
```
matheuscfernandes: I got the same numbers of characters touched, though I would like to point out that it is interesting to analyze the diameter of each character, given that we discussed it in the question.

HW1/P4/P4.txt
```diff
@@ -1 +1,16 @@
-# Your discussion here
+“A description of your graph representation, and the number of touched nodes in each of the searches above. Also, discuss the two questions above (re. graph diameter, and untouched nodes).”
+
+The created graph has as keys all the possible nodes (superheroes). The value of each key is a list of characters that have appeared with this particular superhero in the same comic.
+
+“For the purposes of this problem, you may assume that the diameter of this graph is ten (10). What does this diameter imply about the maximum number of steps in which your search must be executed?”
+
+If the diameter of the graph is ten, it means that the maximum number of iterations that BFS can do is 10 since in each iteration we search for the nodes that are at “i” distance from the starting node.
+
```
matheuscfernandes: Great explanations. :+1:

HW1/P4/P4 bfs.py
```diff
@@ -0,0 +1,50 @@
...
+        it_acc += 1
+        dist = it_acc.value
+
+        nodes_2 = nodes.filter(lambda x: filterer(x[1])).join(graph).flatMap(lambda x: x[1][1]).distinct().map(lambda x: (x, dist + 1)).partitionBy(N)
+
+        nodes = nodes.cogroup(nodes_2).map(lambda x: (x[0], get_first_non_empty_cogroup(list(x[1])))).partitionBy(N).cache()
+        
+        assert copartitioned(nodes_2, nodes)
+
+        if nodes.count() == total_char_acc.value:
```
matheuscfernandes: Why did you not put this statement in the while clause? Why are you quitting here and not after the while?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,49 @@
...
+
+        nodes = nodes.cogroup(nodes_2).map(lambda x: (x[0], get_first_non_empty_cogroup(list(x[1])))).partitionBy(N).cache()
+        
+        assert copartitioned(nodes_2, nodes)
+
+        if nodes.count() == total_char_acc.value or dist == 10:
+            break
+
+        total_char_acc += nodes.count() - total_char_acc.value
+
```
matheuscfernandes: Is there a difference between this file and the P4 bfs.py?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,49 @@
...
+        dist = it_acc.value
+
+        nodes_2 = nodes.filter(lambda x: filterer(x[1])).join(graph).flatMap(lambda x: x[1][1]).distinct().map(lambda x: (x, dist + 1)).partitionBy(N)
+
+        nodes = nodes.cogroup(nodes_2).map(lambda x: (x[0], get_first_non_empty_cogroup(list(x[1])))).partitionBy(N).cache()
+        
+        assert copartitioned(nodes_2, nodes)
+
+        if nodes.count() == total_char_acc.value or dist == 10:
+            break
```
matheuscfernandes: Why did you not put this statement in the while clause? Why are you quitting here and not after the while?

HW1/P4/P4 bfs.py
```diff
@@ -0,0 +1,50 @@
...
+            break
+
+        total_char_acc += nodes.count() - total_char_acc.value
+        
+    return nodes.count(), nodes
```
matheuscfernandes: I just noticed that this file is redundant, and you do not call it. I further commented on your other BFS file.

HW1/P5/P5.txt
```diff
@@ -1 +1,38 @@
...
+10->7->4
+
+For visual simplicity,  I split the nodes based on the distance to the given starting node in the Wikipedia data.
+
+The distance from Kevin Bacon to Harvard University is 2 and the possible paths are:
+
+[u'Harvard_University', [u'Harvard_University', u'Time_(magazine)'], [u'Harvard_University', u'Marisa_Silver'], [u'Harvard_University', u'Six_degrees_of_separation'], [u'Harvard_University', u'John_Lithgow'], [u'Harvard_University', u'College_Bowl'], 
+
+[u'College_Bowl', u'Kevin_Bacon'], [u'John_Lithgow', u'Kevin_Bacon'], [u'Marisa_Silver', u'Kevin_Bacon'], [u'Six_degrees_of_separation', u'Kevin_Bacon'], [u'Time_(magazine)', u'Kevin_Bacon']]
+
```
matheuscfernandes: Very good that you listed different paths. The path mine took is included in your list. :)

HW1/P5/P5.txt
```diff
@@ -1 +1,38 @@
...
```
matheuscfernandes: Same here, thats awesome you got all of the combinations possible. :100:

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,119 @@
...
+                
+        nodes_4 = nodes_3.union(nodes_4)
+        
+        assert copartitioned(nodes_3, nodes_4)
+
+        nodes = nodes.cogroup(nodes_2).map(lambda x: (x[0], get_first_non_empty_cogroup(list(x[1])))).partitionBy(N).cache()
+
+        
+        if nodes_2.map(lambda x: x[0]).filter(lambda x: node_destination[0] == x).count() > 0:
+            print "Distance:", dist
```
matheuscfernandes: same as my last comment, you could put this directly in the while loop. It is a minor change but it makes the code a little cleaner.

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,129 @@
...
+    neighbor_symmetric = symmetric_graph(neighbor_graph)
+
+    [maxC2, nC2] = connected_components(neighbor_symmetric, N)
+
+    print "For the graph where each link is taken as symmetric, the total number of connected components is", nC, "and the biggest component has:", maxC, "elements."
+
+    print "For the graph where two pages are only considered linked if they link to each other, the total number of connected components is", nC2, "and the biggest component has:", maxC2, "elements."
+
+
+
```
matheuscfernandes: Did this actually work for you? If so, that is awesome! I could not figure this problem out.

HW1/P6/P6.py
```diff
@@ -0,0 +1,81 @@
...
+
+import numpy as np
+import random
+
+def counter(list):
+    
+    d={}
+    for i in list:       
+        d[i] = list.count(i)
+    return d
```
matheuscfernandes: I dont understand what counter(list) is doing. I would suggest adding some more comments to your code so that others can provide better feedback.

HW1/P6/P6.py
```diff
@@ -0,0 +1,81 @@
...
+
+	return joined123_4
+
+
+def build_phrases(text_model, number):
+
+    all_phrases = []
+    for j in xrange(number):
+
+        starting_words = text_model.takeSample(False, 1, int(np.random.uniform(0,100000,1)))
```
matheuscfernandes: I think takeSample already chooses a random sample. I do not think you needed the random number generator. I may be wrong though.

HW1/P5/P5_connected_components.py
isadoranun: @matheuscfernandes it did! But not for the  Wikipedia dataset, only for the Marvel one :(

HW1/P2/P2a.py
sdaulton: ^ Good to know

HW1/P2/P2a.py
sdaulton: I agree

HW1/P2/P2a.py
```diff
@@ -1,3 +1,16 @@
...
+
+partitions = sc.parallelize(tuples, 100).map(lambda xy: (xy, mandelbrot((xy[1]/500.0)-2,(xy[0]/500.0)-2)))
```
sdaulton: With the above changes, you also wouldn't need to call sc.parallelized() here

HW1/P2/P2a.py
```diff
@@ -1,3 +1,16 @@
...
+
+draw_image(partitions)
```
sdaulton: A title and axis labels would be make this histogram a little clearer

HW1/P2/P2b.py
sdaulton: Nice hand written partition function!

HW1/P3/P3.txt
sdaulton: I got the same anagram.  There seem to multiple 11 word anagrams

HW1/P3/P3.py
```diff
@@ -0,0 +1,13 @@
...
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext()
+
+wlist = sc.textFile('./EOWL_words.txt')
+
+wlist2 = wlist.map(lambda line: (''.join(sorted(line)), (line)))
+
+wlist3 = wlist2.groupByKey().mapValues(lambda x: [i for i in x ]).map(lambda y: (y[0],len(y[1]),y[1]))
```
sdaulton: As a good rule of practice, you probably want to avoid groupByKey and use ReduceByKey instead

HW1/P4/P4 bfs.py
```diff
@@ -0,0 +1,50 @@
...
+import pyspark
+import numpy as np
+
+
+def copartitioned(RDD1, RDD2):
+    "check if two RDDs are copartitioned"
+    return RDD1.partitioner == RDD2.partitioner
+
+
+def filterer(i):
```
sdaulton: You have parameter i here, but you don't use it in the function.  Did you mean to use itr_acc?

HW1/P4/P4 bfs.py
```diff
@@ -0,0 +1,50 @@
...
+    def filt(x):
+        return x==itr_acc.value
+    return filt
+
+def get_first_non_empty_cogroup(x2):
```
sdaulton: A more descriptive variable name than x2 would make this more readible.  Also some doc strings on the help functions would be nice

HW1/P4/P4 bfs.py
```diff
@@ -0,0 +1,50 @@
+#I discussed the solution for this problem with Avery Faller.
```
sdaulton: Be careful having a space (' ') in your filename.  This can lead to filename errors doing I/O.  Generally, best practice is to use underscores or something

HW1/P4/P4 bfs.py
sdaulton: ^Oh

HW1/P4/P4.py
```diff
@@ -0,0 +1,35 @@
...
+# findspark.init()
+import pyspark
+import numpy as np
+
+from P4_bfs import *
+
+
+# We lode the data and create the graph
+
+superheroes = sc.textFile('./P4/source.csv')
```
sdaulton: Kind of a bizarre way to open the file using "./P4/source.csv".. I think this means python will only find the file if call from HW1.  Or does this work if you run the script from the command line while in HW1/P4/ ?

HW1/P4/P4.py
```diff
@@ -0,0 +1,35 @@
...
+
+ByArticle = superheroes.map(lambda line: (line.split('","')[1][:-1], line.split('","')[0][1:]))
```
sdaulton: You could make this more efficient by only calling line.split() once, instead of for both elements in the tuple

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,49 @@
...
+    graph.partitionBy(N).cache()
+    
+    total_char_acc = sc.accumulator(1)
+    it_acc = sc.accumulator(0)
+
+    while True:
+        it_acc += 1
+        dist = it_acc.value
+
+        nodes_2 = nodes.filter(lambda x: filterer(x[1])).join(graph).flatMap(lambda x: x[1][1]).distinct().map(lambda x: (x, dist + 1)).partitionBy(N)
```
sdaulton: You really need some comments here.  It is quite hard to determine what this line does

HW1/P4/P4_bfs.py
sdaulton: ^ Not sure I follow your point.  This statement is in the while loop, and you quit here since the loop's condition is always True and will never end on its own

HW1/P5/P5.txt
sdaulton: That being said, the fact that you have multiple paths means you are doing more work than you need to.  BFS for shortest path should stop as soon as you find any path!  It is guaranteed to be the shortest!

HW1/P5/P5_connected_components.py
sdaulton: Nice.

HW1/P6/P6.py
```diff
@@ -0,0 +1,81 @@
+import findspark
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext()
+
+import numpy as np
+import random
+
+def counter(list):
```
sdaulton: See https://docs.python.org/2/library/collections.html#collections.Counter

HW1/P6/P6.py
```diff
@@ -0,0 +1,81 @@
...
+    all_phrases = []
+    for j in xrange(number):
+
+        starting_words = text_model.takeSample(False, 1, int(np.random.uniform(0,100000,1)))
+        phrase = []
+        w1,w2 = starting_words[0][0]
+        phrase.append(w1)
+        phrase.append(w2)
+        #We make a list with the possible third words weighted by their number of ocurrences
+        new_list = sum([([k[0]]*k[1]) for k in starting_words[0][1].items()],[])
```
sdaulton: what exactly does this do?  I'm confused by the use of sum(), and by the empty list second parameter

HW1/P6/P6.txt
```diff
@@ -1 +1,10 @@
...
+["even like those you are. But there is twelve pence for you; and to bruising arms. What never-dying honour hath", 
+"desir'd him to befriend himself. Why, know'st thou this? [Aside] see she's like to be abus'd With silken, sly, insinuating",
+"Find we a prescription to die than to see you, by your favour. My lord, we know no hand in",
+"and car'st for nothing. I'll go to Saint Peter--for the heavens. He shows his love and best hearts of princes",
+"send me to my foe, Once did lay siege to the Council House Early and late, debating to and not",
+"[Kneels] Is this our tott'ring colours clearly up, Last in the flattering table of my passion, woman's gentle heart but",
+"space; For every trifle are they passing cowardly. But, beseech you, That you have broach'd here cannot be thus jaded",
+"nature, cesse! Come on, tune. If you would but stink, and putrefy the air. Cassio and certain Commoners. Hence, home,",
+"him comforts. Sir, The manner and true love's fasting pain. would the caged cloister fly. Religious love put in two",
+"do so. You cannot see her. Why? They say miracles are ceas'd; And therefore shall it find heaven, must find"]
```
sdaulton: Nice

HW1/P2/P2a.py
zezhouliu: Nice tip @matheuscfernandes !

HW1/P2/P2a.py
zezhouliu: Agree with above points by @matheuscfernandes and @sdaulton

HW1/P3/P3.py
zezhouliu: I think in this case groupByKey is fine

HW1/P3/P3.txt
zezhouliu: @sdaulton is correct, there are multiple 11-word ones

HW1/P4/P4_bfs.py
zezhouliu: Yes, commenting would be very helpful!

PR #237, opened 2015-10-05
Hw1

christianjunge:
I have several changes that I wasn't able to push to the origin... After a lot of work on problem 5, I tried to do an add/commit/push, but got a failure because the P5 folder contained the very large zip files for the wikipedia pages.  I removed these from my local directory and tried to add/commit/push again, but after trying to add and commit, I receive a message that my working directory is clean and there is nothing to commit.  And when I try to push, I still receive the large file error, even though it's not present in my local directory or the online repo.  

So a few updates to P4 and all the fruitless work I did on P5 don't appear in this submission.


HW1/P2/P2b.py
```diff
@@ -1,3 +1,85 @@
...
-# Your code here
+    ran = range(0,2000, 1)
+    z = [(a,b) for a in ran for b in ran]
+    # a list of the pairs of x,y values to plot
+    
+    distr_z = sc.parallelize(z, 100)
+    # split z up into 100 partitions
+
+    distr_z_1 = distr_z.repartition(100)
+    # randomly repartition into 100 partitions
```
virgodi: If repartition does it on its own, then it's a great tip ! Thanks ! :-)

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,44 @@
...
+    while acc.value == 0:
+
+        i+=1
+
+        # Use join to look up the connections of the next round of connections
+        namesRdd = graph.join(namesRdd)
+
+        # Use mapValues to take just the list of connections and remove the distance
+        namesRdd = namesRdd.mapValues(lambda (x,y): x).partitionBy(2)
+
```
virgodi: Why a partition by 2? Couldn't you have increased the number of partition to speed up the algorithm?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,44 @@
...
+        # Take the connections and flatMap them into one big pool to consider
+        namesRdd = namesRdd.values().flatMap(lambda x: x)
+
+        # Reduce the connections to distinct entries
+        namesRdd = namesRdd.distinct()
+        
+        # Subtract out the names that are already in the rdd
+        namesRdd = namesRdd.subtract(rdd.keys())
+
```
virgodi: You could have used Subtractbykey instead

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+            lastword = nextWord
+
+            # Look up the next pair and remove the outer brackets
+            thirdwords = sets.lookup(nextPair)[0]
+
+        shakSentences[y] = ' '.join(sents)
+
+    for t in range(len(shakSentences)): print shakSentences[t]    
+    
+
```
virgodi: I really like the clarity of your code, especially when you define functions outside of the map/mapValues !

HW1/P4/P4_bfs.py
christianjunge: I experimented with 4, 8, etc. and they were slower.  I think that the Virtual Machine doesn't have as many CPUs, so partitionBy(2) gave me my best result.

HW1/P2/P2b.py
jeewon-hwang: Wow, different approach. I just used random partition, and did not think about repartitioning after cartesian product. (Also surprised you didn't use cartesian product, but implemented the same thing by different way). But I'm wondering whether there might be some chunks with similar values causing high iterations if we use repartitionBy. Then for those chunks, random shuffle might be needed again.

HW1/P4/P4_bfs.py
jeewon-hwang: Oh, I appreciate I could learn from you those built-in methods in spark. I didn't use either rdd.union nor rdd.substract. If I knew them, I would have tried it. I implemented them in a little bit more complicated way... using set and lists.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,85 @@
...
+import time
+ 
+if __name__ == "__main__":  
+    startTime=time.time()
+    
+    sc = pyspark.SparkContext()
 
-# Your code here
+    ran = range(0,2000, 1)
+    z = [(a,b) for a in ran for b in ran]
```
adisuissa, line 14: The ranges and the cartesian product are done in python, but should be done in spark.

HW1/P3/P3.py
```diff
@@ -0,0 +1,26 @@
...
+    # Group by the sorted name, and put the words with the same sorted name (the anagrams) into a list
+    grouped_pairs_lists = wlist_pairs.groupByKey().mapValues(list)
+    
+    
+    # Take the pairs of sorted names and anagrams, and return the sorted name, the number of anagrams, and the 
+    #     list of anagrams
+    grouped_pairs = grouped_pairs_lists.map(lambda (x,y): (x, len(y), y))
+
+    # Sort the grouped pairs by the number of anagrams, and print the first 
+    print grouped_pairs.sortBy(lambda x: x[1], ascending=False).first()
```
adisuissa, line 26: Nice implementation and good documentation.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,91 @@
...
+        namesRdd = namesRdd.subtract(rdd.keys())
+        
+        if namesRdd.isEmpty(): acc = sc.accumulator(1)   # If nothing was added, change the accumulator value to exit the loop
+        else:
+            # create tuples of the connection name and the current distance of the iteration
+            namesRdd = namesRdd.map(lambda x: (x, i)).partitionBy(2).cache()
+
+            # Union the connection tuples to the existing list, and repartition
+            rdd = rdd.union(namesRdd).partitionBy(2).cache()
+
```
adisuissa, line 46: Instead of using iterations, you should have checked if you reached the end target (the target node was found).
This will also reduce the number of iterations that are performed.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+    # This step also filters out words that contain only capital letters and end with a period
+    no_allCaps = no_nums.filter(lambda word: word.upper() != word)
+
+    # Take the words out of an RDD into a list to create the 3 word combos. 
+    wordlist = no_allCaps.collect()
+
+    # Create 3 lists that will line up to make 1st, 2nd, 3rd word groups.  
+    wl1 = wordlist[:-2]  # slice off the last two entries, which will not be the start of 3 word groups
+    wl2 = wordlist[1:-1] # slice off first and last entries
+    wl3 = wordlist[2:]   # slice off first two entries
```
adisuissa, line 29: The word manipulation should have been done in spark, and not in python.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+        for x in range(18):
+            # Create an empty sample space of potential third words to sample from
+            samplespace = []
+
+            for i in thirdwords: 
+	        for z in range(i[1]):
+	            samplespace.append(i[0]) # Append each word as many times as its weight to the sample space
+        
+            # Select one of the words from the weighted list of potenial words
+            nextWord = random.sample(samplespace, 1)[0]
```
adisuissa, line 80: You can use numpy's random.choice to clarify the code, and avoid lists with redundant entries.

PR #239, opened 2015-10-05
Hw1

amauboussin:
Homework 1 submission


HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+import findspark
+findspark.init()
+import numpy as np
+import pyspark
+
+sc = pyspark.SparkContext(appName='mandelbrot')
+
+
+x_values = sc.parallelize(np.random.permutation(2000), 10)
+y_values = sc.parallelize(np.random.permutation(2000), 10)
```
jdlkem: Nice randomization scheme. I think you could improve the standard deviation by randomizing after creating the coordinates, but this definitely works.

HW1/P3/P3.py
```diff
@@ -0,0 +1,18 @@
...
+import pyspark
+sc = pyspark.SparkContext(appName='anagrams')
+
+with open('EOWL_words.txt', 'r') as rfile:
+	words = rfile.read().split('\n')
+
+lex_order = lambda s: ''.join(sorted(list(s)))
+
+words_rdd = sc.parallelize(words, 10)
+lex_to_word = words_rdd.map(lambda w: (lex_order(w), w))
```
jdlkem: You could probably do without the extra lex_order function above and combine it here. It would save space ;)

HW1/P3/P3.py
```diff
@@ -0,0 +1,18 @@
...
+lex_to_wordlist = lex_to_word.groupByKey()#.mapValues(list)
+result_rdd = lex_to_wordlist.map(lambda (lex, wordlist) : (lex, len(wordlist), wordlist))
+max_entry = result_rdd.max(key = lambda (lex, n, wordlist): n)
```
jdlkem: Nice use of max(). I used takeOrdered().

HW1/P5/P5.txt
```diff
@@ -1 +1,6 @@
-# Your discussion here
+Harvard to Kevin Bacon path:
+[u'Harvard_University', u'Gilmore_Girls', u'Liza_Weil', u'Kevin_Bacon']
+Kevin Bacon to Harvard path:
+[u'Kevin_Bacon', u'Time_(magazine)', u'Harvard_University']
```
jdlkem: I would have liked to see the full set of paths for these.

HW1/P4/P4.txt
```diff
@@ -1 +1,15 @@
-# Your discussion here
+My BFS gives the following output:
```
jdlkem: I agree with all of your answers :+1:

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,68 @@
...
+		visited_this_iteration = adjacencies.leftOuterJoin(to_visit).filter(make_to_be_visited(sc.explored, sc.unvisited))
+		visited_this_iteration.count() # force evaluation so accumulators update
+		flattened = visited_this_iteration.map(lambda (n, (adj, distance)): (n, distance))
+		distances = distances.fullOuterJoin(flattened).map(reduce_distances)
+		to_visit = visited_this_iteration.flatMap(lambda (n, (adj, distance)): adj).distinct().map(lambda a:(a, i))
+
+		#  terminate if the number of unvisited and visited nodes is unchanged: this means the bfs hasn't made progress
+		visited_state = (sc.unvisited.value - last_unvisited, sc.explored.value - last_explored)
+		if visited_state == last_visited_state:
+			break
```
jdlkem: Nice convergence!

HW1/P6/P6.txt
```diff
@@ -1 +1,10 @@
-# Your discussion here
+eyes figure unheedy haste; And therefore we admire, What thou dost ours. Be confident to speak, who was in his
+some unrecuring wound. It was his word. was as willing to grapple as he knows not what? Ay, marry, will
+sir. Allons! we will hear further from me. speak not you? Pedro. What a wicked lie! She false with a
+what! Nature, what things you have a head, To try the fair daughter of Signior Baptista may remember me what
+his attempt. If you had it, then were. But if we use delay, Cold biting winter mars our hop'd-for hay.
+spirit, Of human dealings. If do sweat, they are set, enter MOWBRAY, Duke of Gloucester would speak with you that
+Juliet. Par. Happily met, my lord; it can never win The hearts That spaniel'd me at the next by him
```
jdlkem: lol!

HW1/P6/P6.py
```diff
@@ -0,0 +1,64 @@
...
+	words, counts = zip(*word_counts)
+	total = float(sum(counts))
+	counts = [c / total for c in counts]
+	return np.random.choice(words, size=1, p=counts)[0]
+
+
+def valid_word(w):
+	"True unless a word contains only numbers or only capital letters and periods"
+	all_numbers = w.isdigit()
+	all_upper = all([c.isupper() or c == '.' for c in w])
```
jdlkem: I think the criteria are: (1) no numbers; (2) no words that are all upper case; and (3) no words that have a period at the end and are all upper case otherwise.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+coords = x_values.cartesian(y_values)
+
+def make_mandelbrot_image_and_hist(coords_rdd, hist_fname):
```
sbalanovich: Not that it really matters very much, but what was the point in creating this function in P2.py and then copy-pasting it here? I see you have the import of this function commented out in line 2, I'm just wondering why.

HW1/P3/P3.py
sbalanovich: I don't know if I agree with this, I feel like it's easy with all these Spark functions to plot them all out and then get carried away with building long one-liners that only make sense to you in the moment. I think decisions to split like this make the code more readable.

HW1/P3/P3.py
```diff
@@ -0,0 +1,18 @@
...
+sc = pyspark.SparkContext(appName='anagrams')
+
+with open('EOWL_words.txt', 'r') as rfile:
+	words = rfile.read().split('\n')
+
+lex_order = lambda s: ''.join(sorted(list(s)))
+
+words_rdd = sc.parallelize(words, 10)
+lex_to_word = words_rdd.map(lambda w: (lex_order(w), w))
+lex_to_wordlist = lex_to_word.groupByKey()#.mapValues(list)
```
sbalanovich: Try to avoid using groupByKey() whenever you can (for efficiency). A reduceByKey() may have been the better call here.

HW1/P4/P4.py
```diff
@@ -0,0 +1,42 @@
...
+	hero_book = sc.parallelize(links)
+	create_edges = lambda char_list: tuple(permutations(char_list, 2))
+	edges = hero_book.groupByKey().values().flatMap(create_edges)
+	adjacencies = edges.groupByKey().map(lambda (n, adj): (n, tuple(set(adj))))
+	if cache:
+		adjacencies.cache()
+	return adjacencies
+
+
+if __name__ == '__main__':
```
sbalanovich: What's the purpose of this? Just curious, is it good practice in Python to do this? (It just looks redundant to me)

HW1/P4/P4.py
```diff
@@ -0,0 +1,42 @@
...
+def main():
+	adjacencies = load_adjacencies(cache=False)
+
+	characters = ['CAPTAIN AMERICA', 'MISS THING/MARY', 'ORWELL']
+	for c in characters:
+		print 'BFS starting from %s touches %d nodes' % (c, bfs2(sc, adjacencies, c))
+
+
+def load_adjacencies(fname='source.csv', cache=False):
+	with open('source.csv') as rfile:
```
sbalanovich: Nice pre-processing :)
I'm wondering though, given the nature of this problem set, would it have been at all useful to parallelize this? I guess this is such a fast, simple action that it doesn't really matter, but perhaps loading into a textFile and using a map() instead might have been a better approach, just for the purposes of this assignment

HW1/P3/P3.py
jdlkem: @sbalanovih This is just my personal style preference. I wouldn't have to "dig" through the code to find the lex_order() function, when it is just as reasonable to build a "long one-liner", which tells a story. 

Perhaps the best solution to this would be to name the function as explicitly as it is supposed to be used.

HW1/P3/P3.py
amauboussin: I don't use the whole group of words in the next line, and I don't think reducing by an identity fucntion would speed anything up. It might have been possible to reduce by length and then only fetch the contents of the largest group.

HW1/P4/P4.py
amauboussin: I agree it would have been better to do in spark

HW1/P4/P4.py
amauboussin: it makes it so the main() function will execute if the python file is executed directly while stilling allowing the other functions to be imported, I mostly just put it as habit (http://stackoverflow.com/questions/419163/what-does-if-name-main-do)

HW1/P6/P6.py
amauboussin: I filtered all words where every alpha character is uppercase which is a superset of #3 but could over fitler on mixtures of all upper case letters with other punctuation, my mistake.

HW1/P3/P3.py
amauboussin: I appreciate the discussion, I was aiming for the function's output to be evident for the name so the reader wouldn't have to parse ''.join(sorted(list(s))) when reading the second line or do any "digging" unless they were interested in implementation details.

HW1/P2/P2b.py
amauboussin: There's no good reason, the import wasn't working for some reason and I ended up copying and pasting rather than actually fixing it, definitely bad practice.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,68 @@
...
+		last_explored = sc.explored.value
+
+		visited_this_iteration = adjacencies.leftOuterJoin(to_visit).filter(make_to_be_visited(sc.explored, sc.unvisited))
+		visited_this_iteration.count() # force evaluation so accumulators update
+		flattened = visited_this_iteration.map(lambda (n, (adj, distance)): (n, distance))
+		distances = distances.fullOuterJoin(flattened).map(reduce_distances)
+		to_visit = visited_this_iteration.flatMap(lambda (n, (adj, distance)): adj).distinct().map(lambda a:(a, i))
+
+		#  terminate if the number of unvisited and visited nodes is unchanged: this means the bfs hasn't made progress
+		visited_state = (sc.unvisited.value - last_unvisited, sc.explored.value - last_explored)
```
sbalanovich: This could potentially be simplified if you just reset the accumulators to 0 at the start of every loop iteration and check if they're still 0 after all of the above goes through. Then you don't have to do these subtractions and the code might be a tad cleaner.

HW1/P6/P6.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+n_phrases = 10
+phrase_length = 20
+
+def main():
+	words = load_txt()
+	n_words = words.count()
+
+	#  shift indices to create rdd of three-grams
```
sbalanovich: Clever! :thumbsup:

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,61 @@
+from pyspark import AccumulatorParam
+
+class SetAccumulator(AccumulatorParam):
```
sbalanovich: This seems like a good approach. It's surprising to that this wasn't enough to speed up the code to find all paths and that you had to reduce to an arbitrary path to make this finish in a reasonable time. This is a good optimization though.

PR #240, opened 2015-10-05
HW1

draj36:


HW1/P2/P2.py
```diff
@@ -1,3 +1,4 @@
+from __future__ import division
```
dekatzenel: nit: unused import

HW1/P2/P2a.py
```diff
@@ -1,3 +1,59 @@
...
+import pyspark
+import time
+
+
+sc = pyspark.SparkContext(appName = "Spark1")
+
+# Use default partitioning below
+yaxis = sc.parallelize([ii for ii in range(0,2000)],10)#.map(lambda x: (x,x))
+xaxis = sc.parallelize([ii for ii in range(0,2000)],10)#.map(lambda x: (x,x))
+#xaxis = xaxis.partitionBy(10,lambda x: int(np.floor(x/200)))
```
dekatzenel: Leaving commented-out code in your submission is poor style

HW1/P2/P2.txt
```diff
@@ -1 +1,20 @@
...
+In P2b, I use a different random hashing for the x and y value arrays. This results in the image worker computations
+being uniformly equal in expectation. This is because each pixel is equally likely to be in each partition unlike
+the default case which forms partitions from the 10 200x200 pixel squares (meaning the computation heavy pixels in the
+middle are likely to be in the same partitions). The worker computation is more balanced as seen in P2b_hist.png. The 
+load balancing is also seen in custom_events.png. Note the total time for the custom partitioning is more (3.1 min) 
+than the default case (1.6 min). Since my virtual machine only has 1 core, I don't get the speed ups from the load 
+balanced partitiong and the communication and shuffling costs of partitionBy result in the additional 1.5min. 
+The shuffling cost is high because the x and y value arrays are not copartitioned. If they were copartitioned using
+the same random hashing for the x and y arrays, then there would be less randomness so less load partitioning at the 
+cost of less shuffling.
```
lhoangcse: nice writeup and also detailed analysis (i.e. illustrative graphs) below

HW1/P3/P3.py
```diff
@@ -0,0 +1,49 @@
...
+
+allWords = sc.textFile('A Words.csv').cache()
+allWords = allWords.map(lambda x: (1,x))
+allWords = allWords.groupByKey().mapValues(list)
+bWords = sc.textFile('B Words.csv').cache()
+bWords = bWords.map(lambda x: (1,x))
+bWords = bWords.groupByKey().mapValues(list)
+allWords = allWords.join(bWords)
+allWords = allWords.map(lambda x: (x[0],[item for sublist in x[1] for item in sublist])) # read in data below
+fileNames = ['C Words.csv','D Words.csv','E Words.csv','F Words.csv','G Words.csv','H Words.csv','I Words.csv','J Words.csv','K Words.csv',\
```
lhoangcse: Looks like this is a fine solution, but I think it's slightly over-complicated and loading the files in multiple times seems unnecessary.

HW1/P4/P4.py
```diff
@@ -0,0 +1,29 @@
...
+from P4_bfs import *
+import numpy as np
+import pyspark
+
+sc = pyspark.SparkContext(appName = "Spark1")
+
+marvelData = sc.textFile('marvel.csv').cache()
+
+marvelList = marvelData.map(lambda x: x.split('"'))
+charL = marvelList.map(lambda x: (x[1],x[3])).partitionBy(10)
```
lhoangcse: nit: nice work with the partitionBy, although it'd be safer to define a global constant for the number of partitions to make sure RDDs have the same partitioner.

HW1/P4/P4.txt
```diff
@@ -1 +1,26 @@
...
+Connected Components sizes of the undirected Marvel graph:
+
+[6408, 1, 7, 9, 2, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 1]
+
+Connected Component sizes at each iteration:
+
+CAPTAIN AMERICA: [1, 1907, 6370, 6408]
+MISS THING/MARY: [1, 7]
+ORWELL: [1, 9]
+
```
lhoangcse: Nice answer. I like that you also had the connected components info here as well.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,89 @@
...
+
+	oldValue = 0
+	newValue = 1
+
+	while newValue != oldValue:
+		bfsDistances = bfsDistances.leftOuterJoin(currentSet).cache() # (character,((storedPrevious,storedDist),(setDistPrevious,currentSetDist))), if currentSetDist = none, do nothing 
+		bfsDistances.count() 													#       else if new neighbor ('inf'), replace with count
+		oldValue = nNeighbors.value
+		bfsDistances = bfsDistances.map(updateDist).cache() # update distances (character,(prevNodes,updatedDistance))
+		bfsDistances.count()
```
lhoangcse: you could use foreach() loop to update the accumulator so you don't have to count(). But I also like hitting 2 birds with 1 stone.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,89 @@
...
+		newValue = nNeighbors.value
+		if newValue != oldValue:
+			counts.append(newValue)
+			currentSet = currentSet.rightOuterJoin(charAdj)
+			currentSet = currentSet.filter(lambda x: x[1][0] != None) # for nodes in current set, get (node, ((None,inf),[node.neighbors]))
+			currentSet = currentSet.map(lambda x: (x[0],x[1][1])) # (node,[node.neighbors])
```
lhoangcse: Thanks for the comments indicating RDD structure, very helpful for reading.

HW1/P5/P5.txt
```diff
@@ -1 +1,64 @@
...
+I find connected components by starting BFS from a random collection of 500 nodes. While some
+of these nodes may be in the same connected components, I use the indices of the nodes as the 
+stopping criterion. More specifically, for each node, I store the minimum index of any node that 
+has visited it. Once all the stored values stop changing, then all the connected components touched
+by the 500 nodes have been fully explored. Each connected component is labeled with the same 
+minimum index of the node in the connected component that is also in the original set of 500. 
+I use this technique to remove explored nodes and keep track of the component sizes.
+
+Number of connected components: 6373
+Biggest connected component size: 5709800
```
lhoangcse: First of all, nice work with this problem. I'm not sure why my answer for the biggest component matches yours but not the total number of components which is 559 for me. But I think getting any number at all deserves credit for this part.

HW1/P5/P5.txt
```diff
@@ -1 +1,64 @@
...
+also represented as [2152782, 1185660, 1593422, 2729536]
+
+Connected Components:
+
+I form my nodes using all lines in the page_names file. This results in a lot of nodes with no 
+edges as seen below. 
+
+Case 1. Keep all edges (directed edges are treated as undirected) connected components by size:
+
+I find connected components by starting BFS from a random collection of 500 nodes. While some
```
lhoangcse: Does this mean you have 500 different BFS iterations? Just curious, how long does each of your iteration take and how long for the entire run? I initially tried this and it ended up too slow. Also, I think a problem with this approach (correct me if I'm wrong) is that it finds at most 500 components if each node belongs to a unique component. So if there are more components they will not be found.

HW1/P6/P6.py
```diff
@@ -0,0 +1,73 @@
...
+
+wordTriples = wordTriples.groupByKey().mapValues(list)
+wordTriples = wordTriples.map(lambda x: (x[0],[(ii,sum(np.array(x[1])==ii)) for ii in np.unique(x[1])])).partitionBy(10).cache()
+
+def getNextWord(orderedPair): # calculate random third word based on empirical counts
+	counts = np.array([ii[1] for ii in orderedPair[1]])
+	probs = np.cumsum(counts)/sum(counts)
+	randI = np.random.uniform(0,1,1)
+	ind = sum(probs < randI)
+	return (orderedPair[0],[orderedPair[1][ind][0]])
```
lhoangcse: nice, taking the next word randomly seems to be a good approach.

HW1/P6/P6.txt
```diff
@@ -1 +1,12 @@
...
+[u'Leon.', u'Thine,', u'Claudio;', u'thine', u'say.', u'Pedro.', u'You', u'have', u'crafted', u'fair!', u'You', u'have', u'work', u'in', u'me', u'Upon', u'the', u'error', u'that', u'love'], 
+[u'and', u'beget', u'a', u'thousand:', u'here', u'will', u'be', u'bitter', u'with', u'me.', u'am', u'the', u'King', u'will', u'always', u'bear', u'himself', u'with', u'quarrelling,', u'Some'], 
+[u'door', u'without;', u'let', u'him', u'enter.', u'If', u'he', u'were', u'as', u'his.', u'am', u'your', u"sorrow's", u'nurse,', u'And', u'will', u'not', u'here', u'the', u"leaf's"], 
+[u'painted,', u'and', u'the', u'knaves,', u'you', u'need', u'not', u'fear,', u'provokes', u'me.', u'Nay,', u'nay,', u'Octavia,', u'not', u'only', u'die', u'the', u'death,', u'and', u'welcome!'], 
+[u'court.', u'To-morrow,', u'may', u'it', u'be?', u'You', u'must,', u'if', u'you', u'live,', u'draw', u'your', u'neck', u'out', u'of', u'the', u'buck!', u'Buck,', u'buck,', u'buck!'],
```
lhoangcse: I always get a good laugh at some of the generated text, like this one: "court To-morrow may it be? You must if you live draw your next out of the buck! Buck, buck, buck!"

HW1/P6/P6.py
```diff
@@ -0,0 +1,73 @@
...
+
+sc = pyspark.SparkContext(appName = "Spark1")
+
+allWords = sc.textFile('shakespeare.txt') # I copied the text file to my local directory and deleted the beginning and ending legal non-Shakespeare business
+
+allWords = allWords.map(lambda x: np.array(x.split(' '))).flatMap(lambda x: x[x != '']) # split text by line then remove unnecessary white spaces
+
+allWords = allWords.filter(lambda x: x.isupper() == False) # Filter for all caps and all numbers
+allWords = allWords.filter(lambda x: x.isdigit() == False)
+
```
lhoangcse: I think you're supposed to also remove words which are all uppercase AND ending with a period.

HW1/P2/P2.txt
```diff
@@ -1 +1,20 @@
-# Your discussion here
+
+
+In both P2a and P2b, I use 100 partitions. In P2a the partitioning is the default for both the x and y values. 
+The image is formed from the cartesian product the the x and y value arrays. This results in a very 
+unbalanced load distribution as seen in P2a_hist.png. You can also see the results in default_events.png.
```
dekatzenel: More detailed numerical examples in writing would be nice.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,59 @@
...
+workTime = sum_values_for_partitions(newImage).collect()
+endTime = time.time()
+
+plt.hist(workTime)
+plt.title('Default Worker Time Distribution')
+plt.xlabel('Number of Iterations')
+plt.ylabel('Count')
+plt.show()
+
+print endTime - startTime
```
dekatzenel: Not necessary for the assignment

HW1/P2/P2b.py
```diff
@@ -1,3 +1,42 @@
...
+np.random.seed(100)
+rList = np.random.uniform(0,1,2000)
+rPerm = np.array(sorted(range(len(rList)), key = lambda k:rList[k])) # get random permutation of 1,...,2000
+repNum = np.array([ii for ii in range(0,10)]*200) # get 200 copies of each hash key (1,...,10)
+rDictX = repNum[rPerm] # form hash map using random permutation of hash keys, same for y values below
+np.random.seed(200)
+rList = np.random.uniform(0,1,2000)
+rPerm = np.array(sorted(range(len(rList)), key = lambda k:rList[k]))
+repNum = np.array([ii for ii in range(0,10)]*200)
+rDictY = repNum[rPerm]
```
dekatzenel: When you call an identical chunk of code twice, it makes sense to define a separate function and then call that function

HW1/P2/P2.txt
```diff
@@ -1 +1,20 @@
...
+
+In P2b, I use a different random hashing for the x and y value arrays. This results in the image worker computations
+being uniformly equal in expectation. This is because each pixel is equally likely to be in each partition unlike
+the default case which forms partitions from the 10 200x200 pixel squares (meaning the computation heavy pixels in the
+middle are likely to be in the same partitions). The worker computation is more balanced as seen in P2b_hist.png. The
```
dekatzenel: Well thought out and interesting function. However, just calling partitionBy(10) on each of the axes (with a dummy key-value setup so the function works) actually produces more evenly distributed results. Just food for thought.

HW1/P3/P3.py
dekatzenel: For example, a simpler solution would have used .union() rather than .join() with a dummy key

HW1/P3/P3.py
```diff
@@ -0,0 +1,49 @@
...
+	allWords = allWords.map(lambda x: (x[0],[item for sublist in x[1] for item in sublist]))
+
+allWords = sc.parallelize(allWords.map(lambda x: x[1]).collect()[0])
+print len(allWords.collect())
+allWords = allWords.map(lambda x: (''.join(sorted(x)),x)) # make an RDD where key is word in sorted order
+allWords = allWords.groupByKey().mapValues(list) # group by key to get anagrams
+allWords = allWords.map(lambda x: (x[0],len(x[1]),x[1])) # get desired output
+gramCount = allWords.map(lambda x: x[1])
+maxCount = gramCount.max()
+maxEntry = allWords.filter(lambda x: x[1] == maxCount) # get entry with most anagrams
```
dekatzenel: using .takeOrdered() with a function based on length would be simpler than calculating the max and then filtering to get it. However, I appreciate that you retrieved all answers, rather than just 1.

HW1/P4/P4.py
```diff
@@ -0,0 +1,29 @@
...
+comicL = marvelList.map(lambda x: (x[3],x[1])).partitionBy(10)
+
```
dekatzenel: You could have skipped charL entirely by joining comicL with itself. This would also have simplified subsequent processing steps.

HW1/P5/P5.txt
dekatzenel: @lhoangcse is it possible that you left out nodes with no neighbors?
Agreed that getting a solution is impressive - I couldn't get my code to be fast enough.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,118 @@
...
+			intermediate = intermediate.join(bfsDistances).map(lambda x: (x[0],x[1][1][0])) # (prevNode1, prevPrevNodes1),...
+			#intermediate = intermediate.map(lambda x: [(x[0],ii) for ii in x[1]] if x[1][0] != startName else [(x[0],x[1][0])]) # [(prevNode1, prevPrevNode1), (prevNode1, prevPrevNode2),...]
+			intermediate = intermediate.map(lambda x: [(x[0],ii) for ii in x[1]]) # [(prevNode1, prevPrevNode1), (prevNode1, prevPrevNode2),...]
+			intermediate = intermediate.flatMap(lambda x: x).partitionBy(30).cache() # (prevNode1, prevPrevNode1), (prevNode1, prevPrevNode2),...
+			
+			nextIt = getPaths(intermediate) # (prevNode1,prevPrevNodes1),(prevNode1,prevPrevNodes2)
+			revEndNodes = endNodes.map(lambda x: (x[1],x[0])) # (prevNode1,currentNodes1),(prevNode2,currentNodes1),...
+			appendPath = revEndNodes.join(nextIt) # (prevNode1, (currentNodes1,prevPrevNodes1))
+			appendPath = appendPath.map(lambda x: (x[1][0],[x[0]]+x[1][1])).partitionBy(30).cache() # (currentNodes1,prevNodesUpdated)
+			return appendPath
```
dekatzenel: Nice function

HW1/P6/P6.py
dekatzenel: According to the documentation on isupper, it will "Return true if all cased characters in the string are uppercase and there is at least one cased character, false otherwise." You want to check whether words containing only letters have all uppercase letters, plus the ending with a period. This preprocessing removes things that it shouldn't. See https://piazza.com/class/icqfmi2aryn6yv?cid=226

dekatzenel:
Overall a strong submission. I'm especially impressed with connected components.

PR #242, opened 2015-10-05
HW1 Submission

apetschek:


HW1/P2/P2.py
```diff
@@ -29,3 +30,33 @@ def draw_image(rdd):
...
+range2000 = range(2000)
+def randomList(a):
+    b = []
+    for i in range(len(a)):
+        element = random.choice(a)
+        a.remove(element)
+        b.append(element)
+    return b
+
+randomlist = randomList(range2000)
```
stephenshifang: I believe you can achieve the same results using random.shuffle. It should be faster than using python loops.

apetschek: Ah, you know what: I just realized that this file was an initial commit early on, but i ultimately replaced it with P2b found below (where I randomized partitions instead of the pixel arrangement). I guess i never deleted P2.py from my github? Thought i did. Oh well. 


Thanks for the tip though!

HW1/P2/P2.py
```diff
@@ -1,6 +1,7 @@
 import numpy as np
 import matplotlib.pyplot as plt 
 import matplotlib.cm as cm
+import random
```
apetschek: Dear Ray / cs205 staff--

I just noticed that this "P2.py" file should have been omitted from the final pull request. It was ultimately replaced by the two files below: "P2a.py" and "P2b.py" as were requested in the spec. If possible, please ignore this file as it was a remnant of an early commit that was never removed (sorry for the oversight!). 

Thanks,
Andrew

HW1/P2/P2.py
stephenshifang: I see. I'll skip this then.

HW1/P3/P3.py
```diff
@@ -0,0 +1,25 @@
...
+#now combine letters back together using join function. 
+SortedLetterSequence = wordcharacters.map(lambda x: "".join(x))
+
+#make two columns of our alphabetized words and .groupByKey to get anagrams
+keyvaluerdd = words.map(lambda x: ("".join(np.sort(list(x))),x))
+temp = keyvaluerdd.groupByKey().mapValues(list)
+Anagrams = temp.map(lambda x: (x[0],(len(x[1]),x[1])))
+
+#find maximum where we know max=11, so use filter.
+maximum = Anagrams.filter(lambda x: x[1][0] == 11)
```
thereisnoquarter: There should be a way to find the maximum (sorting it, for example) without a-priori knowing that it's going to have 11 occurrences. I think the 11 was just a hint to see if you were right or not, not to look up the max.

thereisnoquarter: There should be a way to find the maximum (sorting it, for example) without a-priori knowing that it's going to have 11 occurrences. I think the 11 was just a hint to see if you were right or not, not to look up the max.

HW1/P4/P4.py
```diff
@@ -0,0 +1,30 @@
...
+Superheroes = data.map(lambda word: word.split('",')[0]).map(lambda word: word.strip('"').encode("utf-8"))
+Comics= data.map(lambda word: word.split('",')[1]).map(lambda word: word.strip('"').encode("utf-8"))
+
+RDD1= Comics.zip(Superheroes)
+RDD2 = RDD1.join(RDD1) #K,(V,W) = Comic, (Superhero,Superhero)
+
+#create edge list and ensure no one is linked to themselves (this assumes
+#every character has at least one neighbor otherwise if they are solo then
+#they will never be found as they are omitted from this graph)...
+Graph = RDD2.filter(lambda x: x[1][0]!=x[1][1]).map(lambda x: (x[1][0],x[1][1])).distinct().partitionBy(8).cache()
```
thereisnoquarter: Looks good - but why partitionBy 8?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,32 @@
...
+    final_rdd1 = node1
+    i = 0
+    
+    #initialize an accumulator and set to 0. Create while loop to run
+    #while accumulator matches value of the i iterator 
+    accumulator = sc.accumulator(0)
+    i = 0
+    while accumulator.value == i:        
+        #find neighbors of input node        
+        node1 = Graph.join(node1).distinct().values().partitionBy(8)
```
thereisnoquarter: Clever solution with the join, I didn't think of this and mine came out much more convoluted as a result.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,32 @@
...
+    #initialize an accumulator and set to 0. Create while loop to run
+    #while accumulator matches value of the i iterator 
+    accumulator = sc.accumulator(0)
+    i = 0
+    while accumulator.value == i:        
+        #find neighbors of input node        
+        node1 = Graph.join(node1).distinct().values().partitionBy(8) 
+        
+        #new node only includes neighbors not already searched/in Final RDD        
+        node1 = node1.mapValues(lambda x: x+1).subtractByKey(final_rdd1)
```
thereisnoquarter: Instead of comparing to the original RDD each time, one approach (and the one that I took) would be to color each node either WHITE, BLACK, or GRAY, standing for - "already been searched, no new information", "hasn't been searched yet", and "going to step from this node on this iteration". You increment the accumulator every time you change a node to GRAY. Then you stop when there are no more GRAY nodes.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,73 @@
...
+def copartitioned(RDD1, RDD2):
+    "check if two RDDs are copartitioned"
+    return RDD1.partitioner == RDD2.partitioner
+    
+sequence = []
+  
+def bfs(SOURCE,Graph,sc,DEST=False):
+    #set source as source node and distance to 0    
+    node1 = [(SOURCE,0)]
+    node1 = sc.parallelize(node1).partitionBy(128)
```
thereisnoquarter: Why 128? I seem to remember 256 being the way to go? Probably doesn't matter all that much, though.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,73 @@
...
+        RDD3 = RDD2.filter(lambda (K,V): V[0] == DEST)
+        RDD4 = RDD3.keys().collect()[0]
+        RDD5 = final_rdd1.filter(lambda (K,V): V == (distance - 2)).partitionBy(128)
+        RDD6 = Graph.join(RDD5)
+        RDD7 = RDD6.filter(lambda (K,V): V[0] == RDD4)
+        RDD8 = RDD7.keys().collect()[0]
+        sequence.append(RDD4)
+        sequence.append(RDD8)
+        sequence.append(DEST)
+
```
thereisnoquarter: This only works if you already know how far away the nodes are - i.e., you ran the search, got the distances, added this code, and ran it again. Another approach would be to start every node off with a "path list" of []. When you step to a node in the BFS, you append your parent to your parent's path list, and that becomes your path list. For example, for a graph v1 -> v2 -> v3 with v1 as the source, the path lists would be:

v1 = []
v2 = [v1]
v3 = [v2, v1]

And you see that this is exactly the result you want.

HW1/P6/P6.py
```diff
@@ -0,0 +1,65 @@
...
+
+#initialize RDD of each triplet of words. Could have used zipwith index but
+#list/array manipulation in python likley needed before calling zipwithindex
+#so i determined the below was just as simple/used python just as much
+
+trips = []
+for i in range(len(string)-2):
+    trips.append(((string[i],string[i+1]),(string[i+2])))
+trips = sc.parallelize(trips)
+
```
thereisnoquarter: I think the zipWithIndex approach would scale much better to larger datasets. For looping over a big dataset serially is O(n).

HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
...
+
+When first plotting the image, it can be seen that there are a few concentrated areas of high iteration count (bright spots) and many areas of low/no iteration counts (black areas). Given our initial partitioning strategy, this means that most workers would have been assigned the ‘black’ areas and had fairly little computation to carry out, while a few (unlucky) workers were assigned the bright areas and were stuck doing most of the computation. My first histogram reflects this where most workers had low iteration counts and a few workers had high iteration counts.
+
+To fix this, i used the .repartition() method to randomly repartition my RDD into 100 random partitions. This effectively distributes a random 1/100 of the pixels to each worker to more evenly distribute the computation across our 100 workers. My 2nd histogram reflects this change where we now have a more balanced load of work across workers. Repartitioning an RDD does involve a cost of time in it of itself, but one would repartition to hopefully save more time moving forward due to this transformation.
```
stephenshifang: Interesting. What's the default partition function for repartition()? It looks like it worked very well.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,53 @@
...
+    plt.imshow(im, cmap=cm.gray)
+    plt.show()
+    
+#Perform same steps as in P2a, however now we call repartition to randomly
+#repartition RDD into 100 partitions. note, i was forced to call .partitionBy
+#first for some reasons or else repartition would not seem to work.
+    
+pixel = range(2000)
+pix1D=sc.parallelize(pixel,10)
+pixels = pix1D.cartesian(pix1D).partitionBy(100).repartition(100)
```
stephenshifang: Interesting finding. I used a custom partition function in my implementation and didn't run into this issue.

HW1/P3/P3.py
stephenshifang: I agree with thereisnoquarter.

HW1/P4/P4.txt
```diff
@@ -1 +1,27 @@
...
+ 
+
+ORWELL - 9
+MISS THING/MARY - 7
+CAPTAIN AMERICA - 6408
+
+If a character does not have a defined distance, then that means that character is disconnected from my source node. With respect to this problem, that would mean that that character does not share a comic with the source node nor does any of his/her neighbors. there is no connection between that character and my source node. 
+
+The diameter of the graph denotes the maximum number of iterations in which my search must be executed for a single connected component. If my understanding of diameter is correct, it is the diameter of a graph of connected components. I could have another separate connected component within the graph that has its own diameter. If this is not the case, then i suppose the diameter could represent the theoretical maximum number of steps needed but i may not be able to run all of those steps starting from some arbitrary source node if i have separate ‘islands’ of connected components.
```
stephenshifang: Great writeup!

HW1/P2/P2.py
zezhouliu: Thanks for the note

HW1/P4/P4_bfs.py
zezhouliu: This sounds reasonable to me!

PR #246, opened 2015-10-06
Hw1

fpeng01:


HW1/P4/P4.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def extractEdges(mytuple):
+	'''create a one-to-one map (an edge) for each pair of characters 
+	   corresponding to an issue'''
+	graph_edges = []
+	char_list = mytuple[1]
+	for i in range(len(char_list)):
+		for j in range(len(char_list)):
+			graph_edges.append( (char_list[i],char_list[j]) )
```
araftery: Note that this will create self-edges (when `i == j`), which aren't necessary.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,16 @@
...
+def bfs(graph, starting_char):
+    char_list = [starting_char]
+    BFS_list  = []
+    
+    while len(char_list)!=0:
+        neighbors = graph.filter(lambda tpl: tpl[0] in char_list).flatMap(lambda tpl: tpl[1]).collect()
+        BFS_list  = BFS_list + char_list
+        char_list = []
+        char_list = char_list + list(set(neighbors)-set(BFS_list))                                     
+        print len(list(set(neighbors)-set(BFS_list)))
```
araftery: Your implementation here doesn't take much advantage of parallelism. You should be able to do each round in the loop with one map and one reduce. One way to do this, is to make `graph` an [adjacency list](https://en.wikipedia.org/wiki/Adjacency_list) that also stores distance, and overwrite it each time, updating the distances of the nodes touched in that round. You can use an accumulator to figure out when to stop (this is what you'd check in the `while` condition), and the output would be an adjacency list of all characters and their distance from the source.

This takes better advantage of parallelism and reduces costs imposed by `collect()`, which is a very expensive operation.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,126 @@
...
+        graph = graph.flatMap(lambda y: (set_parent_and_dist(y, dist_traversed)) )
+        #after we update the distances of children, we "step" further
+        dist_traversed += 1
+        graph = graph.reduceByKey(lambda a,b: (combine_dist_parent_with_children(a,b)))
+        
+        #check if end_char has been touched
+        #if so, bfs has been done
+        check_dist = graph.lookup(end_char)[0][0]
+        if check_dist < 10**6:
+            return (graph, dist_traversed)
```
araftery: You can use an accumulator to help figure out when the loop is done, and this would avoid doing a lookup.

HW1/P6/P6.py
```diff
@@ -0,0 +1,123 @@
...
+        contain only letters which are capitalized,
+        or contain only letters which are capitalized and end with a period.'''
+    if word.isdigit():
+        return False
+    if word.isupper():
+        return False
+    if word[-1]=='.' and word[:-1].isupper():
+        return False
+    return True
+data_check = data_split.filter(word_can_be_accepted)
```
araftery: This will reject words, for example, that contain all capital letters and end with an exclamation point, which should be accepted.

```
In [1]: "TEST!".isupper()
Out[1]: True
```

HW1/P4/P4_bfs.py
zelong430: I agree with "araftery". The BFS you implemented here is very close to the original version. There is a specifically designed parallelized BFS algorithm which you could found online. The basic idea is mentioned in the homework document, update the distance each loop.

HW1/P6/P6.py
zelong430: if word[-1]=='.' and word[:-1].isupper():

also has potential problem

as .isupper() will return True if all letter are capitalized 

it does't care digits or periods.

HW1/P3/P3.py
```diff
@@ -0,0 +1,42 @@
...
+    inp = my_string
+    chars = list(my_string)
+    sorted_chars = sorted(chars)
+    sorted_string = ''.join(sorted_chars)
+    tpl = (sorted_string, [inp])
+    return tpl
+sorted_list = wlist.map(alphabetical)
+#sorted_list.collect()
+
+reduced_list = sorted_list.reduceByKey(lambda a,b: a+b)
```
zezhouliu: Good!

HW1/P4/P4_bfs.py
zezhouliu: Good catch @araftery and @zelong430

PR #248, opened 2015-10-06
HW1 - Anna Whitney

annawhitney:


HW1/P2/P2.py
```diff
@@ -29,3 +31,26 @@ def draw_image(rdd):
...
+    # initialize spark context
+    sc = pyspark.SparkContext("local[4]", "Spark1")
+
+    # create RDD of pixel location tuples
+    i_vals = sc.parallelize(range(2000), 10)
+    ij_vals = i_vals.cartesian(i_vals)
+
+    # associate mandelbrot values with each pixel
+    ijxy_vals = ij_vals.zip(ij_vals).mapValues(lambda (i,j): (j/500 - 2, i/500 - 2))
+    mandel_vals = ijxy_vals.mapValues(lambda xy: mandelbrot(*xy))
```
samighoche: nice maneuvering to use mapValues instead of the more expensive map

HW1/P2/P2.txt
```diff
@@ -1 +1,4 @@
-# Your discussion here
+Partitioning using partitionBy(100) on (i, j) keys immediately before the Mandelbrot calculation results in a very spatially uniform distribution of pixel locations in each partition (see P2a_partitions.png), and thus a very uniform number of interation steps across the partitions (see P2a_hist.png). The partitionBy incurs about 12 seconds of shuffling overhead on my system.
```
samighoche: Do you mean P2b_partitions.png and P2b_hist.png? :)

HW1/P2/P2b.py
```diff
@@ -1,3 +1,19 @@
...
+i_vals = sc.parallelize(range(2000))
+ij_vals = i_vals.cartesian(i_vals)
+
+# associate mandelbrot values with each pixel
+ijxy_vals = ij_vals.zip(ij_vals).mapValues(lambda (i,j): (j/500 - 2, i/500 - 2))
+mandel_vals = ijxy_vals.partitionBy(100).mapValues(lambda xy: mandelbrot(*xy))
+#draw_image(mandel_vals)
+#draw_partitions(mandel_vals)
+
+plt.hist(sum_values_for_partitions(mandel_vals).collect(), bins=200, range=(0,2.5e7))
```
samighoche: I'm not sure why you're explicitly setting the range for the histogram, it would be clearer if you let it automatically choose the range, and see the distribution of the iterations per partition much more closely. 

partitionBy(100) will randomize the xy values that each partition performs the computation so I agree with your solution. I did it slightly differently and it seemed to give a slightly better distribution (narrower, more centered around the mean), when I just partitioned by taking a random number between 0 and 99 and assigning xy to that partition.

HW1/P2/P2.txt
```diff
@@ -1 +1,4 @@
...
+I would expect it to significantly decrease the time spent in the Mandelbrot calculation, because the computation time is much more evenly balanced between partitions, but for some reason the computation phase after the partitionBy takes ~50 seconds, while the default partitioned version takes ~30 seconds for the same computation. I don't know why this version is taking longer despite its demonstrably better load balancing.
+
+In contrast, the default partitioning scheme results in partitions that each cover a contiguous square (see P2a_partitions.png), which results in poor load balancing in which most partitions perform very few iterations but a few perform a very large number of iterations (see P2a_hist.png), because areas near each other are likely to have similar iteration counts.
```
samighoche: Great job with the partitions.png files, they really visualize the point you're trying to make very nicely!

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+
+# associate sorted character list & counter with each word
+slist = wlist.map(lambda w: (''.join(sorted(w)), (1, [w])))
+
+# combine words with same sorted character list
+anagrams = slist.reduceByKey(lambda (n1, wlst1), (n2, wlst2): (n1 + n2, wlst1 +wlst2))
+
+# flatten (since reduceByKey wouldn't play nice with multiple values)
+anagrams_flat = anagrams.map(lambda (a, (b,c)): (a,b,c))
+
```
samighoche: Great implementation, very efficient and concise!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,51 @@
...
+            return ns
+        ttrdd = ttns.flatMap(lambda (k,(d,ns)): count_get_neighbors(ns))
+        to_touch = set(ttrdd.collect())
+
+        # set new distances
+        touched = touched.union(ttns.mapValues(lambda v: dist_so_far)).cache()
+        graph_dists = graph_dists.leftOuterJoin(ttns).mapValues(lambda (v,w): (dist_so_far, w[1]) if w else v).cache()
+
+        dist_so_far += 1
+
```
samighoche: nice implementation! And good use of caching.

I don't really like the ttrdd.collect() because I'm not sure how scalable it is to be dealing with python sets instead of rdds when you have a really huge graph (but I don't know how you would do it without it, maybe something creative with joins? I had an edge tuple representation for my graph instead)
I also found it hard to follow some of your variable names (ttns, ttrdd).

HW1/P4/P4.txt
```diff
@@ -1 +1,25 @@
...
+If a character does not have a defined distance during SS-BFS, that indicates
+that the character is not on a connected component of the graph with the source
+character. SS-BFS traverses the graph along edges only, so any nodes that are
+not connected to the origin will never be reached.
+
+I chose to represent the graph as an adjacency list, i.e., an RDD with elements
+of the form (character, [list of neighbors]). During SS-BFS, I also added a
+distance component to the values associated with each character. An adjacency
+list allows easy identification of outgoing links, and since this graph is
+undirected, working with outgoing vs incoming links is equivalent. Because the
```
samighoche: I agree with everything here:)

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,65 @@
...
+        # update set of neighbors to touch and check if we're at destination
+        def neighbors_and_check_dest(k, ns):
+            if k == destination:
+                found.add(1)
+            return [(n,k) for n in ns]
+        ttrdd = ttns.flatMap(lambda x: neighbors_and_check_dest(*x)).subtractByKey(touched)
+        to_touch = set(ttrdd.keys().collect())
+
+        # set previous nodes
+        touched = touched.union(ttrdd)
```
samighoche: you should probably be caching that to avoid recomputing at every iteration

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,65 @@
...
+    # trace destination path backwards
+    path = [destination]
+    current = destination
+    while current != origin:
+        lst = touched.lookup(current)
+        if lst:
+            current = lst[0]
+        else:
+            break
+        path.append(current)
```
samighoche: You're returning only one path but I believe the question was to return all paths.
The way I achieved that was by having key value pairs of (node, path_to_node_so_far), and once I found the goal, I would return all the paths to the goal.

HW1/P6/P6.py
```diff
@@ -0,0 +1,60 @@
...
+    consec_counts = consec_words.reduceByKey(add)
+
+    # and group by first two words of triplet
+    markov_table = consec_counts.map(lambda (((w1,w2),w3),c): ((w1,w2),(w3,c))).groupByKey().cache()
+
+    # check that we've done this correctly
+    #print list(markov_table.map(lambda x: x).lookup(('Now','is'))[0])
+
+    # sample from Markov table
+    frac = 50/markov_table.count()
```
samighoche: I don't really understand where the 50 is coming from but I'm sure there's a reason

HW1/P6/P6.py
```diff
@@ -0,0 +1,60 @@
...
+    sample = markov_table.sample(False, frac).keys()
+    sentences = sample.zip(sample.map(lambda ws: list(ws)))
+
+    # generate sentences of length 20
```
samighoche: I have to say I don't fully understand how your code is working but the output looks correct to me so well done!

HW1/P4/P4.txt
```diff
@@ -1 +1,25 @@
...
+graph is not highly connected (there are many characters that are only linked
+to a small number of others), an adjacency matrix would be a very sparse matrix
+and so waste a lot of space, so an adjacency list was preferable in this case.
+
+Cap: touched 6408 nodes
+Miss Thing: touched 7 nodes
+Orwell: touched 9 nodes
```
georgelok: I agree with these results.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+wlist = sc.textFile('EOWL_words.txt')
+
+# associate sorted character list & counter with each word
+slist = wlist.map(lambda w: (''.join(sorted(w)), (1, [w])))
+
+# combine words with same sorted character list
+anagrams = slist.reduceByKey(lambda (n1, wlst1), (n2, wlst2): (n1 + n2, wlst1 +wlst2))
+
+# flatten (since reduceByKey wouldn't play nice with multiple values)
+anagrams_flat = anagrams.map(lambda (a, (b,c)): (a,b,c))
```
georgelok: I think this has to do with the representation of key value pairs in an RDD.

With a 3-tuple, you don't have a clear key-value relationship, while it's clearly evident with a 2-tuple.

PR #249, opened 2015-10-06
Working on HW1

neelspatel:


HW1/P3/P3.py
```diff
@@ -0,0 +1,32 @@
...
+#gets all words with the same sorted alphabetical key
+anagrams = words.groupByKey()
+
+#adds anagram count information
+anagrams_and_counts = anagrams.map(lambda (x, anagrams_list): (x, len(anagrams_list), list(anagrams_list)))
+
+#gets the word with the most anagrams (which is why we sort by -1 * count, to get the largest count)
+most_anagrams = anagrams_and_counts.takeOrdered(1, key=lambda (x, anagrams_count, anagrams_list): -1 * anagrams_count)[0]
+print most_anagrams
+
```
TuanTrain: I had a similar method of implementing this. It is a very good parallelizable problem!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,61 @@
...
+
+	#filters down to the list of nodes which we've touched
+	touched = adjacency_list.filter(lambda (name, (neighbors, distance, visited_flag)): distance < 10000)
+	print touched.count()
+
+	#finds the maximum distance
+	counts = touched.collect()
+	print max(counts, key=lambda (name, (neighbors, distance, visited_flag)): distance)
+
+	return adjacency_list
```
TuanTrain: This is a good implementation of bfs. Your closest_visited_node function is sleek.

HW1/P6/P6.py
```diff
@@ -0,0 +1,87 @@
...
+
+import numpy as np
+
+#words_weights is a list of (word, weight) tuples
+def sample_randomly(words_weights):	
+	total_word_count = float(sum([x[1] for x in words_weights]))
+	probabilities = [x[1]/total_word_count for x in words_weights]
+
+	#from http://stackoverflow.com/questions/6432499/how-to-do-weighted-random-sample-of-categories-in-python
+	word_index = np.array(probabilities).cumsum().searchsorted(np.random.rand())
```
TuanTrain: Another way to do random samples in this case is to repeat values that have a higher count (eg (is,3) -> [is,is,is]), then sample out of the larger list. Since these values tend to be rather small in this case, I do not think it is a big space issue, but would become one if the weights are much larger than the usual 1.

HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
...
+
+For the new partitioning strategy, I assigned pixels to partitions "randomly" by randomizing the lists which went into the Cartesian to create pixel values. This leads to a more even distribution of work across partitions.
+
+The results of this new partitioning were that the computation was faster, and the distribution of times was much more even (resembling a normal distribution rather than the lack-of-distribution from the first part.)
+
+The cost of a random partitioning is that we have a larger shuffle cost at the beginning of the computation. There are also tradeoffs in applications where pixels depend on neighboring pixels for values (for example, searching local neighborhoods in a graph). If we partition randomly, then we get the benefit of costly pixels being evenly distributed, at the price of not being able to minimize shuffles by exploiting local substructures.
```
drishi: Good stuff, agree with all the things written here.

HW1/P3/P3.py
drishi: Yes same here, well done seems very concise and readable.

HW1/P3/P3.txt
```diff
@@ -1 +1 @@
-# Your discussion here
+(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])
```
drishi: Got the same thing.

HW1/P4/P4.txt
```diff
@@ -1 +1,11 @@
...
+MISS THING/MARY: touched 7 nodes
+ORWELL: touched 9 nodes
+
+Graph diameter:
+If the diameter of the graph is 10, we should never need to take more than 10 steps outwards from a source node to find all connected components to that source, as the largest distance between any nodes in the graph is 10.
+
+Undefined distances:
+If a character does not have a defined distance in our search, it means that the character is not accessible from our source node. This is possible because the graph is not completely connected.
+
+
```
drishi: Got the same answers as you. Did not realize that the graph was completely connected though so undefined distances are impossible, that's interesting

HW1/P5/P5.txt
```diff
@@ -1 +1,13 @@
...
+	[u'Harvard_University', u'Jack_Lemmon', u'Rod_Steiger', u'Kevin_Bacon']
+	[u'Harvard_University', u'Cornell_University', u'Edmund_Bacon', u'Kevin_Bacon']
+
+From Kevin Bacon to Harvard University: each path takes 2 steps
+
+	[u'Kevin_Bacon', u'Time_(magazine)', u'Harvard_University']
+	[u'Kevin_Bacon', u'Marisa_Silver', u'Harvard_University']
+
+
+
```
drishi: Hmm..I got different results here. I had 120 paths between Harvard -> Kevin Bacon and my shortest path from Kevin Bacon -> Harvard was 5 steps. I may be wrong, would be interesting to compare..

HW1/P6/P6.py
```diff
@@ -0,0 +1,87 @@
...
+
+print build_sentence(tuples_counts_list)
+print build_sentence(tuples_counts_list)
+print build_sentence(tuples_counts_list)
+print build_sentence(tuples_counts_list, word1="Now", word2="is")
+print build_sentence(tuples_counts_list, word1="Now", word2="is")
+print build_sentence(tuples_counts_list, word1="Now", word2="is")
+print build_sentence(tuples_counts_list, word1="Now", word2="is")
+print build_sentence(tuples_counts_list, word1="Now", word2="is")
+print build_sentence(tuples_counts_list, word1="We", word2="will")
```
drishi: Nice job on this bit! I think the "print statements" could perhaps be done a little more easily in a loop? I made a separation function for that bit and while its actually longer than all your prints (haha) it might make it easier in case we needed like 100 phrases.

HW1/P6/P6.txt
```diff
@@ -1 +1,10 @@
...
+England's coat one half which is strange; Nay, it is worse for that, for their lawful suit! Alas, why is
+to sue, but to read philosophy, And while the consults his Council Draw near, ye people. List to your master,
+best Worthy. Enter NATHANIEL, for When in the King's hand; But, thus much, they are given to your royal father
+Now is my conscience whispers in your ear, Which none but shall follow in my terms of love resembleth The
+Now is the bell. Bell rings. Enter Lady Macbeth with a and attendants Where is he coming, In thunder and
+Now is the matter, ho? Who is't can say no more. Pursue him to serve him so, That in your
+Now is your ring, which, in weight to re-answer, his pettiness would bow to th' top of rage is in
+Now is the man Must stead us all, Never like thee. Bright star of England. You are too bold for
+We will ourselves provide. Most holy and so evident, That it but by help of hatchet. Why dost thou come?
+You are jealous now That bear the shapes of men, That had not been desirous of their mistress. My lord,
```
drishi: Good job on the pset!

HW1/P2/P2b.py
```diff
@@ -1,3 +1,38 @@
...
+draw_image(mandelbrot_results)
+
+#compute "effort" on each partition
+timing_results = sum_values_for_partitions(mandelbrot_results)
+
+plt.hist(timing_results.collect())
+plt.show()
+plt.savefig("P2b_hist.png")
+
+print "Done"
```
zezhouliu: Your P2b histogram doesn't look like a well-balanced distribution.  It looks like it's just as likely to have 1.7m iterations as 2.05m iterations as 2.3m iterations

HW1/P3/EOWL-v1.1.2/CSV Format/J Words.csv
```diff
@@ -0,0 +1,1286 @@
+jab
```
zezhouliu: Data files clutter the pull request and it might be cleaner if you don't commit them

HW1/P4/P4.py
```diff
@@ -0,0 +1,50 @@
...
+import csv
+
+from P4_bfs import *
+
+sc = pyspark.SparkContext(appName="P4")
+sc.setLogLevel('ERROR')
+
+#load in the list of characters and appearances in comic books
+with open("source.csv", "rb") as inputfile:
+	reader = csv.reader(inputfile, quotechar='"')
```
zezhouliu: I think you can do most of your processing in spark instead of in the python driver

HW1/P4/P4.py
```diff
@@ -0,0 +1,50 @@
...
+	character_book = sc.parallelize(character_book)
+
+#reverse the list as an association from book to character
+book_character = character_book.map(lambda (x,y): (y,x))
+
+#group by appearance in each comic book
+by_book = book_character.groupByKey()
+
+#for each comic book, create an edge from each character to each other character
+def characterToOtherCharacters(current_book):
```
zezhouliu: Same as above, do the work in spark rather than in the python driver

PR #251, opened 2015-10-06
completed all parts of HW1

georgeqi94:


HW1/P2/P2a.py
```diff
@@ -1,3 +1,30 @@
...
+
+# crate rdd
+rdd = sc.parallelize(xrange(2000), 10)
+rdd = rdd.cartesian(rdd)
+
+# use formula given by pset
+rdd_map = rdd.map(lambda (i,j): ((i,j), (j/500.0-2, i/500.0-2)))
+
+# apply mandelbrot
+result = rdd_map.mapValues(lambda (x,y): mandelbrot(x,y))
```
georgelok: instead of two map steps, you could have applied Mandelblot in one step with rdd.map(lambda (i,j): (mandelbrot(j/500.0-2, i/500.0-2) ))

HW1/P2/P2a.py
```diff
@@ -1,3 +1,30 @@
...
-# Your code here
+import findspark
+findspark.init('/Users/george/Documents/spark-1.5.0')
+import pyspark
+import matplotlib.pyplot as plt
+ 
+sc = pyspark.SparkContext()
+
+# crate rdd
+rdd = sc.parallelize(xrange(2000), 10)
```
georgelok: Problem spec asks for 100 partitions.

HW1/P2/P2.txt
```diff
@@ -1 +1,7 @@
...
+
+In P2b.py, I used a partitioning strategy that randomized the distribution of the nodes. This took about 49 seconds to run, which is slightly (but not significantly) faster than P2a.py. From the histogram, we can see that this yielded a much more normal distribution of iterations run by each of the partitions, making the standard deviation lower since each of the workers is doing (relatively) similar amounts of work. Therefore, the work is much more evenly balanced among the workers. 
+
+However, looking at the costs, the second method had more than 50% of the time occupied by shuffling, which is definitely concerning. As a result, the second method is only faster if the bottleneck caused by shuffling is outweighted by the increased efficiency and benefits of having the work more balanced.
```
georgelok: Good observation!

HW1/P2/P2b.py
```diff
@@ -1,3 +1,34 @@
...
+# apply mandelbrot
+result = rdd_map.mapValues(lambda (x,y): mandelbrot(x,y))
+
+draw_image(result)
+
+# calculate partition work
+answer = sum_values_for_partitions(result).collect()
+
+# plot histogram
+plt.hist(answer, bins=20)
```
georgelok: Something doesn't seem quite right with you graph.  The number of iterations should be much larger (remember that we're dealing with a 2000 by 2000 grid) and the numbers for the partitions also seem off.

HW1/P6/P6.py
```diff
@@ -0,0 +1,63 @@
+import findspark
+findspark.init('/Users/george/Documents/spark-1.5.0')
+import pyspark
+import random
+
+sc = pyspark.SparkContext()
+
+# filter out upper case and all digits
+def filter_word(word):
+    return not word.isdigit() and not word.isupper()
```
georgelok: Your filter misses the cases of "UPPERCASE."

HW1/P6/P6.py
```diff
@@ -0,0 +1,63 @@
...
+
+# merge lists and produce format wanted by pset
+reduced = consec_3.reduceByKey(lambda x,y: x+y)
+answer = reduced.map(lambda x: (x[0], [(ele, x[1].count(ele)) for ele in list(set(x[1]))]))
+
+# take 10 random samples
+starts = answer.takeSample(False, 10)
+
+# print phrases
+for start in starts:
```
georgelok: There is a way to parallelize the creation of phrases instead of creating each phrase one by one.  Of course, for this scale, it doesn't bring much benefit, but something to think about.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,32 @@
...
+    distance = 0
+    
+    # terminate if found nothing new
+    while len(frontier):
+    	# update values
+        distance += 1
+        prev_vals = set(list(visited.value))
+
+        # iterate through current level and update visited 
+        rdd.filter(lambda (x,y): x in frontier)
```
georgelok: Nice use of a set accumulator.  I'm curious how well it performs, because I found set accumulators to be for some reason slow.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,104 @@
...
+        return union
+
+# update parent pointers so we can backtrack
+def setParents(child, parent):
+    pointer = {}
+    for c in child:
+        pointer[c] = parent
+    return pointer
+
+def bfs(sc, graph, source, dest):
```
georgelok: I think on Piazza, we were told this needed to return all possible paths, not just one single path.

HW1/P2/P2a.py
georgeqi94: I believe doing it with 10 is correct because once you perform cartesian in the next line, it will create 100 partitions (I printed it to test this). This also seems to be consistent with most of the other submissions.

HW1/P6/P6.py
georgeqi94: I think that word.isupper() will also filter out "UPPERCASE." 

For example: "HELLO.".isupper() returns True.

HW1/P2/P2a.py
nchainan: i did the same thing, good

HW1/P2/P2a.py
nchainan: ^this is correct, 10 partitions is fine

PR #252, opened 2015-10-06
HW1 Submission

drishi:
Devvret Rishi's Submission for HW1


HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
 # Your discussion here
+
+
+We note that due to the nature of the mandelbrot function, computation is
+focused on certain parts (which we can tell from color intensity of the image it produces)
+much more than others. Thus, when we allocate the partitions sequentially, we see a disproportionate
+amount of work applied across them as a majority of the function evaluation work is done by a minority of partitions. However, in the next part we choose to select them randomly, and the work becomes much more evenly spread giving us a bell-shaped curve when we look at the total compute the partitions are requiring, alibiet this result comes with an overhead shuffle cost that does not seem to be too bad in this particular secenario.
+
+If we were really looking for a more optimal partiiton, perhaps we could iterate through the pixels and then use the mandelbrot function (or an approximation of it) to determine the amount of relative work we would need to do for that fixed point, and then partition accordingly. However, this would be a lot more work from an implementation standpoint, and unless there's a quick approxmation we could do, probably is still relatively inefficient.
```
ppgmg: The idea of doing some pre-processing to determine amount of work required before partitioning is an interesting idea... I wonder if you could estimate the amount of total work necessary without actually having to do it....

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,83 @@
...
+	for i in xrange(len(lst)):
+		lst[i] = lst[i].encode('utf-8')
+	return lst
+
+#steps through one iteration of the search. Checks to see if the node we are currently
+#processing has been visited so far. If it has, update the distances of its neighbors
+#and set their visited to false, and set your own visited to be true. 
+#If you update something new, go ahead and increment the accumulator -- you've earned it.
+def iterate(source, nodeName, nodeNeighbors, nodeDepth, nUpdate, haveVisitedBefore):
+	updates = []
```
ppgmg: Just curious about what this data structure does... does each node have an updates list, and you are storing what in this list? IDs for all nodes connected to it, or...?

HW1/P5/P5.txt
```diff
@@ -1 +1,20 @@
 # Your discussion here
+
+Shortest paths from Kevin Bacon -> Harvard University 
+
+[2729536, 5114592, 2729536, 3229511, 2729536, 4625677, 2729536, 2578703, 2729536, 1124925]
+
```
ppgmg: Does this mean you got the shortest path from Kevin to Harvard as 5 nodes?

HW1/P5/P5_BFS.py
```diff
@@ -0,0 +1,105 @@
...
+					 new_paths.append(path + [nodeName])
+			else:
+				new_paths.append([nodeName])
+			updates.append( (neighbor, ([], nodeDepth + 1, new_paths) ))
+		updatedANode = True
+	updates.append((nodeName, (nodeNeighbors, nodeDepth, [])))
+	return updates
+
+#Answer is the result of running the alg on this. Will return all possible "shortest" paths.
+answer = BFS(cleanLinks, Kevin_Bacon, Harvard_University)
```
ppgmg: Ah. I see... I may have misunderstood the question as looking for the shortest, "shortest path".

Your code is very readable.

HW1/P6/P6.py
```diff
@@ -0,0 +1,130 @@
...
+
+#The text, normally with the word first, and then the index of the word as the value
+text = text.map(lambda x: (x[1], (x[0])))
+#Shift all these back one. So now, index "n" in text is the "nth" word in Shakespeare
+#and #n# in one_shift is the "n+1"th word in Shakespeare
+one_shift = text.map(lambda (x, y): (x-1, y))
+two_shift = text.map(lambda (x, y): (x-2, y)).sortByKey()
+
+#Join and reduce by key so we can get pairs of words. For some reason
+#Reduce by keys messes with the order, so sort it afterwards.
```
ppgmg: It might have been the join...

http://stackoverflow.com/questions/29284095/which-operations-preserve-rdd-order

HW1/P6/P6.txt
```diff
@@ -1 +1,19 @@
...
+
+with wounds,as they do, doubt not but your sister: you brothers, and attendants Old John of Lancaster against the elements, 
+
+mercy outo' the way? Never. To fetch her fan, took 't upon mine ear. Beshrew his hand, thus, and such 
+
+some wrongindeed. What's yours? What's here! [Reads] 'Against the Duke of Milan his brother, and by what he writes; And 
+
+Doth longto hear me, Or at the latter end of all. Insulting Charles! Hast thou slain him, then? Henceforth be 
+
+quality, why,you shall have no more words of hers which you imagine so reserv'd. will wage against your Majesty shall
```
ppgmg: Nicely done!

HW1/P4/P4.txt
```diff
@@ -1 +1,9 @@
 # Your discussion here
+
```
ppgmg: What graph representation did you end up using?  Separate nodes and edges tables?

HW1/P2/P2.txt
patrickday9: Agreed, good point. I wonder if there is something similar in Spark to the AVX dynamic scheduling that we just learned about. Where tasks/workers/partitions who come back quicker get "more work" on the next iteration.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,38 @@
...
+sc = SparkContext()
+
+#create the x and y ranges
+X = range(2000)
+Y = range(2000)
+
+#randomly shuffle the X and Y to 
+#more evenly distribute the load
+
+random.shuffle(X)
```
patrickday9: I did the same thing to get better partitions, great minds think alike ;)

HW1/P4/P4.py
```diff
@@ -0,0 +1,61 @@
...
+#from P4_bfs import BFS, iterate
+from P4_bfs2 import BFS, iterate
+import time
+
+sc = SparkContext()
+#f = open('graph.txt', 'w')
+chars = open('chars.txt', 'w')
+
+
+def char_to_edges(lstc):
```
patrickday9: I like the way you implemented your function but I think we were suppose to do as much as we could via rdds rather than python structures. I think this is possible to do with just spark functions.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,83 @@
+
+import pyspark
+from pyspark import SparkContext
+import time
+import sys
+
+#following function does BFS
+def BFS(sc, graph, source, maxITER=10):
+	#initialize. The nodes have a key (x) which is the character name
```
patrickday9: Cool, I actually implemented the BFS in the same way with the unvisited flag. You are the only other person I have seen do it this way.

HW1/P5/P5.txt
```diff
@@ -1 +1,20 @@
...
+[(1124925, u'College_Bowl'), (2578703, u'John_Lithgow'), (3229511, u'Marisa_Silver'), (4625677, u'Six_degrees_of_separation'), (5114592, u'Time_(magazine)')]
+
+
+Shortest paths from Harvard University -> Kevin Bacon 
+120 such paths, which are listed below in numerical identifiers and then by name. 
+
```
patrickday9: I'm confused, does the list of 3 values mean that there are 3 steps between Kevin Bacon and Harvard? I see that the first value is consistent throughout the 120 paths but the last value seems to change. I guess this means that there are different references for Harvard...

HW1/P5/P5_connected_components.py
```diff
@@ -0,0 +1,105 @@
...
+#The below code tests this approach on a small section of the graph
+# f = open('ccLog.txt', 'w')
+# small_test = sc.textFile('P5_small.txt')
+# clean_small_test = small_test.map(lambda x: cleanRawLinksSymmetrically(x))
+# clean_small_test = clean_small_test.map(lambda x: x[0])
+# clean_small_test = clean_small_test.reduceByKey(lambda x, y: list(set(x + y)))
+# clean_small_test.cache()
+#[(u'1', u' 1664968'), (u'2', u' 3 747213 1664968 1691047 4095634 5535664')]
+
+def CC(graph, source, sink):
```
patrickday9: Wow you finished the connected graph, congratulations!

HW1/P6/P6.py
```diff
@@ -0,0 +1,130 @@
...
+		newWord = PMF[r]
+		sentence += newWord
+		sentence += ' '
+		firstWord = secondWord
+		secondWord = newWord
+	return sentence
+
+for i in xrange(10):
+	print >> f, generateSentence(final)
+
```
patrickday9: Well done on this problem, very clean easy to follow code!

HW1/P5/P5.txt
drishi: Hey Patrick, sorry the meaning of the things below is that the first identifier is Kevin Bacon, the second and the third are the nodes BEFORE Harvard. So each has 2 intermediary steps.

HW1/P5/P5_connected_components.py
drishi: Thank you! Did this before I realized it was extra credit -- unfortunately it runs quite slowly and while it worked for a small sample graph I had written up, I didn't run it for an adequate amount of time on AWS (or maybe its just way too slow to even work on AWS) so I don't have the relevant output for it :(.

HW1/P6/P6.py
drishi: Yeah I think you are right, thanks!

PR #253, opened 2015-10-06
HW1

ghost:


HW1/P2/P2a.py
```diff
@@ -1,3 +1,34 @@
...
+# Plot effort per partition
+plt.figure()
+plt.hist(partitionWork)
+plt.ticklabel_format(style='sci', axis='x', scilimits=(0,0))
+plt.xlabel('Computation per partition')
+plt.ylim(0,70)
+plt.ylabel('Frequency')
+plt.title('P2A: Per-partition work (default partitioning strategy)')
+plt.savefig('P2a_hist.png')
+plt.show()
```
btweinstein: Nice, I appreciate the labels. I would really bin logarithmically on the x-axis, however. Your current histogram makes it look like each partition does O(1e6) iterations. In actuality, the partitions finish over 3 orders of magnitude! The 3 orders of magnitude all fall into the first bin which is misleading.

HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
-# Your discussion here
+The histograms P2a_hist.png and P2b_hist.png show the distribution of the amount of computation carried out by each partition when running the mandelbrot calculation.  
+
+P2a_hist.png shows the effort distribution for the default partitioning strategy, while P2b_hist.png shows the effort distribution when applying partitionBy(100) during the map call that applies the mandelbrot function. The former is visibly unequally balanced, with 60 of the 100 partitions carrying out significantly less computations than the other partitions. By comparison, the latter graph is much more evenly distributed, particularly when considering that the range of the x-axis much narrower for this graph.
```
btweinstein: I'm confused...what did you do to rebalance? Running partitionBy doesn't necessarily rebalance things, right? You'd have to partition by something clever

HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
-# Your discussion here
+The histograms P2a_hist.png and P2b_hist.png show the distribution of the amount of computation carried out by each partition when running the mandelbrot calculation.  
+
+P2a_hist.png shows the effort distribution for the default partitioning strategy, while P2b_hist.png shows the effort distribution when applying partitionBy(100) during the map call that applies the mandelbrot function. The former is visibly unequally balanced, with 60 of the 100 partitions carrying out significantly less computations than the other partitions. By comparison, the latter graph is much more evenly distributed, particularly when considering that the range of the x-axis much narrower for this graph. 
+
+The second approach is more evenly balanced from a computational perspective, with each partition carrying out roughly the same amount of work and a lower chance of any one partition becoming a bottleneck. This is clearly visible in the time spent on the two collect calls:
+- Mandelbrot calculation: 30 seconds with first approach, vs 0.7 seconds with second approach.
+- Plotting histogram: 27 seconds with first approach, vs 2 seconds with second approach.
```
btweinstein: It takes 27 seconds to plot a histogram? That is crazy!

HW1/P2/P2b.py
```diff
@@ -1,3 +1,34 @@
...
+# Define key parameters
+coordWidth = 2000
+coordHeight = 2000
+noPartitions = 100
+
+# Create RDD with i, j coordinates
+pixCoordIJ = sc.parallelize([(x, y) for x in range(coordWidth) for y in range(coordHeight)])
+
+# Apply the Mandelbrot function & set number of partitions
+pixMandel = pixCoordIJ.map(lambda (i, j): ((i, j), mandelbrot(j/500.0 - 2, i/500.0 - 2))).partitionBy(noPartitions)
```
btweinstein: I'm still confused why this helps to balance the workload. I think partitionBy just sorts the RDD by the hash of the keys...you are not guaranteed that this will help balance work. Was the rebalancing simply luck, or am I missing something?

HW1/P2/P2b.py
```diff
@@ -1,3 +1,34 @@
...
+
+# Draw resulting image (using helper function)
+draw_image(pixMandel)
+
+# Calculate effort per partition
+partitionWork = sum_values_for_partitions(pixMandel).collect()
+
+# Plot effort per partition
+plt.figure()
+plt.hist(partitionWork)
```
btweinstein: Whatever you did really seemed to help balance the work, for sure!

HW1/P3/P3.py
```diff
@@ -0,0 +1,36 @@
...
+    return letters, val
+
+# Load word list
+wlist = sc.textFile('EOWL_words.txt')
+
+# Use helper function to sort letters alphabetically
+words_anagrams = wlist.map(lambda w: sortWord(w))
+
+# Combine all words that contain the same letters
+anagram_words_grouped = words_anagrams.groupByKey().mapValues(list)
```
btweinstein: This is a smart way to do things. I think I did something more silly involving counting each letter. If you just sort the word and then group, you are all set.

HW1/P3/P3.txt
```diff
@@ -1 +1 @@
-# Your discussion here
+[(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])]
```
btweinstein: Nice, I got the same thing.

HW1/P4/P4.txt
```diff
@@ -1 +1,17 @@
-# Your discussion here
+1. Graph representation:
+My graph representation includes one entry per character, where each entry has the following format:
+(character name, [connected character #1, connected character #2, ...])
+i.e. the key is the character name, and the value is a list of all (directly) connected characters.
+
+When loading the data to create the list, I clean up the data by removing quotation marks around both the character name and the comic book name. I also strip out trailing whitespace and aggregate characters with otherwise identical names (e.g. 'QUASAR DOPPELGANGER ' vs 'QUASAR DOPPELGANGER'). This results in a lower overall number of unique characters, as well as a potentially lower number of nodes touched during the breadth-first search (depending on the given start node).
```
btweinstein: This was a smart idea. I didn't do this and got slightly different answers from you; this is probably the reason why.

HW1/P4/P4.txt
```diff
@@ -1 +1,17 @@
...
+
+2. Search results:
+6403 touched nodes when starting from CAPTAIN AMERICA
```
btweinstein: Yeah, about the same answer, to within 5 nodes. Probably due to your cleaning of the initial data.

HW1/P4/P4.py
```diff
@@ -0,0 +1,47 @@
+from P4_bfs import *
+
+import itertools
+
+# Initialize Spark
+from pyspark import SparkContext
+sc = SparkContext()
+sc.setLogLevel("ERROR")
```
btweinstein: What does this do? Does it just hide the huge amounts of output?

HW1/P4/P4.py
```diff
@@ -0,0 +1,47 @@
...
+# Removes "" around names and strips out whitespace
+def split_csv(val):
+    split = val.split(',')
+    name = ','.join(split[:-1]).replace('"','').strip()
+    comic = split[-1].replace('"','')
+    return tuple([comic, name])
+
+# Helper function calculates all 2-character permutations for a comic book
+def perm_chars(val):
+    return list(itertools.permutations(val[1], 2))
```
btweinstein: Very cool, I didn't think of using itertools. It is a compact way to deal with links in our networks.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,39 @@
...
+    explored = sc.accumulator(1)
+    
+    # Keep going until there are no further nodes to expand
+    while not searchNodes.isEmpty():
+        
+        # Increment depth
+        dist += 1
+
+        # Look up all characters that are touched by all characters in the frontier
+        children = charGraph.join(searchNodes).flatMap(lambda x: x[1][0]).map(lambda x: (x, dist)).partitionBy(20)
```
btweinstein: Did you avoid a shuffle here with the join? If you partition your RDD's before the join, you can avoid the huge costs.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,39 @@
...
+        # Combine the new characters with the frontier
+        # Drop any characters that are already present with a lower distance
+        frontier = frontier.union(children).reduceByKey(min).partitionBy(20).cache()
+
+        # Filter out the nodes that characters that need to be checked in the next iteration
+        # Update accumulator for newly explored nodes
+        searchNodes = frontier.filter(lambda x: x[1]==dist).map(lambda x: nodes_update(x, explored)).partitionBy(20).cache()
+        
+    # Return accumulator value
+    return explored.value
```
btweinstein: Your code is solid and very readable, nice work!

HW1/P5/P5.txt
```diff
@@ -1 +1,130 @@
-# Your discussion here
+Path search #1: Kevin Bacon >> Harvard University
+All shortest paths:
+[u'Kevin_Bacon', u'John_Lithgow', u'Harvard_University']
+[u'Kevin_Bacon', u'Six_degrees_of_separation', u'Harvard_University']
+[u'Kevin_Bacon', u'Marisa_Silver', u'Harvard_University']
+[u'Kevin_Bacon', u'College_Bowl', u'Harvard_University']
+[u'Kevin_Bacon', u'Time_(magazine)', u'Harvard_University']
+
+Path search #2: Harvard University >> Kevin Bacon
```
btweinstein: Nice, this looks correct. I don't think you had to include every path, but it is pretty cool that you did!

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,90 @@
...
+    targetNodes = sc.parallelize([])
+    dist = 1
+    
+    # Keep looping until the target node has been found
+    while targetNodes.isEmpty():
+
+        dist += 1
+        
+        # Look up new children nodes
+        children = searchGraph.join(searchNodes).map(lambda x: (x[0], x[1][0])).partitionBy(256)
```
btweinstein: Did you find that the large number of partitions helped? I found it lead to weird out-of-memory problems. 40 was about right for me.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,90 @@
...
+        # Look up all paths that end in the target node
+        targetNodes = frontier.filter(lambda x: x[0]==endNodeID).cache()
+        
+        # Update the list of nodes to search in the next iteration
+        searchNodes = frontier.filter(lambda x: x[1][0]==dist).map(lambda x: (x[0], x[0])).distinct().subtractByKey(explored).partitionBy(256).cache()
+
+        # Update the list of explored nodes
+        explored = frontier.flatMap(lambda x: x[1][1]).map(lambda x: (x, x)).distinct().partitionBy(256).cache()
+        
+    return targetNodes.map(lambda x: x[1][1]).collect()
```
btweinstein: Collecting after every iteration seems risky; what if your dataset was larger?

HW1/P5/P5.txt
```diff
@@ -1 +1,130 @@
...
+[u'Harvard_University', u'The_New_York_Times', u'Judith_Miller_(journalist)', u'Kevin_Bacon']
+[u'Harvard_University', u'Matt_Damon', u'Massachusetts_State_Police', u'Kevin_Bacon']
+[u'Harvard_University', u'Plymouth,_Massachusetts', u'Massachusetts_State_Police', u'Kevin_Bacon']
+[u'Harvard_University', u'Texas_A&M_University', u'A_Few_Good_Men', u'Kevin_Bacon']
+[u'Harvard_University', u'Harvard_Law_School', u'A_Few_Good_Men', u'Kevin_Bacon']
+[u'Harvard_University', u'Al_Gore', u'Live_Earth_(2007_concert)', u'Kevin_Bacon']
+[u'Harvard_University', u'Cornell_University', u'Edmund_Bacon', u'Kevin_Bacon']
+[u'Harvard_University', u'2003', u'2003_in_film', u'Kevin_Bacon']
+[u'Harvard_University', u'Legally_Blonde', u'2003_in_film', u'Kevin_Bacon']
+[u'Harvard_University', u'Thomas_Dudley', u'Kyra_Sedgwick', u'Kevin_Bacon']
```
btweinstein: I guess you did not attempt connected components; oh well, I think you could have gotten it! Your code is great.

HW1/P6/P6.py
```diff
@@ -0,0 +1,78 @@
...
+sc.setLogLevel("ERROR")
+
+# Helper functions
+
+# Splits lines and removes uppercase words and digits
+def clean_lines(line):
+    result = []
+    for r in line.split():
+        upper = r.upper()
+        if r != upper and not (r.isdigit()):
```
btweinstein: Nice I used this too.

HW1/P6/P6.py
```diff
@@ -0,0 +1,78 @@
...
+# Assign an index to each word
+words_index = words.zipWithIndex()
+
+# Create 3 RDDs, each with an index incremented by 1
+words_index_1 = words_index.map(lambda w: (w[1], w[0]))
+words_index_2 = words_index_1.map(lambda w: (w[0]+1, w[1]))
+words_index_3 = words_index_2.map(lambda w: (w[0]+1, w[1]))
+
+# Group the RDDs to get three-word sequences
+words_grouped = words_index_3.join(words_index_2).join(words_index_1).map(lambda w: (w[1], 1))
```
btweinstein: Did you partition before doing this? If not the join costs could be gigantic.

HW1/P6/P6.py
```diff
@@ -0,0 +1,78 @@
...
+for i in range(noPhrases):
+    text = text + getPhrase(words_grouped_result, noWords) + '\n'
+
+# Save phrases
+text_file = open("P6.txt", "w")
+text_file.write(text)
+text_file.close()
+
+# Print phrases
+print text
```
btweinstein: Nice job!

HW1/P4/P4.py
dominedo: Yes - this way Spark only prints out actual errors, as opposed to a running log of everything it does.  Helps de-clutter and focus your attention on the output worth reading.

HW1/P5/P5_bfs.py
dominedo: It's actually just one collect after all the iterations are completed (the indentation may not be showing up correctly on your end?).  I wanted to print out all the possible paths in order to compare the number of paths from KB > HU vs HU > KB, so I figured that one collect at the end was inevitable.

HW1/P5/P5.txt
dominedo: That was mainly due to my curiosity - I wanted to compare the number of paths from KB >> HU vs from HU >> KB.  I found it interesting to see that there were so many more in one direction than the other!

HW1/P5/P5.txt
dominedo: Thanks!  Unfortunately I was very much out of time by that point.

HW1/P5/P5_bfs.py
dominedo: That's interesting - did you get out-of-memory problems locally or when running on AWS?  I found that the large number of partitions ran very smoothly on both and did help.

HW1/P6/P6.py
dominedo: No, I didn't.  I can't remember the exact runtime on this function but there were no noticeable bottlenecks.  But yes, I agree, this could have created problems for larger datasets.

HW1/P2/P2.txt
Sdeppen: Did you mean to say 30 seconds for the calculation and .7 seconds to plot for first approach and 27 seconds for the calculation and 2 seconds to plot for the second approach?

HW1/P2/P2b.py
Sdeppen: Also, when you rerun the program several times, did your graphs change or were they pretty consistent? I'm also trying to understand why partitionBy seems to have worked.

dominedo: Let's see if I can clarify the confusion!  The choice of partitionBy was specific to this problem, where you can see from the imagine that the jobs with the highest number of iterations are clustered in the center of the plot. My goal was not to come up with a catch-all partition function, but to find an approach that would work given these specific key-value pairs.

@btweinstein : Yes, my understanding is that partitionBy will do a simple hash partition if it is not passed a specific function. However, in this case the hash partition has done a good job of breaking up that cluster of high-iteration jobs, which is why I went with this approach. 

@Sdeppen : Yes, the graphs were pretty consistent when re-running the problem several times.

HW1/P2/P2.txt
dominedo: Not the actual plotting of the histogram, but the collect statement that gathers the data that is plotted.  There are two collect statements - one before drawing the image and the other before plotting the histogram - and I wanted to compare the runtime of those two collect statements across the two versions of the code.

oyuchoi, line 7: You distribution looks VERY tight--good job!

oyuchoi:
Great feedback everyone--the discussion looks great! :)

PR #255, opened 2015-10-06
HW1

araftery:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,23 @@
...
-# Your code here
+import findspark
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext()
+
+sc.setLogLevel('WARN')
+
+x = np.random.permutation(2000)
```
vivjay30: Rather than randomly permuting and then passing in to sc.parallelize, it seems easier to use sc.partitionby and pass in a lambda function of randint. Not sure which is faster since np is quite fast.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,23 @@
...
+grid = sc.parallelize(x, 10).cartesian(sc.parallelize(y, 10))
+image = grid.map(lambda (x, y): ((x, y), mandelbrot((x / 500.) - 2, (y / 500.) - 2)))
+draw_image(image)
+
+partition_values = sum_values_for_partitions(image).collect()
+plt.hist(partition_values)
+plt.xlabel('Number of Iterations')
+plt.ylabel('Number of Partitions')
+plt.title('Distribution of Partitions by Number of Iterations')
+plt.savefig('P2b_hist.png')
```
vivjay30: It seems like the graph should have less variance in how much work each partition is doing. I used a random partition as well and had much less variance. On a quick glance I can't tell why that is, but it's quite possibly due to the np.random vs partitionby(lambda x: randint) distinction that I brought up earlier

HW1/P3/P3.py
```diff
@@ -0,0 +1,22 @@
...
+    return ''.join(sorted(word))
+
+words = sc.textFile('EOWL_words.txt')
+
+# set key as string of sorted letters, then reduce by key to make a list
+# of all words per sorted combination of letters
+anagrams = words.map(lambda word: word.lower().strip().strip('\n')).map(lambda word: (sort_letters(word), [word])).reduceByKey(add)
+
+# print anagram with max words
+print anagrams.sortBy(keyfunc=lambda (anagram, words): len(words), ascending=False).take(1)
```
vivjay30: There were actually multiple anagrams with 11 orderings, you should do .take() with a higher number to see all the ones tied for max.

HW1/P4/p4_bfs.py
```diff
@@ -0,0 +1,97 @@
...
+
+    accum, add_neighbors = create_add_neighbors()
+    dist = 0
+    old_num_ones = -1
+
+    while accum.value != old_num_ones:
+        old_num_ones = accum.value
+        accum, add_neighbors = create_add_neighbors()
+        vertices = vertices.flatMap(add_neighbors).reduceByKey(collapse_nodes)
+        vertices.count()
```
vivjay30: The result of vertices.count() is not being used anywhere, why is this line present?

HW1/P6/P6.py
```diff
@@ -0,0 +1,70 @@
...
+    for i in xrange(sentence_length - 2):
+        word1, word2 = sentence[-2], sentence[-1]
+        next_word = get_next_word(word1, word2)
+        sentence.append(next_word)
+
+    return " ".join(sentence)
+
+# make 10 sentences of length 20 words
+for _ in xrange(10):
+    print make_sentence(20)
```
vivjay30: Very clean style of abstracting the functionality into discrete functions, and calling with a single line at the end

HW1/P5/P5.py
```diff
@@ -0,0 +1,158 @@
...
+
+    Vertices are of form (node, (flag, dist, [neighbors], previous))
+    """
+
+    def initialize_nodes(x):
+        # initialize flag to 0
+        flag = 0
+
+        # initialize distance to a high number
+        dist = 9999
```
vivjay30: consider using the python construct float("inf"). dist = 9999 is more of an arbitrary workaround

neelspatel:
Your code looks very clean overall, and it's well commented and easy to follow.

HW1/P3/P3.py
zezhouliu: Good point @vivjay30

HW1/P5/P5.py
zezhouliu: Yes, 9999 might not be generalizable

PR #260, opened 2015-10-06
Hw1 final submit

scheng95:


HW1/P2/P2.txt
```diff
@@ -1 +1,20 @@
...
+image of the Mandelbrot set did not require many iterations at all, and many
+others required close to our maximal number of iterations, randomly distributing
+the coordinates could only do so much to balance the load.
+
+While this strategy isn't ideal, it is strong in that it works as a base for
+many calculations in general, it is easy to implement, and it introduces very
+little calculation overhead. This is on the opposite end of the spectrum from
+first fully calculating the Mandelbrot set and using those results to define a
+function to repartition your initial RDD of coordinates into a perfectly
+load-balanced partitions.
```
Chongmoliu: nice explanation.

HW1/P3/P3.txt
```diff
@@ -1 +1,17 @@
...
+        u'reast',
+        u'resat',
+        u'stare', 
+        u'stear',
+        u'strae',
+        u'tares',
+        u'tears',
+        u'teras'
+    ]
+)]
```
Chongmoliu: nice work, I got the similar result.

HW1/P6/P6.py
```diff
@@ -0,0 +1,54 @@
...
+    current = seed[i][0]
+    prose.append([current[0], current[1]])
+    for j in range(2, WORDS_PER_PHRASE):
+        space = grouped_phrases.lookup(current)[0]
+        if len(space) == 0:
+            break
+        prose[i].append(random.choice(space))
+        current = (prose[i][j - 1], prose[i][j])
+
+print prose
```
Chongmoliu: Good way to construct the RDD. But I'm wondering could the random choice in "space" keep the bias on the count?

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,20 @@
...
+def rdd_bfs(root, graph, sc):
+    visited = sc.accumulator(set(), SetAccumulatorParam());
+    neigh_set = set([root])
+    while neigh_set:
+        curr_visited = visited.value
+        # not doing any calls to collect, only using the accumulator
+        graph.filter(lambda (k, v): k in neigh_set).foreach(lambda (k, v): visited.add(set(v)))
+        neigh_set = visited.value - curr_visited
+    # only need nodes touched, not distances
+    return visited.value
```
Chongmoliu: Good job. The way you use the accumulator looks inspiring to me.

HW1/P2/P2.txt
```diff
@@ -1 +1,20 @@
-# Your discussion here
+The strategy I used to more evenly distribute work over partitions was to
+randomize distribution of coordinate points. I found that using Spark's native
+repartition function gave a distribution of work centered closer to lower
+numbers of iteration, while first randomizing input coordinates with the
```
zhiqianseah: Nice comparison! I wonder if there is more time overhead in repartitioning. And shouldn't repartition give essentially the same results as a shuffle?

HW1/P2/P2b.py
```diff
@@ -1,3 +1,31 @@
...
+PART_DIM = 10
+PLOT_BINS = 20
+
+sc = SparkContext("local", "Mandelbrot load balanced")
+
+# randomize coordinates first, then partition as in default
+rand_coords = range(GRID_DIM)
+random.shuffle(rand_coords)
+dim_1 = sc.parallelize(rand_coords, PART_DIM)
+grid = dim_1.cartesian(dim_1)
```
zhiqianseah: I am not sure how the cartesian function does things, but I got a better shuffle results when I call "shuffle(x_y_coordinates)" directly (ie not using the cartesian)

HW1/P4/P4.py
```diff
@@ -0,0 +1,32 @@
...
+from P4_bfs import *
+
+sc = SparkContext("local", "Marvel charater graph")
+
+# read in marvel data
+marvel_csv = sc.textFile("source.csv")
+
+# parse and format into RDD
+marvel_rdd = marvel_csv.map(lambda l: l.split('"'))
+marvel_rdd = marvel_rdd.map(lambda (a, b, c, d, e): (d, b))
```
zhiqianseah: Interesting use of split. I didn't think of that!

HW1/P6/P6.py
```diff
@@ -0,0 +1,54 @@
...
+
+sc = SparkContext("local", "Shakespeare")
+
+shakes = sc.textFile("Shakespeare.txt")
+
+# take out strings that are only numbers, empty, or only upper
+shakes_words = shakes.map(lambda l: l.split(" ")).flatMap(lambda l: l)
+shakes_words = shakes_words.filter(lambda w: not w.isdigit())
+shakes_words = shakes_words.filter(lambda w: len(w) > 0)
+shakes_words = shakes_words.filter(lambda w: not w.isupper())
```
zhiqianseah: I think you missed out the words that are capitalized and ends with a period?

HW1/P6/P6.py
zhiqianseah: Yup, I agree with Chongmo, I am not sure how the bias is calculated. The way I did it is to avoid the Counter, and use the random.choice of a list with repeated entries. Simple way to ensure that a weighted choice is taken

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,20 @@
...
+    def addInPlace(self, v1, v2):
+        return v1 | v2
+
+def rdd_bfs(root, graph, sc):
+    visited = sc.accumulator(set(), SetAccumulatorParam());
+    neigh_set = set([root])
+    while neigh_set:
+        curr_visited = visited.value
+        # not doing any calls to collect, only using the accumulator
+        graph.filter(lambda (k, v): k in neigh_set).foreach(lambda (k, v): visited.add(set(v)))
```
zhiqianseah: I am not sure if the filter is in-place, but an optimization will be to filter out the neigh_set from the graph, so that you have decreasing number of nodes to work with

HW1/P6/P6.txt
```diff
@@ -1 +1,12 @@
...
+  [u'life,', u'One', u'time', u'will', u'come', u'and', u'crush', u'us', u'to', u'compel', u'this', u'offer;', u'And', u'it', u'appears', u'she', u'lives', u"unharm'd.", u'She', u'will'],
+  [u'our', u'reward,', u'thy', u'conscience', u'flies', u'out.', u"'Tis", u'well', u'blown,', u'lads.', u'This', u'morning,', u'uncle.', u'What', u'were', u'best', u'not', u'know', u'thou', u'wouldst?'],
+  [u'gentle', u'wife?', u'No', u'so,', u'my', u'lord.', u'hope,', u'my', u'verse', u'can', u'sit,', u'Your', u'own', u'desert?', u'Ay,', u'not', u'mine', u'own.', u'How', u'darkly'],
+  [u'Rome', u'Which', u'signifies', u'what', u'hate', u'feed', u'not.', u'Dost', u'hate', u'a', u'proud', u'man', u'that', u'dares', u'not', u'fight!', u'No', u'words!', u'Of', u'other'],
+  [u'these', u'years?', u'say', u'my', u'prayers,', u'would', u'repent.', u'Enter', u'Now!', u'whence', u'come', u'you?', u'Fal.', u'My', u'own', u'knee?', u'Fal.', u'My', u'own', u'flesh'],
+  [u'his', u'los', u'To', u'new', u'remembrance.', u'For', u'full', u'well', u'befits,', u'For', u'still', u"'tis", u'just.', u'Why,', u'how', u'now,', u'sons', u'and', u'brothers?', u'Tell']
+]
```
zhiqianseah: Minor point, but it will be nice if you put them into a string, rather than giving a list of words

HW1/P4/P4_bfs.py
zhiqianseah: Agreed, the way you use accumulator is really cool

HW1/P2/P2b.py
```diff
@@ -1,3 +1,31 @@
...
-# Your code here
+# constants
+GRID_DIM = 2000
+PART_DIM = 10
+PLOT_BINS = 20
+
+sc = SparkContext("local", "Mandelbrot load balanced")
+
+# randomize coordinates first, then partition as in default
+rand_coords = range(GRID_DIM)
```
oyuchoi, line 14: This partitioning seems to be done in the python driver rather than in SPARK

HW1/P6/P6.py
oyuchoi, line 26: Yes, it looks like "WORD.".isupper() returns TRUE

PR #262, opened 2015-10-07
Hw1

patrickday9:
Homework submit for HW1


HW1/P6/P6.txt
```diff
@@ -1 +1,5 @@
 # Your discussion here
+
+This was by far the best problem of the homework set! I had so much fun running the shakespeare sentance generated over and over. Here is my favorite output, it pretty much sums up this homework:
+
+Direct. All these the wretches that we find, we stoop and humble
```
foo-bar-baz-qux: Yep, this was a fun problem. I think we had to include 10 phrases with 20 words each in this file though.

HW1/P6/P6.py
```diff
@@ -0,0 +1,86 @@
...
+first_word = sample_2_word[0][0][0]; second_word = sample_2_word[0][0][1] 
+predict_word = sample_2_word[0][1] 
+
+# Initialize arrays prior to sentance builder
+sent_builder = [first_word, second_word]
+lookup_word_key = (first_word, second_word)
+shakespeare_sent = []
+
+### Build shakespeare_sent ###
+for x in xrange(0, 10):
```
foo-bar-baz-qux: I think this would generate 12 words (10 + 2 initial words) in each sentence rather than 20 words.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,109 @@
...
+	    process_nei.foreach(lambda x: accum_nei.add(1))
+	    
+	    # Check if Parent and Children are copartitioned
+	    print("Check Parent, Children, and processed Children partitions with Hero Graph")
+	    assert copartitioned(parent_rdd, hero_graph)
+	    assert copartitioned(process_nei, hero_graph)
+	    
+
+	    # Update Processed to Seen (Seen = 2)
+	    seen_nei = hero_graph.filter(lambda x: x[1][-1] == 1).map(lambda x: (x[0], (x[1][0], 2)), preservesPartitioning=True)
```
foo-bar-baz-qux: It may be less computationally expensive here just to map everything to new values conditionally, rather than performing filters, maps and unions. As in, say we update `(1, (10, 1))`, perhaps this can be done with `map(lambda (k,v):  (k, (v[0],2)) if v[1] == 1 else (k, v), preservePartitioning = True)`.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,109 @@
...
+	### SS_BDS ###
+	while(accum_update!=accum_par.value):
+	    
+	    # Get heros to process from graph, update parent accumlator
+	    parent_rdd = hero_graph.filter(lambda x: x[1][-1] == 1).join(graph_init).map(lambda x: (x[0], x[1][1]), preservesPartitioning=True)
+	    parent_count = parent_rdd.count()
+
+	    # Increment Accumlate passed on the number of parents processed
+	    accum_update = accum_par.value
+	    parent_rdd.foreach(lambda x: accum_par.add(1))
```
foo-bar-baz-qux: Isn't this the same as incrementing `accum_par` by `parent_count`? Another option might be to check the result size of `hero_graph.filter(lambda x: x[1][-1] == 1)` and exit if 0.

HW1/P5/P5.py
```diff
@@ -0,0 +1,163 @@
+#######################################################################
```
foo-bar-baz-qux: I like how well-commented and modularized your code is.

HW1/P4/P4.txt
```diff
@@ -1 +1,17 @@
...
+('CAPTAIN AMERICA', 'depth is:', 3, 'with # nodes:', 6408)
+('MISS THING/MARY', 'depth is:', 1, 'with # nodes:', 7)
+('ORWELL', 'depth is:', 1, 'with # nodes:', 9)
+
+1) What does this diameter imply about the maximum number of steps in which your search must be executed?
+
+The graph diameter is exactly what it’s name implies, it is the distance (or nodes) that we can traverse through the graph before we reach it’s maximum depth. The diameter implies the maximum number of steps to complete a depth full search from the parent node.		
+					
+2) What does it mean if a character does not have a defined distance during the search? 
+If a comic book character does not have a defined distance then he or she does not have many friends! To be more specific it means that he or she does not have any children in which they are connected to the parent node. This is completely possible when you have islanded nodes in a graph which are not connected and therefore isolated from a group of other nodes.
```
foo-bar-baz-qux: Although it could also be that the source superhero has few friends (e.g. if a source super hero has no edges, then everyone else will have an undefined distance.)

HW1/P2/P2b.py
```diff
@@ -1,3 +1,52 @@
...
+	# Create RDD and use Cartesian to Create Pixels
+	x = sc.parallelize(rand_xy, 10)
+	y = sc.parallelize(rand_xy, 10)
+	all_pixels = x.cartesian(y)
+
+	# Call Mandlebrot Function
+	mandelbrot_rdd = all_pixels.map(lambda p: (p, mandelbrot((p[0]/500.) - 2, (p[1]/500.) - 2)))
+	
+	# Repartition the Data and Draw Image
+	mandel_repart = mandelbrot_rdd.repartition(100)
```
foo-bar-baz-qux: Nice, I did this in the driver but this is more elegant.

HW1/P3/P3.txt
```diff
@@ -1 +1,4 @@
 # Your discussion here
+('aerst', 11, ['arets', 'aster', 'rates', 'reast', 'resat', 'stare', 'stear', 'strae', 'tares', 'tears', 'teras'])
```
foo-bar-baz-qux: Interesting, looks like there are multiple; I got a different one with 11 anagrams: 'aeprs'.

HW1/P3/P3.py
```diff
@@ -0,0 +1,35 @@
...
+# Create RDD of sorted letters k, and combinations of words v
+SLS_word = sort_let_seq.groupByKey().mapValues(list)
+
+# Create new RDD with Sorted Sequence, Respective len, and list of words
+final_rdd = SLS_word.map(lambda (sort, words): (sort, len(words), words))
+
+# Sort and take the first entry
+most_ana = final_rdd.sortBy(lambda x: x[1], ascending=False).first()
+print(most_ana)
+
```
georgeqi94: Clean and concise!

HW1/P2/P2.txt
```diff
@@ -1 +1,5 @@
...
+
+I choose to use pyspark's "repartition" function in order to see how it would compare to the default partitioning scheme. As shown by the P2b histogram, the partitions are more evenly spread amongst the workers. This is because repartition uses a shuffle to repartition the randomly redistribute the data. The repartition helps for this problem but we need to be wary of using it (as I found out later in the homework) as it performs costly shuffles. I also decided to permute the pixels before passing them to the mandelbrot function to ensure that the partitions have a uniform spread of easier/more difficult computations to perform. This is another step to ensure that high intensity pixels are distributed randomly across partitions and therefore enforcing workers to complete their tasks around the same time. This problem shows how pyspark uses partitioned RDDs to distribute work.
```
georgeqi94: After the shuffle, which is still faster on average? I think it would be useful if you talked a little bit about the differences in average time (given that the second has a better spread but costly shuffles), and the pros and cons of each.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,47 @@
...
+	mandel_hist_rdd = mandel_parts.collect()
+
+	# Plot and Save
+	plt.figure(figsize=(8,5))
+	plt.hist(mandel_hist_rdd)
+	plt.title("Default RDD Partition Histogram")
+	plt.xlabel("Computitions")
+	plt.ylabel("Number of Paritions")
+	plt.savefig('P2a_hist.png')
+	plt.show()
```
georgeqi94: Looks good! Clear and concise.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,52 @@
...
+	mandel_hist_repart = sum_mandel_repart.collect()
+
+	# Plot and Save
+	plt.figure(figsize=(8,5))
+	plt.hist(mandel_hist_repart)
+	plt.title("RDD Partitioned by 100 Histogram")
+	plt.xlabel("Computitions")
+	plt.ylabel("Number of Paritions")
+	plt.savefig('P2b_hist.png')
+	plt.show()
```
georgeqi94: Parts (a) and (b) look good! Clear and concise.

HW1/P4/P4.txt
georgeqi94: Right -- if they don't have a defined distance, then they're just unreachable from the source node.

HW1/P5/P5.py
```diff
@@ -0,0 +1,163 @@
...
+harvard_path_to_kevin_via_step = find_shortest(harvard_rdd, page_ind, links_rdd, harvard_path_to_kevin, kevin_bacon_name, 2)
+
+# Put all steps in a list and pickle dump
+short_path_harvard = [harvard_path_to_kevin + harvard_path_to_kevin_via_step]
+pickle.dump(short_path_harvard, open("harvard_shortest_path.pkl", "wb"))
+
+#file = open("harvard_shortest_path.txt", "w")
+#for line in kevin_list:
+#  	file.write("%s\n" % item)
+#file.close()
```
georgeqi94: The commenting was great and made it easy to follow; thanks!

HW1/P5/P5.txt
```diff
@@ -1 +1,10 @@
 # Your discussion here
+Use your code to compute one of the shortest paths from “Kevin Bacon” to “Harvard University”, and the reverse (Harvard to Mr. Bacon). Report these results in your writeup below
+
+The shortest path I found from Kevin to Harvard is as follows:
+['Kevin_Bacon', u'College_Bowl’, ‘Harvard_University’]
+
+The shortest path from Harvard_University to Kevin_Bacon was:
+['Harvard_University', u'Cornell_University', u'Edmund_Bacon', 'Kevin_Bacon']
+
+I really struggled on this problem as the code that I originally wrote for my BFS was just too computationally heavy, it would take forever to run! I got help from Patrick Kupier and this problem and he showed me a much simpler way to implement the BFS. Instead of having a flag bit for whether or not I’d seen a node (which required me to do many more filter, unions, and subtract by keys) I implemented this simpler method and everything worked fine. After weeks of struggling with this pset, I’m happy to say that it is over!
```
georgeqi94: Just curious -- how much time did your BFS end up taking to run?

HW1/P6/P6.txt
georgeqi94: ^ I think the problem spec did indeed state that we needed 10 phrases instead of just the one.

HW1/P6/P6.py
```diff
@@ -0,0 +1,86 @@
...
+text_stream_ind = filter_words.zipWithIndex()
+txt_ind_1 = text_stream_ind.map(lambda x: (x[1], x[0]))
+
+# Create the "shifted by 1" lists
+shift_1 = filter_words.collect()[1:] 
+shift_2 = shift_1[1:]
+
+# Create RDDs for shifts
+shift_1_rdd = sc.parallelize(shift_1, 10)
+shift_2_rdd = sc.parallelize(shift_2, 10)
```
georgeqi94: Instead of shifting them in the lists, you could also shift them in the RDDs and then join them

HW1/P6/P6.py
```diff
@@ -0,0 +1,86 @@
...
+lookup_word_key = (first_word, second_word)
+shakespeare_sent = []
+
+### Build shakespeare_sent ###
+for x in xrange(0, 10):
+	# Select word based on occurances in next predicted word
+    word_occur = [predict_word[0][0] for i in xrange(predict_word[0][1])]
+    
+    # Randomly select from all word choices and add to sentance
+    new_word = np.random.choice(word_occur)
```
georgeqi94: To generate the new word, instead of randomly choosing from the list comprehension (which requires O(n) space and time), you could randomly generate the index.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,109 @@
...
+	  
+	  	# Print Depth, Parents and Children
+	    print ('Graph Depth:', depth)
+	    print ("Childred currently being processed (k,v,v=1):", parent_count)
+	    print ("Parents that have been processed (k,v,v=2):", accum_par.value)
+	    
+	    # Turn Neighbors to process and update neighbor accumlator
+	    process_nei = parent_rdd.flatMap(lambda x: x[1][0]).map(lambda x: (x, (depth,1)), preservesPartitioning=True).distinct().subtractByKey(hero_graph).partitionBy(partition)
+	    process_nei.foreach(lambda x: accum_nei.add(1))
```
georgeqi94: As the person above quickly mentioned, but it seems like you could just increase accum_par and accum_nei  by the count of the RDD's.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,109 @@
...
+	processed_parent = hero_values.map(lambda x: (x[0], (0, 2)), preservesPartitioning=True)
+	parent = hero_values.map(lambda x: (x[0][0], (x[1][0], 0, 1)), preservesPartitioning=True)
+	
+	# Update parent, find neighbors, and put together
+	nei_init_rdd = parent.flatMap(lambda x: x[1][0]).map(lambda x: (x,(1,1)), preservesPartitioning=True).subtractByKey(processed_parent).partitionBy(partition)
+	hero_graph = nei_init_rdd.union(processed_parent).partitionBy(partition).cache()
+	
+	# Initialize accumlators and depth after preprocessing
+	accum_par = sc.accumulator(0)
+	accum_nei = sc.accumulator(0)
```
georgeqi94: Seems like your accumulators just act as variables -- there are other ways you can use accumulators to serve some useful functions as well

oyuchoi:
Good feedback @georgeqi94 !

PR #264, opened 2015-10-07
HW 1 

evanyao27:
I wasn't able to finish due to a really busy midterm schedule leading up to the deadline. Taking the class Pass-Fail. Hope something like this won't happen again, will try harder to finish next time.


HW1/P2/P2b.py
```diff
@@ -1,3 +1,26 @@
...
-# Your code here
+import matplotlib.pyplot as plt
+import numpy as np
+import random
+
+X = range(2000)
+Y = range(2000)
+random.shuffle(X) 
+random.shuffle(Y)
+
```
Elena-Zhao: It is interesting that you improve the partition strategy using random.shuffle, however, in order to be more targeted and efficient, customise partition strategy according to pixels distribution on the graph using spark "partitioner" could be an optimised solution.

Elena-Zhao: Where is the 'P2b.jpg'? It is supposed to be committed as well..

HW1/P4/P4.py
```diff
@@ -0,0 +1,57 @@
...
+	print "Finished Size: ", finished.count()
+	finished_neighbors = finished.filter(lambda (K,V): V[0] == i).join(edge_lookup).map(lambda (K,V): (K, (V[0][1], tuple(V[1]))))
+	# mapping to get (neighbor, source), (neighbor, source) pairs 
+	finished_neighbors = finished_neighbors.flatMap(lambda (K,V): zip(list(V[1]), [(K, V[0]) for _ in range(len(V[1]))]))
+	finished_neighbors = finished_neighbors.map(lambda (dest,source): (dest, (source, i+1)), preservesPartitioning=True)
+
+	print "Finished Neighbors, ", finished_neighbors.count()
+	possible_neighbors = finished_neighbors.leftOuterJoin(finished)
+	print "Possible Neighbors", possible_neighbors.count()
+	def addNew(key, value):
```
Elena-Zhao: Umm, anyway, better not def a function inside a while loop..

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,28 @@
+def bfs(root, edge_lookup, sc): 
+	finished = sc.parallelize([(root, 0)])
+	new_size = sc.accumulator(0)
+
+	for i in range(10): 
+		new_size.value = 0
```
Elena-Zhao: Why using "range(10)"? There are no such constraints as to limit the depth of our search to 10. Instead, you could use accumulators to record how many nodes you visited using each loop and terminates when accumulator equals to 0 meaning no nodes are found.

HW1/P2/P2.txt
```diff
@@ -1 +1,3 @@
-# Your discussion here
+In the first part, the partitioning strategy was to have adjacent points be on the same cluster. This is inefficient because if many points, all of which require a large amount of computation time, are all computed in the same RDD, then this will be slow. There will be RDD’s that are just sitting there, doing nothing while another one is still working.
```
zhiqianseah: I don't think there will be "RDD's doing nothing". If a partition has less work and finished early, the worker node will take the next partition in the queue. Your scenario may happen if the waiting queue is empty

HW1/P4/P4.py
```diff
@@ -0,0 +1,57 @@
...
+root = 'CARTER, PEGGY'
+
+finished = sc.parallelize([(root, (0, [[root]]))])
+new_size = sc.accumulator(0)
+
+while True:
+	i = 0 
+	new_size.value = 0 
+	print "Finished Size: ", finished.count()
+	finished_neighbors = finished.filter(lambda (K,V): V[0] == i).join(edge_lookup).map(lambda (K,V): (K, (V[0][1], tuple(V[1]))))
```
zhiqianseah: Is this your bfs? shouldn't it be in P4_bfs.py?

HW1/P6/P6.py
```diff
@@ -0,0 +1,35 @@
...
+triples_freq = triples.map(lambda (K,V): (V, 1)).reduceByKey(lambda x, y: x + y).cache()
+results = triples_freq.map(lambda (K,V): ((K[0], K[1]), (K[2], V))).groupByKey().map(lambda (K,V): (K, tuple(V))).cache() 
+
+next_word_lookup = dict(results.collect()) 
+
+def nextWord(word_pair): 
+	global next_word_lookup
+	return random.choice(sum(([v]* wt for (v,wt) in next_word_lookup[word_pair]),[]))
+
+starts = sc.parallelize(results.keys().takeSample(False, 10))
```
zhiqianseah: Cool that you are doing the markov chain in parallel. I just did it in serial

HW1/P6/P6.py
```diff
@@ -0,0 +1,35 @@
...
+triples = words_index1.join(words_index2).join(words_index3).map(
+	lambda (K,V): (K, (V[0][0], V[0][1], V[1]))).sortByKey().cache()
+
+triples_freq = triples.map(lambda (K,V): (V, 1)).reduceByKey(lambda x, y: x + y).cache()
+results = triples_freq.map(lambda (K,V): ((K[0], K[1]), (K[2], V))).groupByKey().map(lambda (K,V): (K, tuple(V))).cache() 
+
+next_word_lookup = dict(results.collect()) 
+
+def nextWord(word_pair): 
+	global next_word_lookup
```
zhiqianseah: I have not seen this before. Does "global" distribute the global variable to all the worker nodes? Not sure if this will work on a cluster. I think the proper way is to use sc.broadcast

HW1/P4/P4.py
```diff
@@ -0,0 +1,57 @@
...
+char_com = data.map(clean_text)
+com_char = char_com.map(lambda (x,y): (y,x)) 
+com_all_chars = com_char.groupByKey()
+
+char_chars = com_char.join(com_all_chars).values().map(lambda (k,v): (k, tuple(v)))
+edge_lookup = char_chars.reduceByKey(lambda x,y : set(x) | set(y)).cache()
+
+root = 'CARTER, PEGGY'
+
+finished = sc.parallelize([(root, (0, [[root]]))])
```
zhiqianseah: Correct me if I am wrong: You are working with two RDDs where one contains the "finished" nodes, and another with the "unfinished" nodes. You then transfer the nodes from unfinished to finished.
It might be easier to "build" the counter into the edge_lookup RDD. This saves you the trouble of working with 2 RDDs. You set the counter for each character to infinity (or some huge number), then in each iteration, you can update the counter for nodes that have been reached.

PR #265, opened 2015-10-07
HW1a

ghost:


HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+import findspark
+findspark.init()
+from pyspark import SparkContext, SparkConf
+
+conf = SparkConf().setAppName('P2').setMaster('local')
+sc = SparkContext(conf=conf)
+
+# Calculate a mandelbrot set in parallel manner (with load balancing)
+
+data = list(itertools.permutations(range(1, 2001), 2))
```
lhoangcse: nice work, the histogram below threw me off at first but I realized the axis was different than mine (I plotted load against each partition). Our partitioning results are quite similar.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+conf = SparkConf().setAppName('P2').setMaster('local')
+sc = SparkContext(conf=conf)
+
+# Find combination of letters that can produce most anagrams
+
+def anagram(x):
+	wlist = sc.textFile('./words.txt')
+	rdd1 = wlist.map(lambda x: (''.join(sorted(x)), x))
+	rdd2 = rdd1.groupByKey().map(lambda x : (x[0], len(x[1]), list(x[1])))
+	return rdd2.takeOrdered(x, key=lambda x: -x[1])
```
lhoangcse: nice & concise

HW1/P4/P4.py
```diff
@@ -0,0 +1,37 @@
...
+	two_char_distinct = comic_two_char_distinct.map(lambda (x, y): y)
+	return two_char_distinct
+
+graph = getGraph('./source.csv')
+
+def BFS(start):
+	RDD_iter = sc.parallelize([(start, 0)])
+	RDD_visited = RDD_iter
+	while True:
+		RDD_iter = graph.join(RDD_iter).distinct().map(lambda (x, y): (y[0], 0)).distinct().subtract(RDD_visited)
```
lhoangcse: I think some documentation/comments would be helpful here. Also, I guess join() & distinct() would cause a shuffle at this point since graph and RDD_iter are not copartitioned (if you use partitionBy with the same # of partitions then assuming the two RDDs have the same keys, they should end up with same partitioner). I'm not sure about union() but I think it also causes a shuffle if things are not copartitioned.

HW1/P4/P4.py
```diff
@@ -0,0 +1,37 @@
...
+		RDD_visited = RDD_visited.union(RDD_iter)
+		if RDD_iter.count() == 0: break
+	# Use of accumulators doesn't make much sense here; return RDD_visited.count() accomplishes the same
+	accum = sc.accumulator(0)
```
lhoangcse: I guess in this case since you are not setting distances for nodes and your convergence condition is "no additional node touched", the accumulator wouldn't make sense as you said. If you were recording the distances, maybe you could use it to count the total distances over all nodes and check for convergence if the total does not change over iteration.

HW1/P6/P6.txt
```diff
@@ -1 +1,12 @@
-# Your discussion here
+Ten randomly generated phrases:
+
+1.) indeed, that baes like a man of his own hand did make, Breathed forth the stocks! You stubborn ancient knave,
```
lhoangcse: Nice work. It looks like you are taking the most likely next word (instead of sampling from the count distribution) so some of the generated sentences have repeated phrases (e.g. Breathed forth the stocks!). However I think it's a fine solution.

lhoangcse:
Nice work overall. My main feedback would be to add comments & documentation so it's easier for others to understand the code since these problems were more complex.

HW1/P6/P6.txt
newtestsubject: Yes, I am always taking the most likely next word. That is how I initially understood assignment. After rereading it now it seems that you are right, I probably should have done a weighted sampling instead.

HW1/P2/P2b.py
jpbm: importing spark like that is very nice, i got stuck with copying everything into a terminal window. i'll work to implement this for myself.

also, the "itertools.permutation" is very elegant. as comparison, here is what i did "manually":

+q = range(0,4000)
+X = sc.parallelize(q,100)
+coordinates = X.flatMap(lambda x: [(x,y) for y in q])

HW1/P2/P2.txt
```diff
@@ -1 +1,5 @@
-# Your discussion here
+To equally distribute the computing load between the partitions I shuflledled the list of (x, y) value before turning it into an RDD. This  requires very little time (< 1s) compared to calculation of the mandelbrot set (about 180s).
+
+The result is that the workload is very equally distributed between partitions (2,000,000 +/- 150,000) compared to before (5,000,000 +/- 4,000,000).
+
+If the list (alternatively the list can be turned into an RDD before shuffling) becomes so large that the shuffling takes a long time compared to the average processing time (mandelbrot) of an unshufflled RDD and furthermore there are many processors available to process many small partitions in parallel then shuffling the list might actually decrease the performance.
```
jpbm: great idea that should work very generally.

HW1/P3/P3.txt
```diff
@@ -1 +1,7 @@
 # Your discussion here
+
+Letter combinations with the most anagrams:
+
+ (u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])
+ (u'aelst', 11, [u'least', u'leats', u'salet', u'slate', u'stale', u'steal', u'stela', u'taels', u'tales', u'teals', u'tesla']) 
+ (u'aeprs', 11, [u'asper', u'pares', u'parse', u'pears', u'prase', u'presa', u'rapes', u'reaps', u'spaer', u'spare', u'spear'])
```
jpbm: I have the same.

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+shuffle(data)
```
adisuissa, line 17: Both the data and shuffle can be done in spark.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+import findspark
+findspark.init()
+from pyspark import SparkContext, SparkConf
+
+conf = SparkConf().setAppName('P2').setMaster('local')
+sc = SparkContext(conf=conf)
+
+# Find combination of letters that can produce most anagrams
+
+def anagram(x):
```
adisuissa, line 13: You should probably use different variable name (x as the parameter for anagram, and for the transformations may be confusing).

HW1/P4/P4.py
adisuissa, line 28: If you are avoiding the actual distance of the nodes from the root, you can probably change the map (no need for (y[0], 0)).

HW1/P4/P4.txt
```diff
@@ -1 +1,11 @@
...
+
+1.) The graph conssists of all links of two characters. The graph is symmetric, meaning that if (char1, char2) is part of the graph so is (char2, char1).
+
+2.) Number of touched nodes (including starting node):
+    
+    'CAPTAIN AMERICA': 6408
+    'MISS THING/MARY': 7
+    'ORWELL': 9
+
+3.) The diameter of the graph is the number of steps BFS needs to execute to touch all nodes that are directly/inderectly connected to the starting node. If a node does not have a defined distance to the starting node, it means that there is no connection between those two nodes.
```
adisuissa, line 11: Note that a solution to P5 is missing from this repo.

HW1/P6/P6.py
```diff
@@ -0,0 +1,72 @@
...
+	words_two_list1 = words_three_no_index.groupByKey()
+	words_two_list2 = words_two_list1.map(lambda ((a, b,), c): ((a, b), list(c)))
+	words_two_list3 = words_two_list2.map(lambda ((a, b,), c): ((a, b), counter(c)))
+	return words_two_list3
+
+
+graph = getGraph('./shakespeare.txt')
+
+
+def highest_count(l):
```
adisuissa, line 46: It seems that you don't choose a word according to its number of appearances (weighted distribution).

PR #266, opened 2015-10-07
Hw1 Submission

sbalanovich:
2 Late Days Taken


HW1/P2/P2b.py
```diff
@@ -1,3 +1,25 @@
...
+import random
 
-# Your code here
+# Set up partitions and context
+N = 2000
+sc = SparkContext("local", "P2b")
+
+# Partition a 10x10 cartesian grid of workers
+# Same as P2a, except the ordering is randomized
+rdd = sc.parallelize(sorted(range(N), key=lambda k: random.random()), 10)
```
Elena-Zhao: Partition using random could be a better solution, however, in order to be more targeted and efficient, customise partition strategy according to pixels distribution on the graph using spark "partitioner" could be an optimised solution.

HW1/P5/P5.py
```diff
@@ -0,0 +1,54 @@
...
+	return (int(frm), map(int, to.split(" ")))
+
+# Use functor to create a graph of links similar to P4 graph
+graph = links.map(lambda link: createGraphComponent(link)).cache()
+print graph.collect()
+
+# Zip up page names and take into account 1-indexing
+page_names = page_names.zipWithIndex().map(lambda (page, i): (page, i+1)).sortByKey()
+# Use this one for index lookups
+page_names_idx = page_names.zipWithIndex().map(lambda (page, i): (i+1, page)).sortByKey()
```
Elena-Zhao: Great try by using sortBykey before lookup! This could significantly improve lookup efficiency.

HW1/P6/P6.py
```diff
@@ -0,0 +1,91 @@
...
+			nextWord = np.random.choice(words, p=allProbas)
+			sentence = sentence + " " + nextWord
+
+			# Bit hacky, but the code expects start to have a list of follow-up
+			# words, so an empty list will suffice as a placeholder
+			# start = ((word2, word3), [])
+			start = ((start[0][1], nextWord), [])
+		sentences.append(sentence)
+	return sentences
+
```
Elena-Zhao: Instead of generating sentences sequentially, we could use rdd to do it in parallel. 

P.S., your codes are so well commented! Well done!

HW1/P2/P2.txt
```diff
@@ -1 +1,5 @@
...
+
+One other note is that this randomization, while effective in increasing the spread of the histogram and distributing the load such that the entire parallelization ran faster as a whole, was still quite arbitrary. Though splitting by less arbitrary parameters was not found to be any more effective than simply randomizing, it is potentially possible to take advantage of our knowledge of the problem and its difficulty (concentrated computation on central white pixels) and to devise a scheme to capitalize on this and perform even better. Though I could not think of one for this assignment, this is definitely something to think about for future problems.
```
chuckyouliu: Great writeup - I made a box of the heavy computation in the center and split it evenly but it honestly wasn't much better than the randomized algorithm.

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+
+# Group by key to find anagrams. Then introduce counts via a map
+# (sorted, [word1, word2, ...], count)
+groupedCountedWords = sortedWords.groupByKey().map(lambda (w, lst): (w, list(lst), len(lst)))
+
+# Run a reduction to compute the max counted anagram
+maxCount = groupedCountedWords.reduce(lambda tup1, tup2: max(tup1, tup2, key = lambda tup: tup[2]))
+
+# Print the result
+print "FINAL result: ", maxCount
```
chuckyouliu: As discussed in class a reduceByKey is better than a groupByKey -> reduce, so would be preferable

sbalanovich: You're totally right, forgot to go back and fix this, nice catch!

HW1/P2/P2b.py
sbalanovich: Looking at the "Partitioner" docs right now and I don't quite see how this would work unless I actually hard-coded positions based on my prior understanding of Mandelbrot functions, which seems a bit iffy to me.

oyuchoi, line 11: As mentioned about, you could use a partitioner

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,65 @@
...
+		# print stepThrough.take(3)
+		# print "------------"
+
+		# Combine touched neighbors of given character through a simple reduction. For every 2 neighbors
+		# combine their respective neighbors into a single de-duped set-list, find the minimum depth, 
+		# and mark visited as true if at least one was visited. This way we can continue traversing as before.
+		# (char, [[neighb1, neighb2, ...], depth, visited])
+		combinedNeighbs = stepThrough.reduceByKey(lambda n1, n2: ([list(set(n1[0] + n2[0])), min(n1[1], n2[1]), n1[2] or n2[2]]))
+		# print combinedNeighbs.take(3)
+		# print "############"
```
oyuchoi, line 29: There are a significant number of commented out code here.

sbalanovich, line 29: Yes, these were tests I ran prior to submission that I chose to omit but not delete right before I pushed. Is this bad practice? Should I have left these as uncommented code or removed them entirely?

PR #269, opened 2015-10-07
Hw1

jonahkall:
Jonah Kallenbach HW1 submission.


HW1/P2/.ipynb_checkpoints/205_2-checkpoint.ipynb
```diff
@@ -0,0 +1,310 @@
+{
```
sbalanovich: In future, just for ease of readability, could you potentially gitignore the iPython Notebook stuff? Unless you actually intend for this to be looked at - it's kind of hard to comment on a .ipynb file though

HW1/P2/P2.txt
```diff
@@ -1 +1,9 @@
 # Your discussion here
+The graph for part a is scaled by 1e7, which is why it
+looks relatively clustered. In fact, the partitioning is quite poor:
+some partitions did 0 or close to 0 work, while other partitions did
+huge amounts of work.  The graph for part b on the other hand is tightly
+clustered around a small range of values just above 200,000. Thus we see that
+the solution of a random repartition via PartitionBy balances work *extremely* well.
```
sbalanovich: I think *extremely* might be a bit of an over-exaggeration :)
I'm sure if we really wanted to, we could hardcode some configuration of workers to fully optimize for solving this problem, but I do agree with you that given the simplicity of the solution and the low cost, this solution works relatively well!

HW1/P2/P2a.py
```diff
@@ -1,3 +1,30 @@
...
+
+
+rdd1 = np.arange(2000)
+rdd2 = np.arange(2000)
+
+rdd1 = sc.parallelize(rdd1, 10)
+rdd2 = sc.parallelize(rdd2, 10)
+
+# Create the pixels rdd
+pixels = rdd1.cartesian(rdd2)
```
sbalanovich: In order to avoid all of the above duplication, it might have been cleaner to simply make a single rdd and then cartesian it with itself (ie `rdd = sc.parallelize(np.arange(2000), 10)` and then `rdd.cartesian(rdd)`)

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
+# Randomly shuffle the pixels data
+pixels = pixels.partitionBy(100)
+
+res = pixels.map(lambda a: (a,mandelbrot((a[1]/500.0) - 2,(a[0]/500.0) - 2)))
+draw_image(res)
+b = sum_values_for_partitions(res)
+to_plot = b.collect()
+
+plt.hist(to_plot)
+plt.title('Load distribution with default partitioning')
```
sbalanovich: Isn't this no longer "default" partitioning? :grinning:

HW1/P3/P3.py
```diff
@@ -0,0 +1,17 @@
...
+import pyspark
+sc = pyspark.SparkContext(appName="P3")
+
+# Load the data (I parsed the data into one file called 'words')
+words_list = sc.textFile("words")
+
+# Hash every word to its sorted list of characters, map everything with those
+# letters to that word.
+sorted_words = words_list.map(lambda x: (''.join(sorted(x)), (1, [x])))
+output = sorted_words.reduceByKey(lambda a,b: (a[0] + b[0], a[1] + b[1]))
```
sbalanovich: Excellent!

HW1/P3/P3.txt
```diff
@@ -1 +1,7 @@
 # Your discussion here
+(u'aerst', (11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras']))
+
+
+Note: it was a tie!
+
+(u'aerst', (11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])), (u'aelst', (11, [u'least', u'leats', u'salet', u'slate', u'stale', u'steal', u'stela', u'taels', u'tales', u'teals', u'tesla'])), (u'aeprs', (11, [u'asper', u'pares', u'parse', u'pears', u'prase', u'presa', u'rapes', u'reaps', u'spaer', u'spare', u'spear']))
```
sbalanovich: Nice catch!

HW1/P4/P4.txt
```diff
@@ -1 +1,19 @@
...
+as a stop condition!
+
+If a character does not have a defined distance in the graph, this means that that character
+is unreachable, a.k.a. disconnected from the source character (different connected component).
+
+My graph representation was an adjacency list: an RDD which maps each character name to a list
+of characters with whom they have appeared in a comic.
+
+The number of touched nodes in the searches were:
+
```
sbalanovich: These numbers look right to me!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,20 @@
...
+# distance from src_name.
+def bfs(adj_list_rdd, src_name, sc):
+    accum = sc.accumulator(0)
+    adj_list_rdd.cache()
+    distances = sc.parallelize([(src_name, 0)])
+    i = 0
+    curr = 0
+    c = adj_list_rdd.count()
+    last = -1
+    while accum.value != last:
```
sbalanovich: Maybe I'm misunderstanding this, but this seems like a strange approach. You're checking whether the accumulator value is equal to `last` and if it's not, you set `last` to the value, set the value to 0, and then set the accumulator value equal to the number of distances you got. Does the loop ever break when `last` is not 0? It looks to me as though you're always doing one extra iteration - it might be cleaner just to reset the accumulator on each iteration and then break the loop if it's 0 at the start of any one iteration.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,18 @@
+# Takes in a src node and a destination node, and returns an rdd of
+# every shortest path from src to dest in the graph.
+# Caution: if src is not connected to dest, this runs forever
+def find_all_shortest_paths(src, dest, graph, sc):
+  accum = sc.accumulator(0)
+  num_parts = 64
+  graph = graph.partitionBy(num_parts).cache()
+  distances = sc.parallelize([(src, [src])]).partitionBy(num_parts)
+  while 1:
```
sbalanovich: I think `while True` may be the more accepted "infinite loop" notation in Python.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,18 @@
...
+    accum.value = 0
+    print "new iteration"
+    assert(distances.partitioner == graph.partitioner)
+    distances = distances.join(graph).map(lambda (k, v): (v[1], v[0] + [v[1]])).partitionBy(num_parts)
+    test = distances.filter(lambda (k, v): k == dest)
+    test.foreach(lambda x: accum.add(1))
+    if accum.value > 0:
```
sbalanovich: Yeah, this is what I was talking about in my comment on your P4_bfs :+1:

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,18 @@
+# Takes in a src node and a destination node, and returns an rdd of
+# every shortest path from src to dest in the graph.
+# Caution: if src is not connected to dest, this runs forever
+def find_all_shortest_paths(src, dest, graph, sc):
```
sbalanovich: Excellent! Shortest, most concise implementation I've seen for this problem set, well done.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+sc = pyspark.SparkContext(appName="P4")
+
+f = open("pg100.txt")
+raw_text = f.read()
+output = raw_text.split()
+
+# Clean the words in accordance with the spec
+def cleanUp(word):
+    is_word = False
+    try:
```
sbalanovich: This try/catch is a bit silly, I think you should have just used `isdigit()`

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+    if is_word == False:
+        return False
+    if word.isupper():
+        return False
+    if word[-1] == ".":
+        if word[::-1].isupper():
+            return False
+    return True
+
+# Basic tests on cleanUp
```
sbalanovich: Excellent! Nice of you to actually test all posed edge cases.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+assert(cleanUp("51235151") == False)
+assert(cleanUp("AJSDFJKAK") == False)
+assert(cleanUp("AJSDFJKAK.") == False)
+assert(cleanUp("AJSDFAKSFIWEIEIEIEJKAK.") == False)
+
+# clean the data
+cleaned = filter(cleanUp, output)
+
+trigrams = []
+for i in range(len(cleaned) - 2):
```
sbalanovich: Interesting, though this probably should have been parallelized. I would recommend looking into `zipWithIndex()` with offsets to create these trigrams.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+assert(cleanUp("bllahsk28askldH") == True)
+assert(cleanUp("alksdf.") == True)
+assert(cleanUp("1234") == False)
+assert(cleanUp("51235151") == False)
+assert(cleanUp("AJSDFJKAK") == False)
+assert(cleanUp("AJSDFJKAK.") == False)
+assert(cleanUp("AJSDFAKSFIWEIEIEIEJKAK.") == False)
+
+# clean the data
+cleaned = filter(cleanUp, output)
```
sbalanovich: This should be parallelized (a filter on an RDD).

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
 
-# Your code here
+import findspark
+findspark.init()
+
+import pyspark
+sc = pyspark.SparkContext(appName="P2")
+
+rdd1 = np.arange(2000)
+rdd2 = np.arange(2000)
```
adisuissa, line 13: You should have used sc.range to create the ranges and parallelize them.

HW1/P3/P3.py
adisuissa, line 15: Indeed, very nice implementation!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,20 @@
...
+        last = accum.value
+        accum.value = 0
+        neighbors_rdd = distances.filter(lambda x: x[1] == curr).join(adj_list_rdd) \
+                         .flatMap(lambda (key, (num, l)): [(a, curr + 1) for a in l])
+        distances = distances.union(neighbors_rdd).reduceByKey(lambda a, b: min(a,b)).cache()
+        curr += 1
+        distances.foreach(lambda x: accum.add(1))
```
adisuissa, line 19: I think you could have increased the accumulator during the previous reduceByKey function.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,18 @@
...
+  num_parts = 64
+  graph = graph.partitionBy(num_parts).cache()
+  distances = sc.parallelize([(src, [src])]).partitionBy(num_parts)
+  while 1:
+    accum.value = 0
+    print "new iteration"
+    assert(distances.partitioner == graph.partitioner)
+    distances = distances.join(graph).map(lambda (k, v): (v[1], v[0] + [v[1]])).partitionBy(num_parts)
+    test = distances.filter(lambda (k, v): k == dest)
+    test.foreach(lambda x: accum.add(1))
```
adisuissa, line 15: I think you could have used a map instead of filter, and increase the accumulator for the nodes with the distance == dest.

HW1/P6/P6.py
```diff
@@ -0,0 +1,95 @@
...
+# Draw a sample from a list of (next_word, count) pairs
+def draw_biased_sample(seq):
+    seqsum = float(sum([x[1] for x in seq]))
+    probs = [x[1]/seqsum for x in seq]
+    y = random.random()
+    for a in xrange(len(probs)):
+        y -= probs[a]
+        if y <= 0:
+            break
+    return seq[a][0]
```
adisuissa, line 82: You can use numpy's random.choice, to make this more clear.

HW1/P6/P6.py
adisuissa, line 44: Indeed, the creation of the trigrams should have been done in spark.

PR #272, opened 2015-10-08
Hw1

vinayps:


HW1/P2/P2.txt
```diff
@@ -1 +1,3 @@
 # Your discussion here
+1) The partitioning chosen was to uniformly randomly distribute the pixels before the mandelbrot computation. This choice was made based on the first histogram with the default partitioning. In the first histogram, the majority (over 85%) of the pixels had very few iterations. Thus, uniformly mixing the few high iteration pixels along with the majority low iteration pixels intuitively made sense as a more equitable partitioning strategy. The second histogram, while not being uniform in frequency across pixel bins is still much better distributed than the original histogram (it appears to resemble a normal distribution).
```
dekatzenel: Could have been more specific (e.g. numerical examples) in discussing load balancing.

HW1/P2/P2.txt
```diff
@@ -1 +1,3 @@
...
+2) Using partitionBy seems to be expensive due to the shuffle costs. While partitionBy runs, there is a lot of shuffle write and then for the subsequent step there is a lot of shuffle read. Both these steps weren't required in the default version. So, the price paid for better load balancing is the more expensive shuffle costs.
```
dekatzenel: How steep is the price?

HW1/P2/P2.txt
dekatzenel: Also, watch your orders of magnitude. Yes, it's not strictly uniform, but the range is much smaller.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,25 @@
 from P2 import *
+import findspark
+findspark.init()
+import pyspark
 
 # Your code here
+def main():
+    sc = pyspark.SparkContext()
+    arr = [i for i in xrange(2000)]
+    rdd = sc.parallelize( arr, 10 )
```
dekatzenel: Consider sc.range() rather than sc.parallelize on a separately generated list.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,25 @@
...
+    rdd = rdd.cartesian( rdd )
+    rdd.cache()
```
dekatzenel: Nice cache timing

HW1/P2/P2b.py
```diff
@@ -1,3 +1,28 @@
...
+import pyspark
 
 # Your code here
+def main():
+    sc = pyspark.SparkContext()
+    arr = [i for i in xrange(2000)]
+    rdd = sc.parallelize( arr, 10 )
+    rdd = rdd.cartesian( rdd )
+    rdd.cache()
+    rdd_xy = rdd.map( lambda ( i, j ): ( ( i, j ), ( (float(j)/500) - 2, (float(i)/500) - 2 ) ) ).partitionBy( 100, partitionFunc = lambda x: int( np.random.uniform(0,100) ) )
```
dekatzenel: Clever partition function

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+findspark.init()
+import pyspark
+
+def main():
+    sc = pyspark.SparkContext()
+    wlist = sc.wholeTextFiles("./CSV Format")
+    combined_wlist = wlist.flatMap( lambda (a,b): b.split() )
+    combined_wlist_sort = combined_wlist.map( lambda x: ( "".join( sorted( x ) ), [x] ) )
+    combined_wlist_sort_reduce = combined_wlist_sort.reduceByKey( lambda x,y: x + y )
+    combined_wlist_sort_reduce_count = combined_wlist_sort_reduce.map( lambda (x,y): (x, [y]) if ( type(y) is not list ) else (x,
```
dekatzenel: When would type(y) not be a list? Aren't you guaranteeing its list format in line 9?

HW1/P3/P3.py
```diff
@@ -0,0 +1,19 @@
...
+import pyspark
+
+def main():
+    sc = pyspark.SparkContext()
+    wlist = sc.wholeTextFiles("./CSV Format")
+    combined_wlist = wlist.flatMap( lambda (a,b): b.split() )
+    combined_wlist_sort = combined_wlist.map( lambda x: ( "".join( sorted( x ) ), [x] ) )
+    combined_wlist_sort_reduce = combined_wlist_sort.reduceByKey( lambda x,y: x + y )
+    combined_wlist_sort_reduce_count = combined_wlist_sort_reduce.map( lambda (x,y): (x, [y]) if ( type(y) is not list ) else (x, 
+    y) ).map( lambda (x,y): ( x, len(y), y) )
```
dekatzenel: This probably shouldn't be a separate line. Copied the code over from iPython?

HW1/P4/P4.py
```diff
@@ -0,0 +1,32 @@
+from P4_bfs import *
+import findspark
+findspark.init()
+import pyspark
+
+def preprocess_data( filename, sc ):
+    textFile = sc.textFile( filename )
+    textFile.cache()
+    textFile_pre_process = textFile.map( lambda x: ( x.split( "\"" )[3], x.split( "\"" )[1] ) ).partitionBy(num_of_partitions)
```
dekatzenel: In Python, single quotes (') are your friend when you actually want to match on double quotes and vice versa

HW1/P4/P4.py
```diff
@@ -0,0 +1,32 @@
...
+    issue_char_mapping = textFile_pre_process.mapValues( lambda y: [y] ).reduceByKey( lambda a, b: a + b, numPartitions = num_of_partitions )
+    assert copartitioned( textFile_pre_process, issue_char_mapping )
+    char_issue_join = textFile_pre_process.join( issue_char_mapping ).partitionBy(num_of_partitions)
+    char_adjacency_graph = char_issue_join.map( lambda ( x, y ): y ).reduceByKey( lambda x, y: x + y ).map( lambda ( x, y ): ( x, list( set( [q for q in y if q != x] ) ) ), preservesPartitioning = True ) 
+    graph = char_adjacency_graph
```
dekatzenel: Why don't you just cache and return char_adjacency_graph?

HW1/P4/P4.py
```diff
@@ -0,0 +1,32 @@
...
+findspark.init()
+import pyspark
+
+def preprocess_data( filename, sc ):
+    textFile = sc.textFile( filename )
+    textFile.cache()
+    textFile_pre_process = textFile.map( lambda x: ( x.split( "\"" )[3], x.split( "\"" )[1] ) ).partitionBy(num_of_partitions)
+    issue_char_mapping = textFile_pre_process.mapValues( lambda y: [y] ).reduceByKey( lambda a, b: a + b, numPartitions = num_of_partitions )
+    assert copartitioned( textFile_pre_process, issue_char_mapping )
+    char_issue_join = textFile_pre_process.join( issue_char_mapping ).partitionBy(num_of_partitions)
```
dekatzenel: You could have skipped issue_char_mapping entirely by joining textFile_pre_process with itself. This would also have simplified subsequent processing steps.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,31 @@
...
+num_of_partitions = 2
+
+def copartitioned(RDD1, RDD2):
+    "check if two RDDs are copartitioned"
+    return RDD1.partitioner == RDD2.partitioner
+
+def SSBFS( source, graph, accum ): 
+    graph = graph.map( lambda ( x, y ): ( x, ( -1, y ) ) if x != source else ( x, ( 0, y ) ), preservesPartitioning = True )
+    graph.cache()
+    i = 0
```
dekatzenel: nit: you have 0 hardcoded 2x, you could have separated it out into an initial_distance variable or somesuch

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,31 @@
...
+    def adder( (x, y) ):
+        accum.add(1)
+        return (x, y[1][0])
+    while( True ):
+        source = graph.filter( lambda (x,y): y[0] == i ).mapValues( lambda y: y[1] ).flatMapValues( lambda y: y )
+        assert copartitioned( graph, source )
+        accum_before = accum.value
+        not_visited = graph.filter( lambda (x,y): y[0] == -1 ).partitionBy(num_of_partitions)
+        source = source.map( lambda (x,y) : (y, 1) ).distinct().join( not_visited ).map( lambda (x,y): adder( ( x, y ) ) ).partitionBy(num_of_partitions)
```
dekatzenel: This line of code took me several minutes to understand. It may have been helpful to break it up into multiple lines (one of which is labeled neighbors or next_nodes or somesuch) and/or add comments about non-obvious things like using 1 as a dummy value and using adder to get rid of the dummy value.

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,31 @@
...
+    return RDD1.partitioner == RDD2.partitioner
+
+def SSBFS( source, graph, accum ): 
+    graph = graph.map( lambda ( x, y ): ( x, ( -1, y ) ) if x != source else ( x, ( 0, y ) ), preservesPartitioning = True )
+    graph.cache()
+    i = 0
+    def adder( (x, y) ):
+        accum.add(1)
+        return (x, y[1][0])
+    while( True ):
```
dekatzenel: while(True) loops with break statements are often tempting, but it's generally good practice to duplicate the break condition inside of the while loop as an extra guard against infinite looping

HW1/P5/P5.txt
```diff
@@ -1 +1,9 @@
 # Your discussion here
+1) Kevin Bacon --> Harvard University
+[(2, ['2152782']), (1, [u'1124925', u'5114592', u'4625677', u'3229511', u'2578703']), (0, ['2729536'])]
```
dekatzenel: Is this all of the shortest paths? If so, impressive.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,82 @@
...
+def preprocess_titles( titles, start_end_pairs ):
+    titles_index = titles.zipWithIndex().map( lambda (x,y): (x, y+1))
+    titles_index.count()
+    titles_index.cache()
+    pair_dict = []
+    for pair in start_end_pairs:
+        start_index = titles_index.lookup( pair[0] ) #print titles_index.lookup( "Kevin_Bacon") [2729536]
+        end_index = titles_index.lookup( pair[1] ) #[2152782]
+        pair_dict = pair_dict + [ { "start_index": str(start_index[0]), "end_index": str(end_index[0]) } ]
+    return pair_dict
```
dekatzenel: Smart to only stringify the necessary indices.

HW1/P5/P5_bfs.py
```diff
@@ -0,0 +1,82 @@
+from P5_bfs import *
```
dekatzenel: This is P5_bfs. Do you mean P4_bfs?

HW1/P6/P6.py
```diff
@@ -0,0 +1,59 @@
...
+from functools import partial
+import findspark
+findspark.init()
+import pyspark
+
+
+def pre_process_data( filename, sc ):
+    wlist = sc.textFile( filename )
+    wlist_words = wlist.flatMap( lambda x: x.split() )
+    wlist_words_filter = wlist_words.filter( lambda x: not( x.isdigit() or x.isupper() or ( x[:-1].isupper() and x[-1:] == "." ) ) )
```
dekatzenel: According to the documentation on isupper, it will "Return true if all cased characters in the string are uppercase and there is at least one cased character, false otherwise." You want to check whether words containing only letters have all uppercase letters. This preprocessing removes things that it shouldn't. See https://piazza.com/class/icqfmi2aryn6yv?cid=226

HW1/P6/P6.py
```diff
@@ -0,0 +1,59 @@
...
+    wlist_words_index = wlist_words_filter.zipWithIndex().map( lambda (x,y): (y,x) )
+    wlist_words_index_2 = wlist_words_index
+    size = wlist_words_index_2.count()
+    wlist_words_index_2 = wlist_words_index_2.map( lambda ( x, y ): ( x-1, y ) if x != 0 else ( size, y ) )
```
dekatzenel: Line 13 is unnecessary. You could call .count() on the original and use the original as the base for 2

HW1/P6/P6.py
```diff
@@ -0,0 +1,59 @@
...
+    wlist_words_2 = wlist_words_index.join( wlist_words_index_2 )
+    wlist_words_index_2 = wlist_words_index_2.map( lambda ( x, y ): ( x-1, y ) if x != 0 else ( size, y ) )
```
dekatzenel: I don't think that wraparound is the correct behavior in this context because the first word does not follow the last word.

HW1/P6/P6.py
```diff
@@ -0,0 +1,59 @@
...
+    wlist_words_final = wlist_words_2.join( wlist_words_index_2 ).map( lambda ( x, y ): ( ( y[0][0], y[0][1], y[1] ) , 1 ) ).reduceByKey( lambda a,b: a + b ).map( lambda ( x, y ): ( ( x[0], x[1] ), [ ( x[2], y ) ] ) ).reduceByKey( lambda a,b: a + b)
```
dekatzenel: I couldn't figure out how to do this in Spark, so I just used regular Python. Well done.

HW1/P6/P6.py
```diff
@@ -0,0 +1,59 @@
...
+    cdf = 0
+    for key in transition_prob_final_sorted_keys:
+        cdf = transition_prob_final[ key ] + cdf
+        if unif < cdf:
+            return key
+
+
+def main():
+    sentence_length = 20
+    sentence_count = 10
```
dekatzenel: Good style making a variable for these.

dekatzenel:
Overall, solid work. Very few things actually wrong and whose code is always as elegant and efficient as possible?

HW1/P2/P2.txt
vinayps: @dekatzenel - agreed. I should have more explicitly pointed that out. Thanks

HW1/P2/P2a.py
vinayps: Thanks, thats useful

HW1/P3/P3.py
vinayps: yeah - I should have indented it more in line with PEP8

HW1/P3/P3.py
vinayps: When there isnt a common key to join on, y would simply be NULL. I was handling that case

HW1/P4/P4.py
vinayps: thanks - thats useful to know

HW1/P4/P4.py
vinayps: yeah - this is extraneous. nice catch

HW1/P4/P4_bfs.py
vinayps: yes, that should have been a constant with some commentary

HW1/P4/P4.py
vinayps: I thought this made it more readable. The individual RDDs are more clearly labeled and the logic is easier to follow

HW1/P5/P5.txt
vinayps: yes. there are loads more the other way

HW1/P4/P4_bfs.py
vinayps: I think the dummy value required some commentary. But the adder was primarily designed to bring the accumulators into play. 
Also, when considering the variable naming, I thought this would be easy to follow while reading the code. I think breaking it up into individual RDDs raises the memory overhead and so opted for this approach.

HW1/P5/P5_bfs.py
vinayps: Ah, nice catch. Probably got missed when I was cleaning up the library lists.

HW1/P6/P6.py
vinayps: agreed

HW1/P6/P6.py
vinayps: good catch

HW1/P6/P6.py
vinayps: It shouldnt matter. The last two words wont play a role

HW1/P3/CSV Format/J Words.csv
```diff
@@ -0,0 +1,1286 @@
+jab
```
annawhitney: You should make sure to store large files outside your repo, or else add them to your `.gitignore` file so they don't get committed - they make the `diff` much harder to read.

HW1/P6/P6.py
annawhitney: One option to deal with this could be to check `isupper` on the result of stripping all non-alpha characters from the string.

HW1/P6/P6.py
annawhitney: Ray did suggest this as a possible way to avoid running out of words in the case that the last two words didn't appear together anywhere else in the corpus (which could cause your sentences to end prematurely).

HW1/P3/CSV Format/J Words.csv
vinayps: agreed. I thought I had added them to .gitignore but clearly I hadnt.

PR #276, opened 2015-10-13
P2, P3, P4, and P6

Sdeppen:
Didn’t get to finish P4_bfs or P6… :(


HW1/P3/P3.txt
```diff
@@ -1 +1 @@
-# Your discussion here
+[(u'aerst', 11, [u'arets', u'aster', u'rates', u'reast', u'resat', u'stare', u'stear', u'strae', u'tares', u'tears', u'teras'])]
```
jacobbaron: Looks good!

HW1/P4/P4_bfs.py
```diff
@@ -0,0 +1,17 @@
...
+def bfs(character, graph): 
+    #Things to be aware of
+    #   make sure there is a way to prevent infinite loops a->b->a->b->...
+    #       mark each searched value with true or false? 
+    #           true if searched already, false if not
+    #   check list of values for character
+    #   if not found, use those characters as key and check their values
+    #       (this will branch extensively, use partitions to break up work)
+    #   if character is found, backtrack by returning the character name
+    #return a list of characters that the node was connected to
```
jacobbaron: Looks like you are missing your bfs code!

Sdeppen: Yeah, sadly did not get to finish this pset...

HW1/P2/P2.txt
```diff
@@ -1 +1,5 @@
-# Your discussion here
+For part a, I chose to simply partition in sequential order. While for part b, I shuffled the order so that the partitions would be spread out randomly. 
+
+The sequential order does a poor job of balancing out work, as seen in the histogram. The jobs are mainly done by the first few partitions, and much less is done afterwards. This is reflected in how long it takes to compute the graph.
+
+The shuffled order, on the other hand, shows a much more gaussian distribution of partitions to iterations. It also took less time to calculate. This indicates that the word was much more balanced out, however the graph is no longer ordered, so it has to regroup after each shuffle which takes more overhead.
```
jacobbaron: What do you mean by 'regroup after each shuffle?' How does the ordering affect anything other than the partitioning?

HW1/P2/P2b.py
```diff
@@ -1,3 +1,32 @@
...
 
-# Your code here
+sc = SparkContext()
+
+#create lists of the range for rows and cols  
+ilist = range(2000)
+jlist = range(2000)
+
+random.shuffle(ilist)
+random.shuffle(jlist)
```
jacobbaron: I did something similar. However, if the dataset is too large, such that it makes sense to use spark, you might not be able to do the shuffle within python before parallelizing. This is something I'm not sure about. In that case, you'd have to parallelize first and then shuffle. I did not implement it that way myself, but it's just something to think about.

nicodri: @jacobbaron is right. Why don't you do the shuffle inside Spark? For example, you can call a partitionBy with a shuffle on the index on the rdd mandelrdd. Then you won't have any issue if the data set is too large. And also that way the work is even more balanced.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,29 @@
...
+sc = SparkContext()
+
+#create lists of the range for rows and cols  
+ilist = range(2000)
+jlist = range(2000)
+
+#turn the lists into 1 rdd, K: (i,j)
+rddi = sc.parallelize(ilist,10)
+rddj = sc.parallelize(jlist,10)
+rddK = rddi.cartesian(rddj)
```
nicodri: Pay attention to the code redundancy: here you just need one index list mylist = range(2000) and one rdd, you can compute the cartesian product with itself.

HW1/P3/P3.py
```diff
@@ -0,0 +1,26 @@
...
+words = wordsfile.map(lambda w: ("".join(sorted(w)), w))
+
+#create a rdd of the sorted words, number of anagrams, and list of anagrams
+anagrams = words.groupByKey().map(lambda x : (x[0], len(list(x[1])), list(x[1])))
+
+#sorts the anagram list by number of anagrams in descending order and then takes the first value
+#aka, the largest number of anagrams 
+max = anagrams.sortBy((lambda x: x[1]), ascending=False).take(1)
+print max
+
```
nicodri: Nice job. We have the same result. One tiny detail : you may want to look for the Spark built in function rdd.takeOrdered() to execute in native Spark your two last lines of code.

PR #277, opened 2015-10-19
Hw1

xidexia:
hw1


HW1/P2/P2.txt
```diff
@@ -1 +1,7 @@
 # Your discussion here
+
+From the P2a_hist plot, we can see that most partitions are centered at about 1e6 which a few partitions take much larger amount of time. This means partitions take largely different amount of time - over 80% partitions are just doing less than 10% of the work. 
+
```
jaco8: Very good and same idea here! In P2a, we used default partitioning. Just partition all the data into 100 partitions within fixed order, which is fast with the partition time. However, it does not balance the work, since some pixels are concentrated in some certain partitions, such as the light pixels are located in certain partitions, which cause the processing time for those partitions very slow.

HW1/P2/P2a.py
```diff
@@ -1,3 +1,18 @@
...
+import pyspark
+from pyspark import SparkContext, SparkConf
+import random
+sc = SparkContext()
+
+xx = yy = range(2000)
+
+rdd_xx = sc.parallelize(xx, 10)
+rdd_yy = sc.parallelize(yy, 10)
+
```
jaco8: Very good and clean code! Easy to follow.

HW1/P3/P3.py
```diff
@@ -0,0 +1,16 @@
+import pyspark
+import numpy as np
+from pyspark import SparkContext
+sc = SparkContext()
+
+#read data and sort them
+words = sc.textFile('EOWL_words.txt').map(lambda x: ("".join(np.sort(list(x))),x)).groupByKey().mapValues(list)
```
jaco8: np.sort() time complexity is O(nlogn), I think you can use counting sort or radix sort to make the time complexity to be linear.

HW1/P4/P4.py
```diff
@@ -0,0 +1,39 @@
...
+def create_graph(table):
+	t1 = table.map(lambda x: (x[1][0],x[1][1]))
+	# remove the ones whose neighbors are themselves
+	graph = t1.distinct().partitionBy(8).cache()
+	return graph
+
+# read data
+data = sc.textFile('source.csv')
+temp = get_hero_and_issues(data)
+table = make_table(temp[0],temp[1])
```
jaco8: Really good code.

PR #283, opened 2015-10-29
HW2

twkillian:
This assignment was a lot more straightforward than HW1. Difficult, but I learned a lot. I'm grateful for the exposure to all of the different mechanisms we can use to run code in parallel. It's too bad that this assignment seemed to bounce between the different methods from problem to problem, leaving us unable to really develop familiarity with the method as we received with Spark from the last HW. I guess that P5 was a sort of summary of different methods, which was nice. It was also interesting to see how an ensemble of different parallel mechanisms can combine to improve a highly parallelizable, large problem.


HW2/P4/driver.py
```diff
@@ -14,19 +14,54 @@
...
     tmpB = np.empty_like(tmpA)
-
+    Threads = []
+    Events = [] #np.empty([num_threads,iterations])
+    
+    # Create 2D array of events, each event corresponds to a single (thread, iteration) pair
+    for i in range(iterations):
+        for n in range(num_threads):
+            Events += [threading.Event()]
+    Events = np.array(Events).reshape((num_threads,iterations))
```
amauboussin, line 30: You could also write this as
`events = [ [threading.Event() for n in range(num_threads)] for i in range(iterations) ]`,
it doesn't really matter but there probably isn't too much advantage of using np arrays with non-numeric types.

HW2/P3/mandelbrot.pyx
```diff
@@ -32,16 +33,74 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+cdef void counts_to_output(AVX.float8 counts,
+                      np.uint32_t[:, :] out_counts,
+                      int i, int j) nogil:
+    cdef:
+        float tmp_counts[8]
+        int ii
+        
+    AVX.to_mem(counts, &(tmp_counts[0]))
+    for ii in range(8):
+        out_counts[i,j+ii] = <np.uint32_t> tmp_counts[ii]
```
amauboussin, line 105: Great job, this looks concise + readable

HW2/P5/driver.py
```diff
@@ -80,3 +96,21 @@ def randcolor():
             # SUBPROBLEM 3: sort objects by location.  Be sure to update the
             # grid if objects' indices change!  Also be sure to sort the
             # velocities with their object positions!
+            
+            # Get Positions and sort the points according to morton ordering
+            grid_positions = (positions / grid_spacing).astype(int) # Maintain grid spacing
+            zOrderCurve = map(lambda x, y: cmp_zorder(x,y), grid_positions, np.array(list(grid_positions[1:]) + list(grid_positions[:1])))
```
amauboussin, line 102: I looked at the same cmp_zorder from the wikipedia page when I was working on my own code and my understanding was that the cmp_zorder function only provided relatives values, i.e. it was a shortcut to figure out which of the two given elements passed in has a higher z-order. Your code offsets the list and compares each element to the one next to it, then sorts by this relative value. This could provide approximately the right answer but I am not sure if it completely correct because the value each element is mapped to depends on the element adjacent to it. I could be missing something though, maybe doing the comparisons in this particular way ensures correctness.

HW2/P5/physics.pyx
```diff
@@ -113,17 +137,26 @@ cpdef update(FLOAT[:, ::1] XY,
...
+                x_grid = <unsigned int>(XY[i,0] / grid_spacing)
+                y_grid = <unsigned int>(XY[i,1] / grid_spacing)
+                Grid[x_grid, y_grid] = -1
             for dim in range(2):
                 XY[i, dim] += V[i, dim] * t
+        # SUBPROBLEM 2: update the grid values.
+            if (XY[i,0]>=0 and XY[i,0]<=1) and (XY[i,1]>=0 and XY[i,1]<=1): #Check to see if ball is within grid
+                x_grid = <unsigned int>(XY[i,0] / grid_spacing)
+                y_grid = <unsigned int>(XY[i,1] / grid_spacing)
+                Grid[x_grid, y_grid] = i
```
amauboussin, line 160: Nice job, everything here looks very straightforward.

HW2/P2/parallel_vector.pyx
```diff
@@ -100,16 +107,23 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+      for r in range(repeat):
+         for idx in prange(src.shape[0], num_threads=4): 
+         
+            if counts[src[idx]] > 0:
+               
+               #Lock and alter destination
+               acquire(&(locks[dest[idx]/N]))
```
sdaulton, line 120: Really minor, but you could use a variable, say dest_lock_idx to store dest[idx]/N, since you compute this twice.   (same for src[idx]/N)

HW2/P2/parallel_vector.pyx
```diff
@@ -75,16 +75,23 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
-   # data movement.  Be sure to avoid deadlock, and double-locking.
+   # Added requisite code below
    ##########
+   
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
+
```
sdaulton, line 85: There is currently a race condition.  
Consider the following scenario:
  Suppose counts[src[idx]] equals 1.  Assume for simplicity that we only have 1 core (i.e. only one thread can be running at a time).
  Thread1 checks counts[src[idx]] > 0, which is true, and thus enters the body of the if clause.  
  Interrupt!  Thread1 stops, Thread2 runs.
  Then Thread 2 checks counts[src[idx]] > 0, which is true, and it enters the body of the if clause.
  Thread2 gets the lock for counts[dest[idx]], increments it, releases the lock.
  Thread2 gets the lock for counts[src[idx]], decrements it, and releases the lock.
  Interrupt!  Thread2 stops, Thread1 runs.
  Thread1 gets the lock for counts[dest[idx]], increments it, releases the lock.
  Thread1 gets the lock for counts[src[idx]], decrements it, and releases the lock.
  But now counts[src[idx]] equals -1!  Not good.

You can fix this by getting both locks before checking if counts[src[idx]] > 0.  Note, you need to hold both locks when you decrement dest and increment src in order to avoid the race condition.

Note: With my recommended fix, you will also need to acquire locks in order to avoid deadlock.

HW2/P2/parallel_vector.pyx
```diff
@@ -100,16 +107,23 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+               
+               #Lock and alter destination
+               acquire(&(locks[dest[idx]/N]))
+               counts[dest[idx]] += 1
+               release(&(locks[dest[idx]/N]))
+               
+               #Lock and alter source
+               acquire(&(locks[src[idx]/N]))
+               counts[src[idx]] -= 1
+               release(&(locks[src[idx]/N]))
```
sdaulton, line 127: Same note as above on the race condition.

HW2/P3/mandelbrot.pyx
sdaulton, line 105: Agreed

HW2/P4/driver.py
```diff
@@ -14,19 +14,54 @@
...
+        #Change event flag
+        Events[threadNum,i].set()
         # swap direction of filtering
         tmpA, tmpB = tmpB, tmpA
 
-    return tmpA
-
 def numpy_median(image, iterations=10):
     ''' filter using numpy '''
     for i in range(iterations):
```
sdaulton, line 67: Looks good

HW2/P5/physics.pyx
sdaulton, line 160: Looks good

HW2/P2/parallel_vector.pyx
```diff
@@ -75,16 +75,23 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
-   # Use parallel.prange() and a lock for each element of counts to parallelize
-   # data movement.  Be sure to avoid deadlock, and double-locking.
+   # Added requisite code below
    ##########
+   
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
```
oyuchoi, line 84: should acquire lock before accessing counts! because counts could be changing in the meantime!
see what @sdaulton

PR #285, opened 2015-11-01
HW2 KSheng

Kai-Square:
complete solution to HW2

### DIFF
--- a/HW2/P2/P2.txt
+++ b/HW2/P2/P2.txt
@@ -1,0 +1,41 @@
+# solution to P2
+
+For both uncorrelated or correlated problem, the serial version is very fast, much faster than any parallel version I could provide. It is because in this problem parallel version creates locks and derives overheads which take up most of the time.
+
+Comparing fine-grained and medium-grained cases, we notice medium-grained method with N = 1 is equivalent to fine-grained method in theory. (Practically some additional calculation is added, there is a difference of one second.) There is a trade-off for the number of locks - generally speaking, with more locks, we increase concurrency but also overheads. For uncorrelated problem, using less locks decrease concurrency sharply, that we observed increasing time spend with increasing N. However for correlated problem, using less locks do not decrease concurrency much if N is decent. A recommended choice of N would be 50.
+
+Please find the original output of P2.py below.
+
+Serial uncorrelated: 0.235194921494 seconds
+Fine grained uncorrelated: 4.24531388283 seconds
+medium grained, N = 1
+Medium grained uncorrelated: 5.11828589439 seconds
+medium grained, N = 2
+Medium grained uncorrelated: 5.11909413338 seconds
+medium grained, N = 5
+Medium grained uncorrelated: 5.22023510933 seconds
+medium grained, N = 10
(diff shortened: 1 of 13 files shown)


HW2/P2/P2.txt
```diff
@@ -0,0 +1,41 @@
+# solution to P2
+
+For both uncorrelated or correlated problem, the serial version is very fast, much faster than any parallel version I could provide. It is because in this problem parallel version creates locks and derives overheads which take up most of the time.
```
annawhitney, line 3: Yep, the task isn't compute-bound (it's memory-bound), so parallelizing it doesn't help because then there are just more memory accesses and those vastly outweigh the compute contribution.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,6 @@
+# solution to P5
+
+THe performance increases as we finished part 1 (parallelism), part 2 (spacial decomposition), and part 3 (sorting).
+The part 1 help us run the algorithm concurrently, and it reduce the running time. The part 2 enable us to implement an O(N) algorithm instead of the original O(N^2) algorithmm which improved the performance much.
+The part 3 make nearby balls in nearby memory, which reduce the time for the program to access the data. Here we use Morton ordering, which can be defined recursively. The property of Morton ordering allow positions that are geometrically close to each other on a 2D-grid to stay near in 1D-order, with exception of few corner cases.
+The part 4 is a errorproof work by adding locks, which will slow down the program by adding locks.
```
annawhitney, line 6: How much did Part 4 slow down the program? (And for that matter, how much did the previous parts improve performance?)

Kai-Square, line 6: Ah... it's... I finished part 5 and found I need to finish this write-up, and then it was too late to go back. Sorry for that.

HW2/P4/driver.py
```diff
@@ -27,6 +27,33 @@ def py_median_3x3(image, iterations=10, num_threads=1):
 
     return tmpA
 
+# the multithread version
+def py_median_3x3_multi(image, iterations=10, num_threads=4):
+    tmpA = image.copy()
+    tmpB = np.empty_like(tmpA)
+    
+    def py_median_3x3_single(tmpA, tmpB, idx):
+        filtering.median_3x3(tmpA, tmpB, idx, num_threads)
```
therealchuckliu, line 36: Did you make changes to the filtering file? It looks like your thread just calls this method, not sure how that works

HW2/P2/parallel_vector.pyx
```diff
@@ -54,62 +54,110 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+    # data movement.  Be sure to avoid deadlock, and double-locking.
+    ##########
+
+    with nogil:
+        for r in range(repeat):
+            # prange to parallelize
+            for idx in prange(src.shape[0], num_threads = 4):
+
+                # always lock smaller index before larger index to avoid deadlock
+                if src[idx] < dest[idx]:
```
zezhouliu: Usually, you want to do a check again after acquiring the lock, since the condition might have changed between checking the condition and acquiring the lock

PR #286, opened 2015-11-01
Hw2

dekatzenel:
HW2 complete


HW2/P2/P2.txt
```diff
@@ -0,0 +1,6 @@
...
+Serial is slightly faster uncorrelated than correlated at 0.470490217209 seconds and 0.563049077988 seconds, respectively. Though the exact difference varied, the trend was consistent over several runs.
+Fine-grained locking is faster correlated than uncorrelated at 3.96625208855 seconds and 5.60650491714 seconds, respectively. Though the exact difference varied, the trend was consistent over several runs.
+Please reference medium_hist.png for discussions of medium-grained locking.
+For all values of N, medium-grained locking shows about the same difference between correlated and uncorrelated as does fine-grained locking.
+The best value of N appears to be 1, in which case the medium-grained locking case degenerates to the fine-grained locking case. Sharing of locks generally seems to slow down the operation. However, after a certain point, as the number of locks approaches 1, it begins to resemble the serial case (with an extra lock) and begins to speed up again.
```
TuanTrain, line 6: Like your graph, I also found the value of N to have minimal effect on the speed with values of around 1 - 20 N. I do not think 1 was that much better than 10, but pretty much the same actually, in my testing however.

HW2/P2/parallel_vector.pyx
```diff
@@ -70,23 +70,46 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
+                else:
+                    acquire(&locks[src_index])
+                    locks_count = 1
+                
+                #Move value
+                if counts[src_index] > 0:
+                    counts[dest_index] += 1
+                    counts[src_index] -= 1
+                
+                #Release locks
```
TuanTrain, line 107: I had released locks in order as acquiring them but I realized that it is unnecessary since deadlocks cannot happen in the case of releasing two locks in any order.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,16 @@
...
+With instruction-level parallelism, performance improves by a factor of ~1.4 every time the number of threads doubles. In one particular run, the results were:
+1 thread: 1074.94829 Million Complex FMAs in 0.878337144852 seconds, 1223.84473468 million Complex FMAs / second
+2 threads: 1074.94829 Million Complex FMAs in 0.543131113052 seconds, 1979.16905176 million Complex FMAs / second
+4 threads: 1074.94829 Million Complex FMAs in 0.421036958694 seconds, 2553.09722294 million Complex FMAs / second
+
+Interestingly, in both cases, initial runs showed significantly slower results for 4 threads. Without instruction-level parallelism, this occurred once and was not replicable. With instruction-level parallelism, this occurred during the first 3 runs, with a specific instance of: 
+4 threads: 1074.94829 Million Complex FMAs in 2.90435004234 seconds, 370.116643768 million Complex FMAs / second
+I am unsure what would cause such behavior.
+
```
TuanTrain, line 15: I saw a noticeable difference between 1, 2, and 4 threads, and though the jump was almost 2x from 1 to 2 threads, it was about 1.8 from 2 to 4 threads for both initial and AVX cases. I did get a similar speedup of ~6 from the instruction-level parallelism.

HW2/P3/mandelbrot.pyx
```diff
@@ -31,17 +42,46 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
-
+                    # (a + bi)(a + bi) + (c + di) = a^2 + b^2*i^2 + 2abi + c + di
+                    # = a^2 - b^2 + c + i(2ab + d)
+                    z_imag = AVX.add(AVX.mul(AVX.mul(z_real, z_imag), two), c_imag)
+                    z_real = AVX.add(AVX.sub(z_real_sq, z_imag_sq), c_real)
+                    #Increment only those always and currently < 4
+                    increment = AVX.bitwise_and(less_than_4_mask, one)
+                    iters = AVX.add(iters, increment)
+                #Write to output
+                counts_to_output(iters, out_counts, i, j)
```
TuanTrain, line 84: good work!

HW2/P5/P5.txt
```diff
@@ -0,0 +1,17 @@
...
+If prange() divides the loop into contiguous sections, then some sections will take longer than others because sub_update() checks all values from the current to the end and some contiguous sections will be closer to the end than others. Since sections take varying amounts of time, speedup is less than it could be. This also explains the runtime variability.
+
+Grid (spatial decomposition)
+With 1 thread we generally get results between 400 and 600 frames per second, with one result at 255. With 4 threads we get between 25 and 630 frames per second, with high variability.
+Adding the grid is ~100x faster than multithreading alone. For n=10,000, switching from an O(n^2) algorithm to an O(n) algorithm is likely to make a significant difference even if the O(n) algorithm has high overhead, e.g. updating the grid.
+The fact that 1 thread is generally ~5x faster than 4 threads makes no sense. After speaking with the course staff and attempting to debug, no error was found in the code and they agreed it made no sense. *Shrug* multithreading is complicated and perhaps my VM is not the best with threads + complex code.
```
TuanTrain, line 8: I had a similar experience where adding more threads actually showed a slowdown in the code. I also use a VM so that may be it. I did see spikes of numbers for 4 threads that were about 2x improvements over the 1 thread, but it would happen once every 4 print-outs so averaging with the smaller values would bring it back down.

PR #293, opened 2015-11-02
HW2 re-submission

btweinstein:
My last pull request included HW0, HW1, and HW2. This pull request just has the files for HW2 as desired. This should make the homework easier to grade!


HW2/P2/P2.txt
```diff
@@ -0,0 +1,74 @@
...
+will be contained within the same lock; the other indices within the lock will fight for lock control.
+
+See "correlated_vs_N.png" for my results. Even though lock contention is likely increased for N > 10, we find a minimum
+time of roughly N ~ 35. This implies that the overhead associated with maintaining grabbing, and releasing locks is very
+large; there is a much greater penalty for having many locks than fighting to grab a few.
+
+At N larger than 35, the code again becomes slow. This is likely because at large N, less elements can be swapped at
+once, as a single lock controls a large number of indices relative to the total number of
+indices. N ~ 35 must be the sweet spot because it avoids the overhead of having too many locks
+and avoids contention due to too many locked indices at once (large N).
```
lhoangcse, line 74: nice analysis, you definitely did more than was asked!

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +53,72 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+                    acquire(&locks[destination])
+                elif case==0:
+                    acquire(&locks[source])
+                elif case==1:
+                    acquire(&locks[destination])
+                    acquire(&locks[source])
+
+                # Do the transfer
+                if counts[src[idx]] > 0:
+                    counts[destination] += 1
```
lhoangcse, line 108: I think for the fine-grained case, you can just lock destination and source separately when writing to counts. It is sufficient to guarantee that the same amount of transfers happens as in single-threaded case. (Note that it isn't exactly equivalent to the serial version logic-wise because it can double-increment or double-decrement some items, but is good enough for the correctness condition of this part). However, your solution here seems more robust anyway.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,68 @@
...
+be due to a bottleneck in the code (i.e. with an infinite number of cores, we would only see a speedup of about
+10x), or could represent the cost associated with creating and maintaining additional parallel jobs.
+
+When we add AVX and threading, the results are more unexpected. Naively, since we operate on 8 numbers at once,
+we would expect the speed with N threads relative to serial code to be roughly 8N. Instead, with
+1 thread, we speedup of about 5x, 2 threads ~ 10x, 4 threads ~18x, and 12 threads ~ 28x. So, for 1 to 4 threads,
+we see a speedup of roughly 5N instead of 8N. The reason for this imperfect
+scaling is likely that when we are operating on 8 numbers at once, we do not stop operating on them until *all*
+pixels have finished. Hence, the speedup in each set of 8 is limited by the slowest pixel, and on average we get a
+speedup of about 5x.
```
lhoangcse, line 62: Agreed, the cause is probably because of a "lagger". It'd be interesting, though, to see what happens if we apply AVX on an array whose size is a multiple of 8 and see what the maximum speed up is. I suspect not 8x either because we still have to load data into the float8 store (so the copy likely causes some non-trivial overhead).

HW2/P4/filtering.pyx
```diff
@@ -45,3 +45,19 @@ cpdef median_3x3(FLOAT[:, :] input_image,
...
+    cdef:
+        int i, j
+
+    assert input_image.shape[0] == output_image.shape[0], "median requires same size images for input and output"
+    assert input_image.shape[1] == output_image.shape[1], "median requires same size images for input and output"
+
+    with nogil:
+        i = input_row
+        for j in range(input_image.shape[1]):  # columns
+                output_image[i, j] = median9(GETPIX(input_image, i-1, j-1), GETPIX(input_image, i-1, j), GETPIX(input_image, i-1, j+1),
```
lhoangcse, line 61: would be nice to have these parameters broken into separate lines and aligned for readability

HW2/P4/P4.txt
```diff
@@ -0,0 +1,43 @@
...
+we expect that N cores will speed up our code by a factor of N. Instead, we are seeing something like N cores gives
+us a speedup of 0.5N. This inefficiency could be created by two effects:
+
+1) Our code to create parallelism comes at a high cost, i.e. python threads require a lot of computational overhead
+2) Different rows take vastly different amounts of time to complete; each thread is not appropriately load balanced
+
+2 seems unlikely as I expect the work per pixel to be roughly the same for a simple median filter. I expect that
+the inefficiency is caused by 1, as python threading is likely much slower than C threading. Instead of
+creating and controlling threads in python, we should simply parallelize the outer loop in cython with prange; I bet
+that would create much more efficient speedups.
```
lhoangcse, line 40: Yes there always seems to be some overhead with multi-threading such that performance does not increase linearly with # of threads, even for embarassingly parallel setup. I can understand that when there are significantly many threads, this is expected because we might have reached other limits such as memory bandwitdth or resource constraint etc... However, for small number of threads, why is it that 4 threads doing independent work don't get us at least a factor of 3 boost in perf? @thouis can you comment?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+
+>>Spatially Coherent Sorting
+
+I use Morton order to sort the positions by utilizing an excellent module written by the Lawrence Livermore National Lab
+(see morton.py). Reading about Morton order on Wikipedia, "Morton order maps multidimensional data
+to one dimension while preserving the locality of the data points." It basically uses a multi-dimensional fractal that
+efficiently maps data from N dimensions to one while preserving data locality in N dimensions. Our strategy is consequently to take
+the 2d position data and sort it based on Morton order; this way, points close in space are close in memory. This is
+ advantageous as points close in space frequently interact with each other; we want memory lookups between them to be
+ very fast.
```
lhoangcse, line 26: nice explanation

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+ ]
+
+This gives us a new set of logical indices for our grid. These indices tell us that the top left element (0) should be
+first in memory, then the one below that (1), and then the element to the right of the top left element (2), etc. On
+average, elements with similar logical indices are now closer together in space. A logical indexing scheme for a 2d
+matrix  in morton order can be seen in "morton_ordering.png." On average, it looks like points closer in space
+have similar colors, indicating that they will be closer in memory as expected.
+The fractal nature of the morton ordering is evident when you zoom in on the sorted matrix; it looks nearly
+identical on the large and small scale! See "morton_ordering_zoomed.png" for a zoom-in.
+
```
lhoangcse, line 55: very interesting, and enjoyable discussion on Morton ordering

HW2/P5/physics.pyx
```diff
@@ -69,17 +74,48 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+                        # Always acquire the smaller lock first
+                        acquire(&(locks[i]))
+                        acquire(&(locks[j]))
+                        collide(XY1, V1, XY2, V2)
+                        release(&(locks[i]))
+                        release(&(locks[j]))
+
+                    # give a slight impulse to help separate them
+                    for dim in range(2):
+                        V2[dim] += eps * (XY2[dim] - XY1[dim])
```
lhoangcse, line 118: why don't you lock this part (update of V2)? I didn't do this either and realized afterwards that it seems needed. Although without locking it the collision still worked well for me (maybe because this update is of minor importance since we are only giving it epsilon impulse)

lhoangcse:
overall a strong submission

HW2/P2/P2.txt
btweinstein, line 74: Thanks! I was very curious and wanted to try to figure out what was going on.

HW2/P4/P4.txt
btweinstein, line 40: Yes I'm also curious. I was hoping for a larger speedup!

HW2/P5/P5.txt
btweinstein, line 55: Thanks, I appreciate it!

HW2/P5/physics.pyx
btweinstein, line 118: Oops, I agree, I think it is probably needed. I should have locked both statements. I didn't see anything bizarre in my simulation, however, so things must have qualitatively worked without it, as you suggested!

HW2/P4/P4.txt
thouis, line 40: Cache behavior is a likely culprit.  This is not a computationally intense problem, so if you have two cores (4 with hyperthreading, perhaps), 2x is about as fast as you could expect to go.

HW2/P5/morton.py
```diff
@@ -0,0 +1,234 @@
+################################################################################
```
foo-bar-baz-qux, line 1: Interesting visualisation!

HW2/P5/physics.pyx
```diff
@@ -69,17 +74,48 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-            # give a slight impulse to help separate them
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+
+    # Get list of all objects within 2 grid squares of you
+
+    cdef int center_xgrid = <int>(XY[i, 0]/grid_spacing)
+    cdef int center_ygrid = <int>(XY[i, 1]/grid_spacing)
+
+    cdef int distance_to_check = 2
```
foo-bar-baz-qux, line 83: Would this also handle the case where you have (x, y) colliding with (x, y+2)?

HW2/P2/P2.txt
```diff
@@ -0,0 +1,74 @@
...
+In the correlated case, the source and destination indices are within 10 indices of each other. This suggests that we can
+decompose the problem into groups of 10 indices at a time. Naively, we expect lock contention to be minimized when
+each index has its own lock. If N > 10, there should always be contention for locks, as the source and destination
+will be contained within the same lock; the other indices within the lock will fight for lock control.
+
+See "correlated_vs_N.png" for my results. Even though lock contention is likely increased for N > 10, we find a minimum
+time of roughly N ~ 35. This implies that the overhead associated with maintaining grabbing, and releasing locks is very
+large; there is a much greater penalty for having many locks than fighting to grab a few.
+
+At N larger than 35, the code again becomes slow. This is likely because at large N, less elements can be swapped at
```
zezhouliu, line 71: this is an interesting result, and not something I would expect

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +53,72 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+                # the source and the destination...
+
+                source = src[idx]
+                destination = dest[idx]
+
+                # To avoid deadlock, always lock the smaller one first.
+                if source < destination: case = -1
+                elif source > destination: case=1
+                else: case=0
+
```
zezhouliu: what if source and destination changes between these lines? acquiring the lock past here may be too late, and counts[src[idx]] may have been invalidated since checking the condition (e.g., it is no longer > 0)

PR #294, opened 2015-11-02
Hw2

k-ye:


HW2/P2/parallel_vector.pyx
```diff
@@ -70,23 +70,44 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
+    # Your code here
+    # Use parallel.prange() and a lock for each element of counts to parallelize
+    # data movement.  Be sure to avoid deadlock, and double-locking.
+    ##########
+    with nogil:
+        for r in prange(repeat, num_threads=4):
+            for idx in range(src.shape[0]):
+                srcIdx, destIdx = src[idx], dest[idx]
+                
+                if srcIdx < destIdx:
```
LeonhardFS, line 88: Why are you using here the double if? Shouldn't it be sufficient to check for screed != destIdx to avoid the double locking case in the case of equality?

HW2/P2/P2.txt
```diff
@@ -0,0 +1,99 @@
...
+Medium grained correlated: 4.71206903458 seconds
+
+N = 100
+-----------------------
+Serial correlated: 0.369651794434 seconds
+Fine grained correlated: 5.9009950161 seconds
+Medium grained correlated: 5.99991798401 seconds
+
+We can see that serial version is much faster than parallel version, because of the large overhead introduced by using lock mechanism.
+As we can see, with the number of locks decreased using larger N, the amount of time spent using medium grained locks first decreased and then increased again, with N = 50 being the optimal. Since the data are correlated this time, decreasing the number of locks does not effect the concurrency too much, as usually we will only acquire one lock (instead of two under uncorrelated situation).
```
LeonhardFS, line 99: very nice explanation with good formation :)

HW2/P3/mandelbrot.pyx
```diff
@@ -9,15 +9,35 @@ from cython.parallel import prange
...
+       AVX.float8 zreal
+       AVX.float8 zimag
+       AVX.float8 creal, cimag
+       AVX.float8 zrealSqr, zimagSqr
+       AVX.float8 newZreal, newZimag
+       
+       AVX.float8 iterCount, iterMask
+       AVX.float8 magnitudes
+       int maskSign
+       # float outIterCount[8]
```
LeonhardFS: remove commented code before final commit

HW2/P4/P4.txt
```diff
@@ -0,0 +1,9 @@
+# P4 write up
+
+1 thread:   3.59872698784 seconds for 10 filter passes.
+2 threads:  1.79542899132 seconds for 10 filter passes.
+4 threads:  1.63268899918 seconds for 10 filter passes.
+
+We can see that when changing from serial version (1 thread) to 2 threads, the amount of time decreased almost half. However, it did not change too much as the number of threads increased to 4. I might need a new Mac...
```
LeonhardFS, line 7: ...or a better algorithm ;)

HW2/P4/driver.py
```diff
@@ -21,7 +21,16 @@ def py_median_3x3(image, iterations=10, num_threads=1):
     tmpB = np.empty_like(tmpA)
 
     for i in range(iterations):
-        filtering.median_3x3(tmpA, tmpB, 0, 1)
+        threadList = []
```
LeonhardFS, line 24: a better speedup could be achieved by using signals to avoid the start of 4 (new) threads each iteration as this is typically very costly

HW2/P5/driver.py
```diff
@@ -16,9 +16,22 @@
 def randcolor():
     return np.random.uniform(0.0, 0.89, (3,)) + 0.1
 
+def morton_grid(grid_size): 
+    grid = np.array([0, 1, 2, 3], dtype=np.int32).reshape((2,2))
+    size = 2
+    while size < grid_size:
```
LeonhardFS, line 22: some comments here would greatly speed up code understanding

HW2/P5/physics.pyx
```diff
@@ -69,17 +74,27 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+    XY1_gridX, XY1_gridY = int(XY[i, 0] / grid_spacing), int(XY[i, 1] / grid_spacing)
+    # for j in range(i + 1, count):
+    for grid_i in range(max(0, XY1_gridX - 3), min(Grid.shape[0] - 1, XY1_gridX + 4)):
+        for grid_j in xrange(max(0, XY1_gridY - 3), min(Grid.shape[1] - 1, XY1_gridY + 4)):
+            j = Grid[grid_i, grid_j]
+            if j <= i:
+                continue
+            acquire(locks + j)
```
LeonhardFS, line 84: reading in parallel is never a problem, you should use locks only around the blocks of code that perform writing operations because then the rest can still be parallelized.

Also, avoid pointer arithmetic in this context here. locks[j] = locks + j Pointer arithmetic is a good choice when iterating over elements but here it is simple addressing.

HW2/P2/parallel_vector.pyx
```diff
@@ -94,22 +115,43 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+    # to parallelize data movement.  Be sure to avoid deadlock, as well as
+    # double-locking.
+    ##########
+    with nogil:
+        for r in prange(repeat, num_threads=4):
+            for idx in range(src.shape[0]):
+                srcLockIdx, destLockIdx = src[idx] / N, dest[idx] / N
+
+                if srcLockIdx < destLockIdx:
+                    acquire(&(locks[srcLockIdx]))
```
draj36, line 136: Looks good, same strategy I did.

HW2/P3/mandelbrot.pyx
```diff
@@ -26,23 +46,70 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                iterCount = AVX.add(iterCount, AVX.bitwise_and(iterMask, AVX.float_to_float8(1.0)))
+                # iterCount = AVX.add(iterCount, AVX.bitwise_and(iterMask, AVX.float_to_float8(1.0)))
+                # newZ = z * z + c
+                # z = zr + zi * i
+                # c = cr + ci * i
+                # z * z = zr^2 - zi^2 + 2*zr*zi * i
+                # newZr = zr^2 - zi^2 + cr
+                # newZi = (2*zr*zi + ci) * i
+                newZreal = AVX.add(AVX.sub(zrealSqr, zimagSqr), creal)
+                newZimag = AVX.add(AVX.mul(AVX.float_to_float8(2.0), AVX.mul(zreal, zimag)), cimag)
```
draj36, line 98: Nice job succinctly multiplying by 2 and adding using AVX commands.

HW2/P4/driver.py
```diff
@@ -21,7 +21,16 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
+        for threadidx in xrange(num_threads):
+            th = threading.Thread(target=filtering.median_3x3, 
+                                    args=(tmpA, tmpB, threadidx, num_threads))
+            th.start()
+            threadList.append(th)
+
+        for th in threadList:
+            th.join()
```
draj36, line 32: join achieves synchronization, but could do other techniques to increase parallelization. in this case, thread i+1 can't start join until thread i finished 'joining'.

HW2/P5/physics.pyx
draj36, line 84: Unless I'm mistaken, should lock i and j in increasing order to avoid deadlocking.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,69 @@
...
+    686.016355904 simulation frames per second
+    640.879263383 simulation frames per second
+    621.444131654 simulation frames per second
+    658.640337619 simulation frames per second
+
+For 4 threads, the number of simulation frame is around 1100-1200 per second.
+
+    1206.46491432 simulation frames per second
+    1213.96375753 simulation frames per second
+    1197.5009211 simulation frames per second
```
draj36, line 55: Good results, but morton_grid code could use comments to improve readability.

HW2/P5/physics.pyx
k-ye, line 84: @LeonhardFS Reading in parallel of course can be a problem when you want to do read and write in atomic way. And as far as I know, there is no difference between <del>locks[j]</del> sorry I mean &locks[j] and locks + j at all.

@draj36  Yes you are right. But I checked if j <= i above, so that could save some effort to do another check here :)

LeonhardFS, line 84: @k-ye True, if you want to do reading/writing as one atomic operation. But reading is no problem if you care about the 'write' updates (i.e. make them threadsafe). Regarding the pointer arithmetic, yes there is no semantic difference but there is an important difference in writing good, readable code. Imagine you pass this code to someone who has never done pointer arithmetic or doesn't know that this is possible in C/C++. He'll better recognize the concept locks[j]. Thus, you should use the language specific features only, when they really make sense. Also in the given case, the code somehow looks like python (which does not have a pointer concept) but you are using a specific C feature.

PR #297, opened 2015-11-02
Homework 2 Submission

ppgmg:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,153 @@
...
+the dest is assigned to be near the begining of an adjacent lock section. We
+are not convinced that a number near 10 or 20, as examples, should pose any
+particular significance if the total number of elements in the counts array
+is much larger.
+
+From the graph, for this size problem, and this distance for the 
+correlated transfer case, we could choose N where the graph (green line)
+is at a minimum -- around 30 to 35 -- although any N within the range of 
+15-100 appears to outperform the version that uses fine-grained locking.
+
```
rpestourie, line 153: very thorough analysis!

HW2/P3/P3.txt
```diff
@@ -0,0 +1,64 @@
...
+a drop from 4 times performance in the four thread case from one thread
+case. This is not that unexpected since we are using static scheduling.
+According to the Cython documentation, static scheduling is "most 
+appropriate when the scheduling overhead matters and the problem can be cut
+down into equally sized chunks that are known to have approximately the 
+same runtime". This assumption may not hold true in our problem and/or
+implementation using AVX instructions.
+
+What is more interesting is that we have achieved at least 4-5 time speedups
+with the use of instruction-level parallelism over the multi-threaded version.
```
rpestourie, line 64: Any idea on why it is a 4 times speedup instead of 8?

HW2/P4/P4.txt
```diff
@@ -0,0 +1,131 @@
...
+competitive with the synchronous case, particularly when four threads were
+used. This may provide further evidence that there is clearly overhead
+associated with communication costs between threads. 
+
+	Number of threads: 1
+	3.30246901512 seconds for 10 filter passes.
+	Number of threads: 2
+	1.64469194412 seconds for 10 filter passes.
+	Number of threads: 4
+	1.44320011139 seconds for 10 filter passes.
```
rpestourie, line 131: Very good, very complete!

HW2/P5/driver.py
```diff
@@ -74,9 +124,68 @@ def randcolor():
...
+                    py = (positions[i, 1] / grid_spacing).astype(int) 
+
+                grid[px, py] = i
+
+    # performance summary
+    print("{} iterations in simulation".format(k_iters))
+    print("minimum simulation frames per second: {}".format(min_t))
+    print("maximum simulation frames per second: {}".format(max_t))
+    print("mean simulation frames per second: {}".format(mean_time/k_iters))
+
```
rpestourie, line 191: Very nice over all!

HW2/P5/physics.pyx
```diff
@@ -113,18 +218,61 @@ cpdef update(FLOAT[:, ::1] XY,
...
+                tempy = <int>(0)
+            elif XY[i, 1] > 1.0 - R:
+                tempy = <int>(1.0 / grid_spacing)
+            else:
+                tempy = <int>(XY[i, 1] / grid_spacing)    
+
+            Grid[tempx, tempy] = i
 
 def preallocate_locks(num_locks):
     cdef omp_lock_t *locks = get_N_locks(num_locks)
```
rpestourie, line 278: I did not get a performance improvement using the locks on my computer, I am also using Hilbert and not Morton sorting, but what you get looks very consistent. Good job!

HW2/P3/P3.txt
ppgmg, line 64: Thanks, Raphael, for your positive feedback!

With respect to this question, I'm not sure whether it is appropriate to even expect 8x speed up - when I look at the code that my AVX instructions replaced, it is not like we were originally working with one complex value at a time and processing them serially.  Put another way, if we had been performing one operation at a time serially, but now we were doing 8 of those same things in parallel when using AVX, then we might be able to argue we should expect an 8x amount of performance increase.  However, I think it's much harder to quantify exactly an expected performance increase here because we are effectively comparing apples to oranges, so to speak.  Complicating things even more is that the original code that the AVX instructions replaced were numpy matrix operations, and I'd think there were probably some optimizations when performing those operations, which were eliminated when we introduced the AVX instructions because we no longer have the original code.

I was honestly surprised that we even received the noticeable amount of speedup that we did get!

rpestourie, line 64: Thanks for the followup!
I also think that the reason we don't get 8x speedup is due to the compiler (being it for numpy or cython in general).

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +61,79 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+    # data movement.  Be sure to avoid deadlock, and double-locking.
+    ##########
+
+    with nogil:
+
+        # with gil:
+        #    print "entering fine grained"
+       
+        for r in range(repeat):
+            for idx in prange(src.shape[0], nogil=False, num_threads=4):
```
araftery, line 98: Why the `nogil=False` flag?

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +61,79 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+               
+                    ### critical section, increment and decrement counters
+                    if counts[src[idx]] > 0:
+                        counts[src[idx]] -= 1
+                        counts[dest[idx]] += 1
+                    ##### #####
+
+                    # remove locks in reverse order
+                    omp_unset_lock(&(locks[second_to_lock]))
+                    omp_unset_lock(&(locks[first_to_lock]))
```
araftery, line 134: Looks good! They provided `acquire` and `release` helper functions that you could have used, but `omp_set_lock` and `omp_unset_lock` work just the same.

HW2/P3/mandelbrot.pyx
```diff
@@ -16,56 +40,96 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+
+                ci = AVX.make_float8(i_coords[i, j*8+7],
+                                     i_coords[i, j*8+6],
+                                     i_coords[i, j*8+5],
+                                     i_coords[i, j*8+4],
+                                     i_coords[i, j*8+3],
+                                     i_coords[i, j*8+2],
+                                     i_coords[i, j*8+1],
+                                     i_coords[i, j*8])
+
```
araftery, line 88: You could avoid the multiplication by using the optional `step` argument to [`range`](https://docs.python.org/2/library/functions.html#range), e.g.:

```
for j in range(0, n_8cols, 8)
```

HW2/P3/mandelbrot.pyx
```diff
@@ -5,9 +5,33 @@ import numpy
...
+                                  int j, AVX.float8 vals) nogil:
+    cdef: 
+        int k
+        float tmp_counts[8]
+
+    AVX.to_mem(vals, &(tmp_counts[0]))
+    for k in range(8):
+        out_counts[i, j+k] = <int>tmp_counts[k]
+
+    return out_counts
```
araftery, line 34: Nice. Factoring this out makes the main function much cleaner.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,131 @@
...
+        for thread in threadlist:
+            thread.join()
+
+        tmpA, tmpB = tmpB, tmpA
+
+It was instructive to note that the performance without using any
+communication constructs (other than what is implicit in join) was actually 
+competitive with the synchronous case, particularly when four threads were
+used. This may provide further evidence that there is clearly overhead
+associated with communication costs between threads.
```
araftery, line 124: Good point to call out. I figured this was the case but didn't think to try to quantify it.

HW2/P5/physics.pyx
```diff
@@ -69,17 +86,96 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+    bottom = my_gloc_y - 2
+    if bottom<0:
+        bottom = 0
+    
+    top = my_gloc_y + 2
+    if top>=gridsize:
+        top = gridsize - 1 # inclusive
+
+    for j in range(left, right+1):
+        for k in range(bottom, top+1):
```
araftery, line 122: I like how you calculate the bounds before the loop, instead of doing all of the math in one or two lines. It's a little more verbose but clearer what's going on.

HW2/P3/mandelbrot.pyx
ppgmg, line 88: True, thanks!

HW2/P2/parallel_vector.pyx
ppgmg, line 98: Ha.  I wanted it to be false (given that I noticed we are in a nogil section), but I was probably too lazy at the time to check if false was the default so I just set it explicitly.

HW2/P5/physics.pyx
ppgmg, line 122: Thanks for the positive feedback!

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +61,79 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+                # lock acquisition. In our case, src[idx] and dest[idx]
+                # represent elements of our counts array; we always lock a
+                # lower index of the counts array before a higher index of the
+                # counts array, otherwise we may encounter deadlock.
+                #
+                # Note we also add a check to ensure that we do not lock (or
+                # do anything) if the src and dest indices are the same.
+
+                if src[idx]!=dest[idx]:
+
```
zezhouliu: between these two lines, src[idx] and dest[idx] may have changed!

PR #299, opened 2015-11-03
Hw2

1Reinier:
I might take a late day to optimize my answers.


HW2/P5/driver.py
```diff
@@ -0,0 +1,112 @@
...
+
+    while True:
+        with Timer() as t:
+            update(positions, velocities, grid,
+                   radius, grid_size, locks_ptr,
+                   physics_step)
+
+            # Key and sort positions list for indexing
+            keyed_list = [(i, (positions[i][0], positions[i][1])) for i in xrange(num_balls)]
+            sorted_list = sorted(keyed_list, cmp=cmp_zorder, key=lambda x: (int(x[1][0] / grid_spacing), int(x[1][1] / grid_spacing)))
```
amauboussin, line 93: I looked at the same cmp_zorder from the wikipedia page when I was working on my own code and my understanding was that the cmp_zorder function only provided relatives values, i.e. it was a shortcut to figure out which of the two given elements passed in has a higher z-order. Your code offsets the list and compares each element to the one next to it, then sorts by this relative value. This could provide approximately the right answer but I am not sure if it completely correct because the value each element is mapped to depends on the element adjacent to it. I could be missing something though, maybe doing the comparisons in this particular way ensures correctness. (I posted the same comment on the other code review I did as well)

HW2/P5/physics.pyx
```diff
@@ -0,0 +1,167 @@
...
+
+        # update positions
+        #
+        # SUBPROBLEM 1: parallelize this loop over 4 threads (with static
+        #    scheduling).
+        # SUBPROBLEM 2: update the grid values.
+        for i in prange(count, num_threads=4, schedule="static", chunksize=count/4):
+            x = 0
+            y = 0
+            for dim in range(2):
```
amauboussin, line 155: It is a little confusing to have a loop over the dimensions and then condition on each one, I would either get rid of the condition and add the updated values to a array of length dim or just update x and y separately.

HW2/P4/driver.py
```diff
@@ -0,0 +1,96 @@
...
+    """Repeatedly filter with a 3x3 median"""
+    tmpA = image.copy()
+    tmpB = np.empty_like(tmpA)
+    for i in xrange(iterations):
+        threads = []
+        # Fire up some threads
+        for n in xrange(num_threads):
+            thread = Thread(target=filtering.median_3x3, args=(tmpA, tmpB, n, num_threads))
+            thread.start()
+            threads += [thread]
```
amauboussin, line 31: `threads.append(thread)`

HW2/P4/driver.py
```diff
@@ -0,0 +1,96 @@
...
+        for t in threads:
+            t.join()  # wait until done and kill threads
+        # swap direction of filtering
+        tmpA, tmpB = tmpB, tmpA
+    return tmpA
```
amauboussin, line 36: Nice job, very succinct implementation.

HW2/P3/mandelbrot.pyx
```diff
@@ -0,0 +1,146 @@
...
+    int refcount = 0
+
+@cython.boundscheck(False)
+@cython.wraparound(False)
+cdef void write_out(float[8] counts_array, int out_i, int out_j, np.uint32_t [:, :] out_counts) nogil:
+    global refcount
+    while refcount != 0:
+        pass
+    refcount = refcount + 1
+    out_counts[out_i, out_j] = <np.uint32_t> counts_array[0]
```
amauboussin, line 105: I imagine you tried, but is it possible to use a for loop here?

HW2/P2/P2.txt
```diff
@@ -0,0 +1,15 @@
...
+implementation, the serial execution time is under 1 second, while fine-grained completes within 14 seconds, and
+medium-grained within 19 seconds. Locking apparently introduces an overhead (creating and destroying the locks), and
+in the case of medium-grained locking also introduces a sizable waiting overhead (threads waiting on locks to be
+released).
+
+OPTIMIZATION OF N FOR MEDIUM GRAINED LOCKING
+--------------------------------------------
+By choosing values of N from 1 to 2^5 I studied a whole range of N's. The execution time for these values of N is
+available in P2_experiment_N.png. What becomes apparent in this graph is that higher values of N are correlated with
+higher execution times. In both the correlated as in the uncorrelated cases, a value of N = 1 had the best
```
Kai-Square, line 14: Great discussion on overhead. And it seems I got different results for correlated part - the time goes down and up as N comes large.

HW2/P2/parallel_vector.pyx
```diff
@@ -0,0 +1,134 @@
...
+   # Use parallel.prange() and a lock for each element of counts to parallelize
+   # data movement.  Be sure to avoid deadlock, and double-locking.
+   ##########
+   with nogil:
+       for r in range(repeat):
+           for idx in prange(src.shape[0], num_threads=4):
+               if counts[src[idx]] > 0:
+
+                   # Lock dest
+                   acquire(&(locks[dest[idx]]))
```
Kai-Square, line 88: Looks nice! I locked both src[idx] and dest[idx] and needed to deal with possible deadlocks which made things more complicated. Thank you!

HW2/P4/driver.py
```diff
@@ -0,0 +1,96 @@
...
+import pylab
+import pylab as plt
+import seaborn as sns
+
+import filtering
+from timer import Timer
+from threading import Thread
+
+
+def py_median_3x3(image, iterations=10, num_threads=1):
```
Kai-Square, line 21: Nice job for this part! And clear comments.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,23 @@
+
+MULTITHREADING
```
Kai-Square, line 2: Very detailed result! Great work!
Part 3 (sorting) accelerated my algorithm... maybe it is due to different machine performances. It is serial but only happens once for hundreds of iteration. And it may accelerate the algorithm by reducing the time accessing data from memory.

HW2/P5/driver.py
Kai-Square, line 93: I quite agree with amauboussin. Assigning a value to each cell of the grids seems to be easier than defining a comparison for this problem.
amauboussin mentioned 'approximately the right answer' and I do not think it is a serious problem. It is because what we are trying to achieve here is to make the geometric neighborhood also neighborhood in memory. The methods are never correct or wrong, but only good or bad.

HW2/P3/mandelbrot.pyx
1Reinier, line 105: That is possible. How do you like my reference count locking? Couldn't get normal locking to work here.

HW2/P5/driver.py
1Reinier, line 93: In the sorting function the comparison function is continuously applied until the sort order stays stable. Compare when doing an `a` greater than `b` comparison as `a - b`. These values are indeed relative, which is the point.

HW2/P2/parallel_vector.pyx
```diff
@@ -0,0 +1,134 @@
...
+
+   ##########
+   # Your code here
+   # Use parallel.prange() and a lock for each element of counts to parallelize
+   # data movement.  Be sure to avoid deadlock, and double-locking.
+   ##########
+   with nogil:
+       for r in range(repeat):
+           for idx in prange(src.shape[0], num_threads=4):
+               if counts[src[idx]] > 0:
```
zezhouliu, line 85: what if counts[src[idx]] changes between now and acquiring the lock? You usually want to check the same condition again after acquiring the lock

PR #301, opened 2015-11-03
Finished First Version

FroxieYe:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,11 @@
...
+Serial uncorrelated: 0.243279933929 seconds
+Fine grained uncorrelated: 5.97802305222 seconds
+Medium grained uncorrelated: 7.45749115944 seconds
+
+###### Performance on correlated data ######
+Serial correlated: 0.345690011978 seconds
+Fine grained correlated: 5.96004486084 seconds
+Medium grained correlated: 7.22283792496 seconds
+
+From the two plots attached (corresponding to correlated data and uncorrelated data respectively), I noticed that generally as N increases (the number of adjacent elements falling into the same lock increases), the time it takes increases with flutuation. I think this might be due to the fact that my machine (OSX 16GB Memory, 3.1GHz i7) can perform well enough with a serial method on the amount of data this problem provides. As N increases, the extra computational time and memory increases as well, which might not be a good trade-off in my machine. The best N, from the plot, for uncorrelated data, is 1, whereas the best N for correlated data is 9.
```
jonathandlee12, line 11: Interesting plots. Given your range of N's, I got the opposite trend for correlated data (e.g., the times went down).

HW2/P2/parallel_vector.pyx
```diff
@@ -81,10 +81,16 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           #for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
+                   acquire(&locks[dest[idx]])
                    counts[dest[idx]] += 1
+                   release(&locks[dest[idx]])
+
```
jonathandlee12, line 90: You need to lock before the "if counts[src[idx]] > 0:" statement to avoid potential negative counts.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,11 @@
...
+
+# Thread = 1: 1074.656613 Million Complex FMAs in 4.01384401321 seconds, 267.737512834 million Complex FMAs / second,
+# Thread = 2: 1074.656613 Million Complex FMAs in 1.97863984108 seconds, 543.128967025 million Complex FMAs / second,
+# Thread = 4: 1074.656613 Million Complex FMAs in 1.29965496063 seconds, 826.878398923 million Complex FMAs / second
+
+avx parallel:
+
+# Thread = 1: 1074.926712 Million Complex FMAs in 0.620010852814 seconds, 1733.72241328 million Complex FMAs / second,
+# Thread = 2: 1074.905555 Million Complex FMAs in 0.306569099426 seconds, 3506.24233496 million Complex FMAs / second
+# Thread = 4: 1074.930116 Million Complex FMAs in 0.199064016342 seconds, 5399.92177266 million Complex FMAs / second
```
jonathandlee12, line 11: Nice speedup! I'm wondering why the number of FMAs is variable across your avx runs.

HW2/P3/mandelbrot.pyx
```diff
@@ -41,8 +43,57 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                    if (AVX.signs(avxmask) == 0):
+                        break
 
+                    # Update the real part and the imaginary part.
+                    avxzr, avxzi = AVX.add(AVX.fmsub(avxzr, avxzr, AVX.mul(avxzi, avxzi)), avxcr), AVX.add(AVX.fmadd(avxzr, avxzi, AVX.mul(avxzr, avxzi)), avxci)
+              
+                    avxmag = AVX.fmadd(avxzr, avxzr, AVX.mul(avxzi, avxzi))
+                    avxmask = AVX.less_than(avxmag, AVX.float_to_float8(4.0))
+                    
+                # Copy them to memory
```
jonathandlee12, line 93: I think you needed to write a helper function to avoid collisions across threads in assigning values to out_vals (or fix the memory structure, which I don't think I saw).

HW2/P4/driver.py
```diff
@@ -27,6 +27,22 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
     return tmpA
 
+def parallelize_py_median_3x3(image, iterations = 10, num_threads = 1):
+    tmpA = image.copy()
+    tmpB = np.empty_like(tmpA)
+    for i in range(iterations):
+        threads = []
+        # Parallel in the median_3x3 call by using the num_threads as step with n as the initial value (offset.)
+        for n in range(num_threads):
+            th = threading.Thread(target = filtering.median_3x3, args = (tmpA, tmpB, n, num_threads, ))
```
jonathandlee12, line 37: Okay, you locked each iteration in step. I think you could have used another communication primitive to do this.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,29 @@
...
+19.4908942205 average simulation frames per second with 1000 iterations
+
+Subproblem 2: spatial decomposition:
+882.601256482 average simulation frames per second with 1000 iterations
+
+Subproblem 3: spatial coherent sorting:
+959.56498753 average simulation frames per second with 1000 iterations
+
+Subproblem 4: Locking:
+744.779292039 average simulation frames per second with 1000 iterations
```
jonathandlee12, line 27: I agree with your trends! :)

HW2/P2/P2.txt
virgodi, line 11: I got similar results ! But someone mentioned as well that he was getting the same results with an increasing time when plotting vs N to the exception of a brutal drop when N was equal to 80 ! So i'm wondering how much plotting and testing for a discrete number of possibilities can tell us which one is the best... maybe averaging multiple experiments? Too lazy to check it myself :-)

HW2/P3/P3.txt
virgodi, line 11: Wow ! almost all your avx runs are twice as fast as mine !

HW2/P4/driver.py
virgodi, line 37: I found that using Events was very useful, like having red and green lights for threads ! It would have avoided locking everything. But my results are similar to yours !

HW2/P5/P5.txt
virgodi, line 27: I observed some more significant differences between 3 and 4 when using 4 threads.

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +113,15 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
    ##########
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           #for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
```
zezhouliu, line 118: acquiring the lock here may be too late, and counts[src[idx]] may have been invalidated since checking the condition (e.g., it is no longer > 0)

PR #302, opened 2015-11-03
Hw2 - Tibo Perol

tperol:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,35 @@
...
+
+Medium grained correlated: 4.08418488503 seconds
+
+
+
+For the uncorrelated data exchanges, the serial (similar to a coarse -grained since serial is equivalent to locking the entire array counts) is way faster than any other method. It is longer with the fine grained because, once one location is locked in one array , the other threads cannot access it if they have the same index. Therefore there are pauses during the run. The worst is the medium-grained, it becomes more frequent that one thread tries to access an index that is locked. Roughly, if we had 2 threads, the second thread would have 1/1000 to have an index that is already locked while it would have 1/100 to have a locked index for the medium-grained.
+
+
+
+For the correlated cased, because the 2 indexes (src[idx] and dest[idx]) are closer to each other, the medium-grained threads usually tries to access an index that is not already locked. Therefore it is faster.
```
annawhitney, line 35: More precisely, the medium-grained threads are likely to try to access an index that they already hold the lock for, so there's no waiting at that point and only one lock needs to be acquired.

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +55,85 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+            #--------------------------------------
+			else:
+				acquire(&(locks[src[idx]])) 
+           	# move data
+				if counts[src[idx]] > 0:
+					counts[dest[idx]] += 1
+					counts[src[idx]] -= 1
+           	# release the lock
+				release(&(locks[src[idx]]))
+
```
annawhitney, line 134: Rather than duplicating all this code, you could wrap just the locking code in your if/elseif/else construct, and have the actual data manipulation only once.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,43 @@
...
+
+1074.656613 Million Complex FMAs in 2.06194496155 seconds, 521.185886646 million Complex FMAs / second
+
+4 threads:
+
+1074.656613 Million Complex FMAs in 1.10520887375 seconds, 972.356120662 million Complex FMAs / second
+
+Because this problem is embarassingly parallel, multitheading reduces the run time: 2 threads divides the run time roughly by 2, 4 threads divide the run time roughly by 4.
+
+A strange thing happens with prange when using 1 thread. The code is slower with 1 thread and prange than when using  a simple range without parallelizing.
```
annawhitney, line 19: Ray mentioned that this could be because when you use prange, even with only one thread, the compiler gets confused and doesn't make certain optimizations because it's not sure that they're thread-safe.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,43 @@
...
+
+2 threads: 
+
+1074.949755 Million Complex FMAs in 0.583459854126 seconds, 1842.3714115 million Complex FMAs / second
+
+4 threads:
+
+1074.949755 Million Complex FMAs in 0.296026945114 seconds, 3631.25645399 million Complex FMAs / second
+
+It looks like AVX gives us a speeding factor of roughly four when parallelizing the columns. It is not 8 because we actually do more computations (1074.949755 Million Complex FMAs instead of 1074.656613 Million Complex FMAs) using only prange. This is because in the previous version we used break once one value of z has a magnitude higher than 4 while here, with AVX, we use break when 8 adjacent values have a magnitude higher than 4.
```
annawhitney, line 39: This doesn't fully account for why it would be around 4x speedup rather than something closer to 8x speedup - the difference in the number of complex FMAs is on the order of one ten-thousandth the total number of complex FMAs, so that wouldn't cause the difference between 4x and 8x speedup.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,40 @@
...
+SUBPROBLEM 1: Multithreading
+---------------------------------------
+
+For the serial case the average number of simulation frames per second is 13.
+
+When using 4 threads with prange a chunksize of count/100 = 1e3 the average number of simulation frames per second is 50.
+
+We can play with the chunksize. Using a chunksize of count/4 leads to an average number of simulation frames per second of 30.
+
+Since we are using a static scheduler because the chunks have roughly the same runtime it is interesting to try to optimize the chunksize. Note that, when we don't prescribe the chunksize the arerage number of simulation frames per second is 30. It looks like, because we prescribe 4 threads, the automatic chunksize is count/4.
```
annawhitney, line 12: That's an interesting benchmark that I didn't try. Good to know.

HW2/P5/physics.pyx
```diff
@@ -69,17 +81,45 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+            # don't check when there is no ball (j = int_max not -1 like said in the driver)
+            if j>i and j != int_max :
+                XY2 = &(XY[j, 0])
+                V2 = &(V[j, 0])
+                if overlapping(XY1, XY2, R):
+                    # SUBPROBLEM 4: Add locking
+                    # to avoir deadlock, we access them in the same order
+                    # here it is smaller first
+                    if not moving_apart(XY1, V1, XY2, V2):
+                        acquire(&(locks[i]))
```
annawhitney: You should acquire the first lock before you get the values of XY1 & V1 and the second lock before you get the values of XY2 & V2, otherwise they might have changed between when you read the values and when you do the collision. I think this is why you're not seeing much of a slowdown with locking while most everyone else is - you're not locking for long enough to actually ensure accuracy.

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +55,85 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+	for r in range(repeat):
+
+		for idx in prange(src.shape[0], nogil = True , num_threads = 4 ):
+
+
+			#--------------------------------------
+			# need to separate to avoid deadlocking
+			# ex: src[idx] = 5 and dest[idx] = 6 on thread 1
+			# while src[idx] = 6 and dest[idx] = 5 on thread 2
+			#--------------------------------------
```
jonahkall, line 96: nice! some people didn't realize this, but this is a critical detail.

HW2/P2/parallel_vector.pyx
```diff
@@ -105,11 +154,57 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+
+			else:
+
+				acquire( &locks[lock_src])
+
+				if counts[src[idx]] > 0:
+					counts[dest[idx]] += 1
+					counts[src[idx]] -= 1
+
+				release( &locks[lock_src])
```
jonahkall, line 204: you could use variables to reduce the # of lines of code here.

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +55,85 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+	assert src.size == dest.size, "Sizes of src and dest arrays must match"
+	with nogil:
+		for r in range(repeat):
+			for idx in range(src.shape[0]):
+				if counts[src[idx]] > 0:
+					counts[dest[idx]] += 1
+					counts[src[idx]] -= 1
```
jonahkall, line 67: spacing got weird here somehow, maybe worth checking out?

HW2/P2/P2.txt
```diff
@@ -0,0 +1,35 @@
...
+
+Fine grained correlated: 3.88704681396 seconds
+
+Number of locks for medium-grained: 100
+
+Medium grained correlated: 4.08418488503 seconds
+
+
+
+For the uncorrelated data exchanges, the serial (similar to a coarse -grained since serial is equivalent to locking the entire array counts) is way faster than any other method. It is longer with the fine grained because, once one location is locked in one array , the other threads cannot access it if they have the same index. Therefore there are pauses during the run. The worst is the medium-grained, it becomes more frequent that one thread tries to access an index that is locked. Roughly, if we had 2 threads, the second thread would have 1/1000 to have an index that is already locked while it would have 1/100 to have a locked index for the medium-grained.
```
jonahkall, line 31: good analysis.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,43 @@
...
+
+1074.656613 Million Complex FMAs in 4.0765440464 seconds, 263.619526925 million Complex FMAs / second
+
+2 threads:
+
+1074.656613 Million Complex FMAs in 2.06194496155 seconds, 521.185886646 million Complex FMAs / second
+
+4 threads:
+
+1074.656613 Million Complex FMAs in 1.10520887375 seconds, 972.356120662 million Complex FMAs / second
```
jonahkall, line 15: results look similar!

HW2/P3/mandelbrot.pyx
```diff
@@ -5,73 +5,192 @@ import numpy
...
+	cdef:
+		int i,m_8,k, iterations
+		AVX.float8 c_real, c_imag, iter_float8, z_mag
+		AVX.float8 mask, z_float8_real, z_float8_imag, z_float8_real_new
+		AVX.float8 four_float8, two_float8, one_float8, zero_float8
+
+	four_float8 = AVX.float_to_float8(4.0)
+	two_float8 = AVX.float_to_float8(2.0)
+	one_float8 = AVX.float_to_float8(1.0)
+	zero_float8 = AVX.float_to_float8(0.0)
```
jonahkall, line 30: i had similar lines! saves a lot of memory vs. doing it every time

HW2/P3/mandelbrot.pyx
```diff
@@ -5,73 +5,192 @@ import numpy
...
+				if not AVX.signs(AVX.less_than(z_mag,four_float8)):
+					break		
+
+				# update z: z = z^2 + Re(c)
+
+				# with z^2 = x^2 - y^2 + i 2xy where z = x+iy
+				z_float8_real_new = AVX.sub(AVX.fmadd(z_float8_real, z_float8_real, c_real),
+													 AVX.mul(z_float8_imag, z_float8_imag))
+
+
```
jonahkall, line 81: maybe reduce the number of spaces?

HW2/P3/mandelbrot.pyx
```diff
@@ -5,73 +5,192 @@ import numpy
...
+				mask = AVX.bitwise_and(mask, one_float8)
+
+
+				# update iter_float8: add +1 when |z|^2 < 4 and +0 when |z|^2 > 4
+				iter_float8 = AVX.add(iter_float8,mask)
+
+				
+				# Assign the iterations to out_counts
+				for k in range(8):
+					out_counts[i,m_8*8+k]= <np.uint32_t> (<np.float32_t *> &iter_float8)[k]
```
jonahkall, line 100: nice!

HW2/P4/driver.py
```diff
@@ -26,6 +39,126 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
+            # create the threads, pass the function and the inputs
+            t = threading.Thread(target=worker, 
+            args=(tmpA, tmpB, iterations, thread_id, num_threads))
+            threads.append(t)
+        # all threads start working now
+        print 'threads start working now'
+        map(lambda t:t.start(), threads)
+        # join all the threads at each iteration
+        print 'now joining all the threads'
+        map(lambda t: t.join(),threads)
```
jonahkall, line 80: i like the maps here!

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +55,85 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+
+		for idx in prange(src.shape[0], nogil = True , num_threads = 4 ):
+
+
+			#--------------------------------------
+			# need to separate to avoid deadlocking
+			# ex: src[idx] = 5 and dest[idx] = 6 on thread 1
+			# while src[idx] = 6 and dest[idx] = 5 on thread 2
+			#--------------------------------------
+			if src[idx] < dest[idx]:
```
zezhouliu: what if src[idx] or dest[idx] changes between now and acquiring the lock? You usually want to check the same condition again after acquiring the lock

PR #303, opened 2015-11-03
finished

xchang4:


HW2/P2/P2
```diff
@@ -0,0 +1,16 @@
...
+
+Serial correlated: 0.674780130386 seconds
+Fine grained correlated: 4.54506778717 seconds
+Medium grained correlated: 6.94881796837 seconds
+
+No matter for uncorrelated or correlated data, serial implementation are the fastest among the three and Fine
+grained implementation is faster than medium grained. Medium grained implementation does better when the data
+is correlated.
+
+From this plot we could see that the best N to choose for medium-grained locking in the correlated case is 20.
```
LeonhardFS, line 15: Labelling the plot would be great! I did not really understand which algorithm it showed until reading...

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +113,20 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
    ##########
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+                   if src[idx]/N == dest[idx]/N:
```
LeonhardFS, line 118: src[idX]/N == dest [idx]/N  <=> src[idX] == dest [idx](the N for comparison is not necessary)

HW2/P3/mandelbrot.pyx
```diff
@@ -42,7 +42,88 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+    for k in range(8):
+        out_counts[i, j*8+k]=<unsigned int>tmp_counts[k]
+
+cpdef mandelbrot_new(np.complex64_t [:, :] in_coords,
+                     np.uint32_t [:, :] out_counts,
+                     int max_iterations=511):
+    cdef:
+       int i, j, iter, k
+       float [:, :] real, imag
+       AVX.float8 cr, ci, zr, zr_n, zi, magz2, signs, counts
```
LeonhardFS, line 65: give variables either a short explanation or a meaningful name. I.e. I guessed zr = real part of z and an 8 vector but I would be lost for cr

HW2/P4/driver.py
```diff
@@ -54,14 +104,17 @@ def numpy_median(image, iterations=10):
...
+        with Timer() as t:
+            new_image = py_median_3x3(input_image, 10, num_threads)
+        print("{0} seconds for 10 filter passes with {1} threads.".format(t.interval, num_threads))
 
     pylab.figure()
     pylab.imshow(new_image[1200:1800, 3000:3500])
     pylab.title('after - zoom')
 
-    print("{} seconds for 10 filter passes.".format(t.interval))
+    # print("{} seconds for 10 filter passes.".format(t.interval))
```
LeonhardFS, line 119: nice work :+1:

HW2/P2/parallel_vector.pyx
isadoranun, line 118: @LeonhardFS I think @xchang4 is right by doing src[idX]/N == dest [idx]/N. Since we are taking the floor of the division, that expression is not equivalent to src[idX] == dest idx.
For example if N = 5 and src[idX] = 9 , dest [idx] = 8 then  src[idX] /N =  dest [idx]/N = 1, and src[idX] = 9 !=  src[idX] = 8.

HW2/P3/P3
```diff
@@ -0,0 +1,31 @@
...
+We could see that when we used instruction-level parallelism, performance improved by a factor of about 6-7.
+
+For without instruction-level parallelism case, when increasing the number of threads, the speedup is about
+as expected. When we increased the number of threads from 1 to 2, the performance improved by a factor of
+1.8125 and when we increased the number of threads from 2 to 4, the performance improved by a factor of
+1.7.
+
+However, for instruction-level parallelism case, the speedup of the performance decreased when we increased
+the number of threads. The performance improved by a factor of 1.917, when we increased the number of thread
+from 1 to 2 while it only improved by a factor of 1.333 when we increased the number of thread from 2 to 4.
```
isadoranun, line 31: I found the same results :+1:

HW2/P3/mandelbrot.pyx
```diff
@@ -42,7 +42,88 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                zi = AVX.float_to_float8(0.0)
+                counts = AVX.float_to_float8(0.0)
+
+                for iter in range(max_iterations):
+
+                    # @x: calculate the magnitude square
+                    magz2 = AVX.add(AVX.mul(zr,zr), AVX.mul(zi,zi))
+
+                    # @x: check if magnitude square of z less than 4, return 1 if true, return 0 if false
+                    signs = AVX.bitwise_and(AVX.less_than(magz2, AVX.float_to_float8(4.0)),AVX.float_to_float8(1.0))
```
isadoranun, line 107: You should also check for magnitude square = 4 (not only less than 4).
Also, for clarity you could have created a variable for the mask, since you are using it many times (i.e. mask = AVX.less_than(magz2, AVX.float_to_float8(4.0)) )

HW2/P3/mandelbrot.pyx
```diff
@@ -42,7 +42,88 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                    zi = AVX.fmadd(AVX.mul(AVX.float_to_float8(2.0),zr),zi,ci)
+
+                    # @x: update zr value with zr_n
+                    zr = zr_n
+
 
+                counts_to_output(counts, out_counts, i, j)
 
 # An example using AVX instructions
 cpdef example_sqrt_8(np.float32_t[:] values):
```
isadoranun, line 129: Besides those comments, I agree with all the rest of the code :+1:

HW2/P4/P4
```diff
@@ -0,0 +1,13 @@
...
+
+From the result above, we could see that multi-threading speedup the process however it's less than linear speedup.
+
+I used matrix of events to coordinate the threads. Problem here is about waiting for neighbor threads to finish last
+iteration before current thread can proceed. It is a coornation problem rather than a resource sharing problem.
+So I don't need lock here (condition contains lock and event) but events. After deciding on using events, the problem
+is reduced to how to implement so that current event will be notified to proceed when the neighbors finish their
+i - 1 th iteration. A natural idea is using a matrix of events, with different rows reperesenting different iterations,
+and different columns reperesenting different iterations, so that the thread can easily check whether the neighboring
+threads has finished i - 1 th iteration or not.
```
isadoranun, line 13: Nice.

HW2/P5/physics.pyx
```diff
@@ -113,15 +139,15 @@ cpdef update(FLOAT[:, ::1] XY,
...
 
         # update positions
         #
         # SUBPROBLEM 1: parallelize this loop over 4 threads (with static
         #    scheduling).
         # SUBPROBLEM 2: update the grid values.
-        for i in range(count):
+        for i in prange(count, schedule='static', num_threads=4, chunksize=2500):
             for dim in range(2):
                 XY[i, dim] += V[i, dim] * t
```
isadoranun, line 153: :+1:

HW2/P2/parallel_vector.pyx
LeonhardFS, line 118: @isadoranun you are right, I oversaw that. Nevertheless, I would recommend to improve this code strongly by using a new temporary variable 

```
sidx = src[idx] / N # floored

#or the actual meaning that is used

sidx = <int>floor(src[idx] / N)
```

the reason for this is that it is quite difficult to get the meaning out of this code (noone looks up type definitions that are many lines away):

(1) depending on the type the result will be different. => python is not typesafe, so an actual floor operation should be not obfuscated by assuming types, even in cython code. For C/C++ code or any other typesafe language this is ok, but for languages relying on dynamic typing this is an error source.

I.e. try (which is not good code at all as it has many flaws)

```
arr = [1, 2, 3, 4, 5]

# as division is slow, a common optimization includes multiplying by the inverse!
m = 0.5 * len(arr)

arr1 = arr[:m]
arr2 = arr[m:]
```

(2) the operation is used multiple times => store it in a temporary variable, which makes the code better readable and also helps the compiler

HW2/P3/mandelbrot.pyx
LeonhardFS, line 107: @isadoranun I think checking for magnitude >= 4 or magnitude > 4 makes no difference in practice. The chances converge against zero that the case == 4 will ever be reached (thx. to floating point arithmetic), and even if so the 'error' of one false iteration count is neglectable as our coloring is done absolutely arbitrarily. I.e. the assumption that the series converges only if the result after 511 iterations is still inside an arbitrary bound is not really strict. There might be cases where we find out that the bound may be passed in the 512 iteration.

:+1: for mask

PR #310, opened 2015-11-03
Hw2

Chongmoliu:


HW2/P4/driver.py
```diff
@@ -15,13 +15,33 @@
...
     ''' repeatedly filter with a 3x3 median '''
     tmpA = image.copy()
-    tmpB = np.empty_like(tmpA)
+    tmpB = np.empty_like(tmpA) #Return a new array with the same shape and type as a given array
 
     for i in range(iterations):
-        filtering.median_3x3(tmpA, tmpB, 0, 1)
+        threads = []
+        for j in range(num_threads):
+            t=threading.Thread(target=filtering.median_3x3, args=(tmpA, tmpB, j, num_threads, ))
```
apetschek, line 38: Having an iterations loop and starting/joining the threads each time per iteration is a clever way of controlling the threads and preventing a thread from getting too far ahead in its iterations than its neighbors. In my code, i created the threads just once and then had each of them do their 10 iterations but used Events to control them. In this way, I was able to permit a thread to continue on with its j+1 iteration when threads n+1 and n-1 had completed their jth iteration, instead of having to wait for all threads to finish their jth iteration.

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +115,22 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
+
+               acquire(&locks[src[idx]/N])
+            
                if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
+
                    counts[src[idx]] -= 1
+                   if (src[idx]/N) == (dest[idx]/N):#src and dest in the same lock
```
zezhouliu, line 125: will this prevent deadlock?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,70 @@
+1. Multithreading
```
zezhouliu, line 1: good discussion

PR #313, opened 2015-11-03
Finished Hw2

chengchengqin:


HW2/P2/parallel_vector.pyx
```diff
@@ -81,10 +81,14 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
    ##########
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0],num_threads=4):
                if counts[src[idx]] > 0:
```
ghost, line 85: How are you ensuring that counts[src[idx]] > 0 continues to be met? With your locking strategy, it is possible for this to eventually be violated.
e.g. say counts[2] = 1 and 2 threads are operating on that same entry. If they both reach (and pass the condition), then they will end up setting counts[2] = -1 once they are both done.
The same comment also applies to your medium-grained locking below.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,35 @@
...
+Serial correlated: 0.448005914688 seconds
+Fine grained correlated: 4.11850094795 seconds
+
+Medium grained correlated:
+N=5, 6.68246507645 seconds
+N=10, 6.85533285141 seconds
+N=15, 7.27448797226 seconds
+N=20, 8.3435189724 seconds
+N=50, 9.59883904457 seconds
+
```
ghost, line 23: I saw the same relationship between the three approaches, i.e. serial runtime << fine grained runtime < medium-grained runtime.
I used a different locking strategy, where I locked both the source and destination to avoid any value going < 0 (see my comment below). My uncorrelated medium-grained graph looks roughly similar, but my correlated medium-grained graph has a different shape due to my locking strategy - i.e. it bottoms out at ~N=20.
I'm guessing that the "correct" approach depends on how important that <0 condition is, as your approach does ensure that multiple threads can't update the same value concurrently.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,26 @@
...
+For 2 threads:
+1074.9472 Million Complex FMAs in 0.34584403038 seconds, 3108.18492029 million Complex FMAs / second
+
+For 4 threads:
+1074.9472 Million Complex FMAs in 0.329706907272 seconds, 3260.31143507 million Complex FMAs / second
+
+
+3. Conclusion: 
+1) With instruction-level parallelism, the process is almost 9 times faster than that without instruction-level parallelism.
+2) In each case, When the number of thread increases from 1 to 2, the speed improves a lot. When the number of thread increases from 2 to 4, the improvement of speed is relatively small. The reason behind it remains further consideration. It may lie in that there is some overhead in communications between multiple threads. When the number of threads increases, costs in communicating increase as well, leading to the improvement of speed is not that obvious.
```
ghost, line 25: I wonder whether some of this may be hardware specific? I still saw a good improvement when going from 2 to 4 threads without instruction-level parallelism (2.1 to 1.1 seconds).

HW2/P3/mandelbrot.pyx
```diff
@@ -31,12 +31,66 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
-                out_counts[i, j] = iter
+                    count = AVX.add(count, AVX.bitwise_and(mask, one))
+                    # update z = z * z + c
+                    z_real_new=AVX.sub(AVX.mul(z_real,z_real),AVX.mul(z_imag,z_imag))
+                    z_imag=AVX.add(AVX.mul(z_real,z_imag),AVX.mul(z_imag,z_real))
+                    z_real=AVX.add(z_real_new,c_real)
+                    z_imag=AVX.add(z_imag,c_imag)
+                AVX.to_mem(count, &(out_counts[i,j]))
+                
+# An example using AVX instructions
```
ghost, line 66: Nice job! :+1:

HW2/P4/P4.txt
```diff
@@ -0,0 +1,15 @@
...
+
+2. Conclusion
+The codes in driver.py allow threads to begin an iteration of filtering asychronously.
+We firstly set up an array of threading.Event() with the size of (num_threads,iterations). Notice that their internal flags are initial False.
+For iteration 1, we firstly use Event.set() to set the internal flags of all n threads at iteration 1 as True. Similarly in the remaining iterations, we use Event.set() to mark that this Event has been completed. 
+Notice that for iteration i, thread n must wait until thread n-1, n, n+1 complete iteration i-1. Therefore, we still need to modify a little bit. Here, Event.wait() is an easy way to make it. 
+For the event of iteration i thread j, that is Event[j,i], using Event[j-1,i-1].wait() and Event[j+1,i-1].wait() to ensure that only the internal flags of Event[j-1,i-1] and Event[j+1,i-1] are set as True, Event[j,i] can start. Notice that there are two special cases: thread 0 and thread (# of threads -1), which only need to wait for one thread. 
+
```
ghost, line 15: I interpreted the two special cases differently - e.g. for 4 threads, the 0th thread has to wait for the 1st thread and the 4th thread (i.e. if there is no thread_id - 1, then the wait condition "wraps around" to the highest thread id; the same logic applies for the greatest thread id). This left me with a lot more if-then-else's when setting up the wait logic... did you find that you were able to pass the checks for multiple iterations, even without "wrapping around"?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,30 @@
...
+Problem 4- Locking
+With 1 thread, the speed is about 900 frames per second.
+With 4 threads, the speed is about 800 frames per second.
+With locking, the speed slows down than without locking. I think the reason is that "acquire" and "release" of the locks produce some overhead, making the speed slower.
+
+Conclusion:
+1. Although using multithreading increases the speed, the best way to improve performance is to find a better algorithm, which has a smaller time complexity.
+2. When we turn to an algorithm with a smaller time complexity, we should at the same time make sure that it will not produce relatively large overhead. Much overhead may offset the benefits of the advanced algorithm.
+3. Using multithreading not necessarily performs better than a single thread. We should take the overhead of communications between threads (and the configuration of the computer) into consideration.
+
```
ghost, line 30: That's interesting - I didn't see a 4x speed-up with 4 threads, but I did see a ~1.5x speedup (I also used Morton ordering).  In your comment for problem 3 above, what kind of communication between threads are you worried about?

HW2/P5/driver.py
```diff
@@ -80,3 +96,18 @@ def randcolor():
...
+            grid_loc = (positions/grid_spacing).astype(int)
+
+            # Morton sort and get objects' indices 
+            mortonsort = map(lambda a,b: cmp_zorder(a,b), grid_loc, np.array(list(grid_loc[1:])+list(grid_loc[:1])))
+            grid_index = np.argsort(mortonsort)
+            velocities, positions= velocities[grid_index], positions[grid_index]
+            # set old positions as -1
+            grid = - np.ones((grid_size, grid_size), dtype=np.uint32) 
+            # set new positions
+            for i in range(num_balls):
```
ghost, line 110: You can vectorize this operation, in the same way as the initialization above.  This might speed things up?

HW2/P5/physics.pyx
```diff
@@ -69,17 +73,28 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+                # SUBPROBLEM 4: Add locking
+                    if not moving_apart(XY1, V1, XY2, V2):
+                        collide(XY1, V1, XY2, V2)
+
+                    # give a slight impulse to help separate them
+                    for dim in range(2):
+                        V2[dim] += eps * (XY2[dim] - XY1[dim])
+                if i != Grid[jx, jy]:
+                    release(&locks[Grid[jx, jy]])
+    release(&locks[i])
```
ghost, line 97: Another potential source of overhead - you are holding on to locks[i] for the entire loop, while other threads might be waiting for it.  You may want to try acquiring and releasing both locks only at the exact points where you need them?

HW2/P2/parallel_vector.pyx
zezhouliu, line 85: this is a good point.  what if counts[src[idx]] changes between now and acquiring the lock? You usually want to check the same condition again after acquiring the lock

HW2/P5/P5.txt
```diff
@@ -0,0 +1,30 @@
...
+Again, the speed with sorting is much faster than the speed without sorting no matter using 1 or 4 threads.
+Unexpectedly, using 4 threads is slower than using 1 thread. I think there may be two reasons. Firstly, communications between threads produce a lot of overhead, hurting the overal performance. Secondly, I ran these codes on the virtual machine, which cannot perform well for computation-intensive tasks.
+
+Problem 4- Locking
+With 1 thread, the speed is about 900 frames per second.
+With 4 threads, the speed is about 800 frames per second.
+With locking, the speed slows down than without locking. I think the reason is that "acquire" and "release" of the locks produce some overhead, making the speed slower.
+
+Conclusion:
+1. Although using multithreading increases the speed, the best way to improve performance is to find a better algorithm, which has a smaller time complexity.
```
zezhouliu, line 27: maybe in this case, but its hard to generalize this!

zezhouliu:
@dominedo good insightful comments!

PR #315, opened 2015-11-03
HW2a

Peilin-D:
I created HW2a because I kinda messed up with HW2 branch...


HW2/P2/P2.txt
```diff
@@ -0,0 +1,24 @@
...
+Fine grained uncorrelated: 4.79588294029 seconds
+Medium grained uncorrelated: 5.32975411415 seconds
+Serial correlated: 0.272443056107 seconds
+Fine grained correlated: 4.33650183678 seconds
+Medium grained correlated: 4.09460902214 seconds 
+
+The time for medium grained locking on correlated data is expectedly less than that on uncorrelated data, because the src and dest are no more than 10 elements away, so when I lock with 10 adjacent elements, I only need to lock once for both src and dest, thus reduce the time.
+
+To find an appropriate N, I tested N in range(1,50,3), and get the image P2.png.
+In the image, the N didn't seem to reduce the time for uncorrelated data, but for correlated data, it seems N=37 would be a good choice. But note that the time is still much more than serial version.
```
vivjay30, line 13: I found that N=20 was the best due to the elements not being more than 10 apart. I wonder if doing more trials and taking the averages would give a more consistent result since N=37 seems rather arbitrary

HW2/P2/parallel_vector.pyx
```diff
@@ -0,0 +1,130 @@
...
+       int idx, r
+       omp_lock_t *locks = get_N_locks(counts.shape[0])
+
+   ##########
+   # Your code here
+   # Use parallel.prange() and a lock for each element of counts to parallelize
+   # data movement.  Be sure to avoid deadlock, and double-locking.
+   ##########
+
+   for r in prange(repeat, nogil=True, num_threads=4):
```
vivjay30: You'd want to use nogil and prange here instead of just range

HW2/P3/mandelbrot.pyx
```diff
@@ -0,0 +1,66 @@
...
+       #
+       # You may find the numpy.real() and numpy.imag() fuctions helpful.
+
+    assert in_coords.shape[1] % 8 == 0, "Input array must have 8N columns"
+    assert in_coords.shape[0] == out_counts.shape[0], "Input and output arrays must be the same size"
+    assert in_coords.shape[1] == out_counts.shape[1],  "Input and output arrays must be the same size"
+    
+    with nogil, parallel(num_threads=2):
+      # malloc space for buffer
+      local_buf=<float *> malloc(sizeof(float)*8)
```
vivjay30, line 37: Do you have to malloc here? It might be easier to just use a numpy array of the desired size where memory will be handled for you

HW2/P4/P4.txt
```diff
@@ -0,0 +1,12 @@
...
+3.14122104645 seconds for 10 filter passes
+
+2 threads:
+1.63222002983 seconds for 10 filter passes
+
+4 threads:
+1.09681010246 seconds for 10 filter passes
+
+My implementation of multithreading is simple. In one iteration, every thread filters certain rows of the image. For every iteration, I wait for all the threads to complete and then start the next iteration all at once. To do this, I used Event objects in the threading module.
+Each thread has an event, called worker_event, and they are stored in a list. And there's an main_event which is used by main thread to monitor all the threads. When a thread is done one iteration, it notifies the main thread by seting its worker_event to be ture, and also waits for the main_event. And the main thread will wait for all the worker_events. When they are all done, the main thread will set the main_event to be true and then all the threads can run the next iteration. And the main thread will clear the main_event, to prepare for the next round.
```
vivjay30, line 12: I also used events, seems to be the right choice

HW2/P5/P5.txt
```diff
@@ -0,0 +1,78 @@
...
+2000.9282561 simulation frames per second
+2022.94272603 simulation frames per second
+
+As you could see, the griding largely improves the performance.
+For serial version, the griding enables over 100x more fps, while for 4 threads version, it brings about 70x more fps.
+This is expected because the griding reduces large amount of unnecessary computations (only check collision for a constant number of neighbors), reducing the Update algorithm from O(N^2) to O(N).
+
+
+3. Spatially Coherent Sorting
+For this part, I didn't use the complex z-ordering or hilbert ordering. Rather, I simply sort the balls by their x-coordinates in each row of the grid. For example, if I have 3 balls in the first row of the grid, from left to right, I assign them number 0, 1, 2. In the second row, suppose there're 2 balls, so from left to right, I assign them number 3 and 4, and so on. And I use the resulting indices to reorder the positions array and velocities array.
```
vivjay30, line 47: This could result in a large gap at the end of the rows, especially when only considering X coordinates. Whether the tradeoff of a more complex ordering is worth the improvement in performance is yet to be determined. However a complex ordering does a better job of having spatially close balls in close proximity in memory.

PR #316, opened 2015-11-03
Hw2

haosutang:
Homework 2 is quite self-contained and helped me in thinking in parallel and I learned ways to implement multithreading, locking, events handling, SIMD, and improving algorithm.


HW2/P2/P2.txt
```diff
@@ -0,0 +1,16 @@
+(Plot: plot.png)
+
+To evaluate the performance of coarse, medium and fine grained computation time, I ran 3 simulations and took the average. The performance is plotted in 'plot.png'. For medium grained, I ran N values ranging [2,4,6,8,10,12,14].
+
+Comparing the three expriments, serial (coarse grained) yields the best performance, possibly due to its low overhead and lack of locking and unlocking. Fine grained is better than medium grained when data is uncorrelated, due to the high concurrency. For correlated medium grained, the larger N gives lower time, at N=14, it is even smaller than the fine uncorrelated data. This is probabily because of the small optimization I implemented within the lock acquire and release. In this experiment, I would choose a number N=14 (largest).
+
```
dekatzenel, line 6: What small optimization? Please be specific.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,16 @@
...
+
+------------------------
+Data for checking...:
+t_serial = [0.33290719985961914, 0.39013195037841797, 0.4000849723815918]
+t_serial_c = [0.39083003997802734, 0.408282995223999, 0.40824103355407715]
+t_fine = [6.470325946807861, 6.814625978469849, 6.4532999992370605]
+t_fine_c = [11.607882022857666, 6.2165350914001465, 6.243289947509766]
+t_medium = [[ 10.54831409,  10.09884906,   9.01622605,   8.18335915,   8.10742593,   8.54140878,   9.26269102], [  8.28127003,   8.02125406,   8.45465493,   8.43094397,   8.093467,   8.24035811,   8.12902403], [  8.02197814,  10.16742802,  10.01023197,  11.06203604,   9.48417807,   9.36058092,   9.69525409]]
+t_medium_c = [[  9.60763001,   9.427598,     7.0186758,    6.81106901,   6.8893621,   6.15112615,   5.85080194], [  7.33836699,   7.05583501,   6.75943804,   6.73647022,   6.93541193,   6.53956294,   6.60008192], [  7.48048687,   7.33036494,  11.19877601,   8.02634788,   6.30609393,   6.21858692,   5.94270992]]
```
dekatzenel, line 16: Nice job providing a summary and specific data to back it up. It would have been nice if it were more explicitly clear for the medium case which times belonged to which N.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,16 @@
+(Plot: plot.png)
+
+To evaluate the performance of coarse, medium and fine grained computation time, I ran 3 simulations and took the average. The performance is plotted in 'plot.png'. For medium grained, I ran N values ranging [2,4,6,8,10,12,14].
```
dekatzenel, line 3: Why such a small range of values? My max value was 1000, i.e. serial

HW2/P2/parallel_vector.pyx
```diff
@@ -81,10 +80,15 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
    ##########
    with nogil:
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):  
                if counts[src[idx]] > 0:
```
dekatzenel, line 84: You could have optimized the code slightly by storing src[idx] and dest[idx] in local variables rather than looking them up every time.

HW2/P2/parallel_vector.pyx
```diff
@@ -81,10 +80,15 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
+                   # Add locks and release when done
+                   acquire(&locks[dest[idx]])
                    counts[dest[idx]] += 1
+                   release(&locks[dest[idx]])
+                   acquire(&locks[src[idx]])
                    counts[src[idx]] -= 1
+                   release(&locks[src[idx]])
```
dekatzenel, line 91: This procedure definitely prevents deadlock and double-locking. However, counts[src[idx]] may change between the time you check its value and the time you decrement it because other threads can edit it in the meantime since it's not locked yet. Therefore, it is possible that this code will decrement counts[src[idx]] to a negative value, which is incorrect behavior. To prevent this, you would need to lock src[idx] before checking its count.

I suspect you should lock both src and dest prior to modifying either, but I don't have an explanation off the top of my head so I could be wrong.

HW2/P2/parallel_vector.pyx
```diff
@@ -95,7 +99,7 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
                                int repeat,
                                int N):
    cdef:
-       int idx, r
+       int idx, r, destidx, srcidx
```
dekatzenel, line 102: nit: unused variables

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +111,22 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+           for idx in prange(src.shape[0], num_threads=4):
                if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+                   # Add locks if they do not belong to the same lock, reduces locking for correlated data
+                   if dest[idx]/N==src[idx]/N:
+                     acquire(&locks[dest[idx]/N])
+                     counts[dest[idx]] += 1
+                     counts[src[idx]] -= 1
+                     release(&locks[src[idx]/N])
```
dekatzenel, line 121: Why use this optimization here and not in the fine-grained code?

HW2/P3/P3.txt
```diff
@@ -0,0 +1,25 @@
...
+1) nonAVX -> AVX_1: 706%
+2) AVX_1 -> AVX_2: 208%
+3) AVX_2 -> AVX_4: 142%
+4) AVX_4 -> AVX_8: 110%
+The increase rate, however, is decreasing. 1) is implementing the SIMD idea and I got a 7x speed increase, close to the AVX block of 8 floats. Putting more threads in, 2) is having 2 threads instead of 1, the performance enhanced by factor of 2, close to the thread number ratio. However, when using 4 threads, it has only 42% performance increase and 8 threads has only a marginal 10% increase.
+
+
+Cases:
+1. With no AVX:
+1074.656613 Million Complex FMAs in 6.03112888336 seconds, 178.184985561 million Complex FMAs / second
```
dekatzenel, line 11: What happened to checking 1, 2, and 4 threads in the non-AVX case?

HW2/P3/P3.txt
```diff
@@ -0,0 +1,25 @@
+As shown below, the performance of implementing AVX yields much better time efficiency. And the more threads, the faster.
+1) nonAVX -> AVX_1: 706%
+2) AVX_1 -> AVX_2: 208%
+3) AVX_2 -> AVX_4: 142%
+4) AVX_4 -> AVX_8: 110%
```
dekatzenel, line 5: I'm curious what machine you ran this on since you got more pronounced increases than I did. I'm using an Ubuntu VM.

HW2/P3/driver.py
```diff
@@ -22,14 +22,16 @@ def make_coords(center=(-0.575 - 0.575j),
 
     x = np.linspace(start=(-width / 2), stop=(width / 2), num=count)
     xx = center + (x + 1j * x[:, np.newaxis]).astype(np.complex64)
-    return xx, np.zeros_like(xx, dtype=np.uint32)
-
+    #return xx, np.zeros_like(xx, dtype=np.uint32)
+    return xx, np.zeros_like(xx, dtype=np.float32)
```
dekatzenel, line 26: nit: unnecessary modification

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
-       # You may find the numpy.real() and numpy.imag() fuctions helpful.
+       AVX.float8 c_r, c_i, real_part, img_part, counter_avx, mask_avx, criteria_avx, increment_avx, z_i, z_r, mask_avx2
+       float [:, :] real, img
 
     assert in_coords.shape[1] % 8 == 0, "Input array must have 8N columns"
     assert in_coords.shape[0] == out_counts.shape[0], "Input and output arrays must be the same size"
     assert in_coords.shape[1] == out_counts.shape[1],  "Input and output arrays must be the same size"
 
+    real = numpy.real(in_coords)
+    img = numpy.imag(in_coords)
```
dekatzenel, line 31: Elegant and I'm surprised this works with Cython memoryviews.

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
+
     with nogil:
-        for i in range(in_coords.shape[0]):
-            for j in range(in_coords.shape[1]):
-                c = in_coords[i, j]
-                z = 0
+        for i in prange(height, schedule='static', chunksize=1, num_threads=1):
+            for j in range(width/8):
```
dekatzenel, line 35: Alternatively, keep the range the same and modify the step size from 1 to 8.

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
+                                           real[i, j*8+6],
+                                           real[i, j*8+7])
+                c_i = AVX.make_float8(img[i, j*8],
+                                          img[i, j*8+1],
+                                          img[i, j*8+2],
+                                          img[i, j*8+3],
+                                          img[i, j*8+4],
+                                          img[i, j*8+5],
+                                          img[i, j*8+6],
+                                          img[i, j*8+7])
```
dekatzenel, line 55: This would have been a clean place to deal with the AVX order reversal since you typed out the entries anyway. Actually, I don't see you handling it before the to_mem() call. Are you writing things to memory backwards?

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
+                criteria_avx = AVX.float_to_float8(4)
+
+                # Do Mandelbrot fractal...
                 for iter in range(max_iterations):
-                    if magnitude_squared(z) > 4:
+
+                    real_part = AVX.add(AVX.sub(AVX.mul(z_r, z_r), AVX.mul(z_i, z_i)), c_r)
+                    img_part = AVX.add(AVX.mul(AVX.mul(z_r, z_i), AVX.float_to_float8(2)), c_i)
+                    mask_avx = AVX.less_than(AVX.add(AVX.mul(real_part, real_part), AVX.mul(img_part, img_part)), 
+                                            AVX.float_to_float8(4))
```
dekatzenel, line 67: Should have replaced this with criteria_avx

dekatzenel, line 67: You've squished so many operations into 3 lines of code that it becomes hard to read. Comments and/or more lines would have been useful.

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
+
+                    # Break out if all units in the avx are done calculation
+                    if AVX.signs(mask_avx)==0:
```
dekatzenel, line 70: I believe that your counts are off-by-one from the original version because you update z prior to incrementing, rather than the other way around.

HW2/P4/driver_plot.py
```diff
@@ -0,0 +1,83 @@
...
+    for n in range(1,17):
+        with Timer() as t:
+            new_image = py_median_3x3(input_image, 10, n)
+        t_record[n-1] = t.interval
+
+    print t_record
+    plt.plot(range(1,17), t_record)
+    plt.xlabel('number of threads')
+    plt.ylabel('time')
+    plt.show()
```
dekatzenel, line 82: Nice use of a plot to explain your results. You probably didn't need to include a whole separate file for the plotting code, though.

HW2/P4/driver.py
```diff
@@ -15,11 +15,66 @@
...
+            # swap direction of filtering
+            tmpA, tmpB = tmpB, tmpA
+    else:
+        # initialize the threads
+        for tid in range(num_threads):
+            t = threading.Thread(target=filterHelper, 
+                                 args=(iterations, tid, num_threads, tmpA, tmpB, events))
+            threads.append(t)
+            t.start()
+        map(lambda t: t.join(), threads)
```
dekatzenel, line 47: Nice use of map

HW2/P4/driver.py
```diff
@@ -15,11 +15,66 @@
...
+
+    return tmpA
+
+def filterHelper(iterations, tid, num_threads, tmpA, tmpB, events):
+    ''' Function that coordinates the filtering on each line of pixels'''
+    # The first iteration only sets events, no more waiting
+    filtering.median_3x3(tmpA, tmpB, tid, num_threads)
+    # swap direction of filtering
+    tmpA, tmpB = tmpB, tmpA
+    events[tid, 1].set()
```
dekatzenel, line 57: I think you mean events[tid, 0].set() because of 0-indexing

HW2/P4/driver.py
```diff
@@ -15,11 +15,66 @@
...
+
+    for i in range(1, iterations):
+        #Handle events
+        if tid>0:
+            events[tid-1, i-1].wait()
+        if tid<num_threads-1:
+            events[tid+1, i-1].wait()
```
dekatzenel, line 64: Clever implementation of the tricky condition

HW2/P4/P4.txt
```diff
@@ -0,0 +1,13 @@
+(Plot: num_threads_result.png)
+
+In this problem, the original image is divided into "stripes" where every thread can perform median filtering on one set of stripes. For example, in the case with 4 threads, thread 1 will work on 0,4,8,... and thread 2 will work on 1,5,9,... After sufficient number of iterations, the parallelized output filtered image will be the same as that done by serial numpy median code. 
+
+However, the transformation under this parallel case should use extra caution because when performing median filtering, one thread may be transforming certain line another thread is currently reading. Therefore, I used events to avoid such issue. I initialized the events matrix that will store all the event states of threads in multiple iterations. When one thread is transforming the lines, it "locks" (set their event to wait) both the line above it and the line below it. When finished processing (no collision issue any more), the event will be set.
```
dekatzenel, line 5: Your explanation forgets that the prior iterations of the adjacent threads must be complete before the current iteration of a thread. However, your code accounts for this.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
+1) Serial run yields performance of between 5 to 6 simulations per second, while parallel run yields 7-8 simulations per second. Therefore performance is better for 4 thread case, but not four times.
```
dekatzenel, line 1: Explanation?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+
+2) Spatial decomposition surprisingly yields really good performance improvement from 7 to around 300 simulations per second. This is because a large part of calculation is done comparing whether two balls are close enough to each other to induce a collision. By decomposing space, the comparison's complexity is linear O(N) instead of a quadratic O(N^2) -  a significant increase and therefore boosting up the performance.
```
dekatzenel, line 3: Serial vs parallel?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+
+3) Spatially coherent sorting improves performance from around 300 to around 360 simulations per second - a 20% increase. I thought for a while and used the following weighting grid to assign weights to each ball:
+ 1  2  9 10
+ 4  3  8 11
+ 5  6  7 12
+16 15 14 13...
+Then I googled and found a better weighting scheme - zorder curve, which is more balanced and efficient, but I think complexity is slightly higher than my scheme. The position list is then sorted according to the weights assigned.
```
dekatzenel, line 10: Unclear which scheme was used in your reported results. 

Serial vs parallel?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+
+4) After adding locks, the performance decreased from 360 to around 320 simulations per second. Locking is necessary to prevent updating simutaneously. These safety checks impede performance by acquiring the locks and wait for potential collisions to pass and then release them, and they are essential for the sound physical movement of the colliding balls system.
```
dekatzenel, line 12: Serial vs parallel?

HW2/P5/driver.py
```diff
@@ -80,3 +110,19 @@ def randcolor():
...
+
+            # reset positions and velocities
+            positions = positions[idx]
+            velocities = velocities[idx]
+            # reset grid
+            grid[:,:] = -1
+            for i in range(num_balls):
+                if positions[i,0]<1 and positions[i,1]<1:
+                    grid[(int)(positions[i,0]/grid_spacing), (int)(positions[i,1]/grid_spacing)] = i
+
```
dekatzenel, line 128: free the locks?

HW2/P5/driver.py
```diff
@@ -80,3 +110,19 @@ def randcolor():
             # SUBPROBLEM 3: sort objects by location.  Be sure to update the
             # grid if objects' indices change!  Also be sure to sort the
             # velocities with their object positions!
+
+            # Assign weights to each ball
+            loc = [((int)(ball[0]/grid_spacing), (int)(ball[1]/grid_spacing)) for ball in positions]
```
dekatzenel, line 115: duplicated code from grid part of physics.pyx - could have reused instead

HW2/P5/driver.py
```diff
@@ -80,3 +110,19 @@ def randcolor():
...
+
+            idx = np.argsort(weight)
+
+            # reset positions and velocities
+            positions = positions[idx]
+            velocities = velocities[idx]
+            # reset grid
+            grid[:,:] = -1
+            for i in range(num_balls):
+                if positions[i,0]<1 and positions[i,1]<1:
```
dekatzenel, line 126: What about checking the lower bound, i.e. > 0?

HW2/P5/driver.py
dekatzenel, line 128: Is the correct procedure to not assign a grid space to balls out of bounds or is it to put them in the closest valid grid space? (Or is either fine?)

HW2/P5/physics.pyx
```diff
@@ -1,10 +1,11 @@
-#cython: boundscheck=False, wraparound=False
+#cython: boundscheck=True, wraparound=False
```
dekatzenel, line 1: Did you leave boundscheck=True when calculating your times?

HW2/P5/physics.pyx
```diff
@@ -69,17 +72,29 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-            # SUBPROBLEM 4: Add locking
-            if not moving_apart(XY1, V1, XY2, V2):
-                collide(XY1, V1, XY2, V2)
-
-            # give a slight impulse to help separate them
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+
+    x_loc = <int>(XY[i, 0]/grid_spacing)
+    y_loc = <int>(XY[i, 1]/grid_spacing)
```
dekatzenel: What if the ball is outside the 0-1 grid?

HW2/P5/physics.pyx
```diff
@@ -69,17 +72,29 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+
+    # Check neighboring balls.
+    for x in range(x_loc-2, x_loc+2):
+        for y in range(y_loc-2, y_loc+2):
+            if x>-1 and x<Grid.shape[0] and y>-1 and y<Grid.shape[1]:
+                j = Grid[x, y]  # find colliding counterparty
+                if j>i and j<count:
```
dekatzenel, line 84: is the j<count check necessary?

dekatzenel:
Overall, strong effort

HW2/P2/P2.py
```diff
@@ -21,31 +23,40 @@
...
+            move_data_fine_grained(counts, src, dest, 100)
+        assert counts.sum() == total, "Wrong total after move_data_fine_grained"
+        print("Fine grained uncorrelated: {} seconds".format(t.interval))
+        t_fine[i] = t.interval
+
+        ########################################
+        # You should explore different values for the number of locks in the medium
+        # grained locking
+        ########################################
+        for N in [j*2 for j in range(1,8)]:
```
drishi, line 53: Yep, did similar things to you. One thing to note is that you can accomplish the for loop you have typed here in a slightly more "pythonic" way by doing for N in range(1, 8, 2) <- skips by 2.

HW2/P2/P2.txt
drishi, line 6: I think it may have been a good idea to perhaps try more values of N. I went from 2 to 30 by twos and found the optimal to be around 20-22. 
I'd also be interested in what optimization you were able to make if you decide to share.

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +111,22 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+
+                   else:
+                     acquire(&locks[dest[idx]/N])
+                     counts[dest[idx]] += 1
+                     release(&locks[dest[idx]/N])
+                     acquire(&locks[src[idx]/N])
+                     counts[src[idx]] -= 1
+                     release(&locks[src[idx]/N])
```
drishi, line 129: Looks similar-ish, do you impose an order here (i.e. check if src[idx] > dest[idx]) in order to determine the order in which you lock things? I found that necessary for mine (looked up dining philosophers problem). The last pull request I reviewed did it more similarly to yours though so this may work just fine.

HW2/P3/P3.txt
drishi, line 5: Nice, I got the same trends as you I believe, but I structured my answer to contrast the performance in number of seconds for a particular # of threads with and without instruction-level parallelism. Here's an example of a result I got (I think they should be close to yours?): 
For example: 2 Threads WITHOUT was 2.71843004227 for me, and 2 threads WITH was 0.374233961105

HW2/P3/P3.txt
```diff
@@ -0,0 +1,25 @@
...
+The increase rate, however, is decreasing. 1) is implementing the SIMD idea and I got a 7x speed increase, close to the AVX block of 8 floats. Putting more threads in, 2) is having 2 threads instead of 1, the performance enhanced by factor of 2, close to the thread number ratio. However, when using 4 threads, it has only 42% performance increase and 8 threads has only a marginal 10% increase.
+
+
```
drishi, line 8: Do you check 2 threads and 4 threads without instruction-level parallelism?

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
                         break
-                    z = z * z + c
-                out_counts[i, j] = iter
+                    increment_avx = AVX.bitwise_and(mask_avx, AVX.float_to_float8(1))
+                    counter_avx = AVX.add(counter_avx, increment_avx)
+                    z_r = real_part
+                    z_i = img_part
+                AVX.to_mem(counter_avx, &(out_counts[i, j*8]))
```
drishi, line 77: Hm did this work for you? You didn't need to make a counts_to_output function (piazza mentioned something about this)?

HW2/P3/mandelbrot.pyx
drishi, line 67: I am also guilty of writing these lines with multiple operations occurring on them :), think comments could go a long way to help with that.

HW2/P3/mandelbrot.pyx
```diff
@@ -12,35 +12,68 @@ cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
...
+                z_r = AVX.float_to_float8(0.0)
+                z_i = AVX.float_to_float8(0.0)
+                c_r = AVX.make_float8(real[i, j*8],
+                                           real[i, j*8+1],
+                                           real[i, j*8+2],
+                                           real[i, j*8+3],
+                                           real[i, j*8+4],
+                                           real[i, j*8+5],
+                                           real[i, j*8+6],
+                                           real[i, j*8+7])
```
drishi, line 47: I did [i, j+7], [i, j+6] ... and just had my inner for loop go from range(0, in_coords.shape[1], 8)

HW2/P4/driver.py
```diff
@@ -15,11 +15,66 @@
...
 def py_median_3x3(image, iterations=10, num_threads=1):
     ''' repeatedly filter with a 3x3 median '''
     tmpA = image.copy()
     tmpB = np.empty_like(tmpA)
 
+    threads = []
+
+    # Create events that signals the threads to avoid collisions
+    events = np.array([threading.Event()]*(iterations*num_threads)).reshape([num_threads, iterations])
+
```
drishi, line 34: Nice, cool to see an Events implementation for this (I did conditional variables which i think handles it related manner)

HW2/P4/P4.txt
```diff
@@ -0,0 +1,13 @@
...
+
+I plotted the performance curve from 1 thread to 16 threads 'num_threads_result.png'. The performance enhancement is apparently significant. From 1 thread to 2 threads, time decreases to one half. After this, the performance is somehow just fluctuating between 1.6 to 2.4.
+
+* Consulted Taylor on event updating: initially I let event of iteration+1 wait and resulted in blank outputs. I modified into letting event iteration-1 wait and then it worked fine.
+
+-----------------------
+Time performance list of thread number from 1 to 16.
+[3.8952291011810303, 1.9527161121368408, 1.9355039596557617, 1.8978679180145264, 1.6866109371185303, 1.7588369846343994, 1.7240731716156006, 1.8186109066009521, 2.281008005142212, 2.4645941257476807, 1.7189021110534668, 1.6698811054229736, 1.8659181594848633, 2.0020480155944824, 1.9044520854949951, 1.7266809940338135]
```
drishi, line 13: My results were similar for threads 1,2 and 4 (although my thread 4 did a little bit better relative to the other values.)

HW2/P4/driver.py
drishi, line 34: Also did conditional variables, seems like they are closely related in the handling.

HW2/P5/P5.txt
drishi, line 1: Exact same speed-ups I got. My guess for explanation is not 4x better because of wait times + overhead.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+
+Through this problem I learned to use several ways to improve the performance of simple ball collision physics simulation: multithreading, decomposition and algorithm. The speed enhancement is significant. It also makes me think sometimes it would be better to first stop and think about the problem itself: is there any way to optimize the problem itself, rather than justing parallelizing it. It'll often surprise us how good this can perform with some simple optimization like the space decomposition we have in this problem.
+
```
drishi, line 15: Would have been interesting to see your results for the various section also broken up by how many threads there were (i.e. for 1 thread and for four threads). For example, 4 threads & spatial decomposition, I got to around 500-680 fps. Was in-line with yours for the 1-thread case.

HW2/P5/physics.pyx
```diff
@@ -113,18 +128,27 @@ cpdef update(FLOAT[:, ::1] XY,
...
+            for j in range(Grid.shape[1]):
+                Grid[i, j] = -1
+        for i in prange(count, chunksize = count/4, schedule='static', num_threads=4):
             for dim in range(2):
                 XY[i, dim] += V[i, dim] * t
 
+        # update grids
+        x_loc = <int>(XY[i, 0]/grid_spacing)
+        y_loc = <int>(XY[i, 1]/grid_spacing)
+        if x_loc<Grid.shape[0] and y_loc<Grid.shape[1]:
```
drishi, line 149: I made separate functions to make sure i was in-bounds since it was used often, but this seems fine too.

HW2/P2/P2.txt
haosutang, line 6: Small optimization means adding locks if they do not belong to the same lock, which reduces locking for correlated data.

I agree a bigger range for N will be better. I then tried [2,5,10,20,50,100] and indeed 20 is fastest

HW2/P2/P2.txt
haosutang: Resolved my comment above.

HW2/P2/P2.txt
haosutang, line 3: Resolved in the comment above.

HW2/P2/parallel_vector.pyx
haosutang, line 102: Right. Was used for testing.

HW2/P2/parallel_vector.pyx
haosutang, line 84: Agree. Not sure can be noticeably faster though.

HW2/P2/parallel_vector.pyx
haosutang, line 91: In my code, modifying count only happens when lock is on - a correct behavior. I suspect if one needs to lock count for just reading it, as you suggested.

Locking both src and dest would yield the same behavior as locking them separately.

HW2/P2/parallel_vector.pyx
haosutang, line 121: For fine-grained, it is not partitioned and thus checking if dest[idx]/N==src[idx]/N will be redundant.

HW2/P3/driver.py
haosutang, line 26: it is in fact necessary for that the argument passed into the next function will have float type.

HW2/P3/P3.txt
haosutang, line 8: Indeed,1 thread -> 2 threads: 220%, 2 threads -> 4 threads: 180%.

HW2/P3/P3.txt
haosutang, line 11: See above

HW2/P3/mandelbrot.pyx
haosutang, line 67: Right. I forgot I already created a variable criteria_avx

HW2/P3/mandelbrot.pyx
haosutang, line 70: Shouldn't it be this way? When plotting, it takes log intensity and it cannot be zero

HW2/P3/mandelbrot.pyx
haosutang, line 77: Interesting, I did not call counts_to_output and it works. I guess here I copied the counter_avx to out_counts at i, j*8 directly and this does not seem to cause problem.

HW2/P4/P4.txt
haosutang, line 5: Right.

HW2/P4/driver.py
haosutang, line 57: Yes, this is a redundant line, it gets covered later in the loop.

HW2/P5/P5.txt
haosutang, line 3: I forgot to report the multi-thread case, which I note here in the comments. Comparing again 1 thread vs 4 threads. 4 threads yields a performance increase of 160%.

HW2/P5/P5.txt
haosutang, line 10: 4 thread performance is 180% more than 1 thread

HW2/P5/P5.txt
haosutang, line 12: In this case, 4 threads performance is 185% better than 1 thread.

HW2/P5/physics.pyx
haosutang, line 1: Right. No wonder it is slower...

HW2/P5/P5.txt
```diff
@@ -0,0 +1,86 @@
...
+7.14878067072 simulation frames per second
+7.26390805951 simulation frames per second
+7.0847695252 simulation frames per second
+7.80511333714 simulation frames per second
+7.61783541292 simulation frames per second
+7.2985797165 simulation frames per second
+7.00751156139 simulation frames per second
+7.69372605057 simulation frames per second
+
+3) Spatial decomposition
```
haosutang, line 46: Note: Here I left the boundary check=True. Should be faster if boundary check is turned off.

HW2/P5/driver.py
haosutang, line 128: Lock is introduced below.

HW2/P5/physics.pyx
haosutang, line 84: Maybe not. Just to be safe.

HW2/P2/parallel_vector.pyx
zezhouliu, line 84: The cpu/tlb probably already optimizes for this since it probably already remains in cache

HW2/P2/parallel_vector.pyx
zezhouliu, line 91: agree with @dekatzenel:
Usually, you want to do a check again after acquiring the lock, since the condition might have changed between checking the condition and acquiring the lock

zezhouliu:
@dekatzenel great discussion and feedback +EC in your hw
@haosutang, @drishi +EC as well for responding to feedback

PR #321, opened 2015-11-04
*Hw2* Raphael Pestourie **CS205**

rpestourie:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,61 @@
...
+With 6 Locks
+Medium grained uncorrelated: 15.6373009682 seconds
+With 7 Locks
+Medium grained uncorrelated: 16.2506999969 seconds
+With 8 Locks
+Medium grained uncorrelated: 16.4075810909 seconds
+With 9 Locks
+Medium grained uncorrelated: 16.0944931507 seconds
+With 10 Locks
+Medium grained uncorrelated: 16.2846899033 seconds
```
georgelok, line 47: If you looked beyond 10 locks for medium grained uncorrelated, I think you might have found out that there is actually an optimal N for medium grained uncorrelated (to the point it does better than fine-grained).

HW2/P2/parallel_vector.pyx
```diff
@@ -79,12 +79,22 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
-       for r in range(repeat):
-           for idx in range(src.shape[0]):
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+   for r in range(repeat):
+       for idx in prange(src.shape[0], nogil = True , num_threads = 4 ):
+           sidx, didx = src[idx], dest[idx]
+           if sidx == didx:
+               acquire(&(locks[didx]))
```
georgelok, line 86: This is semantically correct.  Computationally though, I think you can do a small optimization here?  If src and dest are equal, does this do anything? :)

HW2/P3/P3.txt
```diff
@@ -0,0 +1,25 @@
...
+With AVX with 1 thread(s)
+1059.236124 Million Complex FMAs in 3.0956158638 seconds, 342.172986121 million Complex FMAs / second
+With AVX with 2 thread(s)
+1059.236124 Million Complex FMAs in 1.60934400558 seconds, 658.178810947 million Complex FMAs / second
+With AVX with 4 thread(s)
+1059.236124 Million Complex FMAs in 1.93741607666 seconds, 546.726197207 million Complex FMAs / second
+
+
+discussion
+My computer has only one CPU, that can only simulate two threads properly.
```
georgelok, line 19: Thanks for this comment.  This make the results make more sense.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,25 @@
...
+I think that is why 4 threads is worse than 2 threads because it adds some overheard to simulate 2 more threads.
+We see a small difference in the number of FMAs.
+It might be because the program without AVX handles complex numbers differently from the AVX program.
+I tried to play with fmadd in order to be as efficient as possible. So I am happy to see that the number of FMAs is less in the AVX case.
+For 1 and 2 threads (which are the only interesting cases given my computer), AVX is about 6 times quicker than without AVX.
```
georgelok, line 24: Good. :)

HW2/P3/mandelbrot.pyx
```diff
@@ -5,73 +5,168 @@ import numpy
...
+
+
+	# parallize the rows with prange
+	for i in prange(in_coords.shape[0], nogil = True, schedule = 'static', chunksize =1,
+	 				num_threads = n_threads):
+
+		# we now go by steps of 8 within the loop because each AVX takes care of 8 values
+		for j in range(0,in_coords.shape[1],8):
+			# define c_real (in the AVX 7 to 0 format)
+			c_real = AVX.make_float8(in_coords_real[i, j+7], in_coords_real[i, j+6], in_coords_real[i, j+5],
```
georgelok, line 38: As a style thing, it might be good to put each of the arguments on it's own line, and do some indentation (this is possible in python if the statements are enclosed in parentheses.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,51 @@
+results
+
+After a few runs, my best results are:
+
+Synchronized (result from driver.py)
+With 1 threads
+9.84096980095 seconds for 10 filter passes.
+With 2 threads
+5.13529586792 seconds for 10 filter passes.
+With 4 threads
```
georgelok, line 10: Sadness about the single core :(

HW2/P4/P4.txt
```diff
@@ -0,0 +1,51 @@
...
+5.15511703491 seconds for 10 filter passes.  
+
+Does it mean the hours I spent trying figuring out the events business was complete waste?
+Well, I don't know but let me develop on that.
+My first implementation was using a list of list of events, one event per thread and one list of thread per iterations.
+I was only using wait and set.
+Then I figured out that by using clear I can get rid of the list of events for each iteration and keep only one (at least on my computer).
+I would say that there might be a case when one thread goes from the line wait to clear (because the event has just been sent).
+If all the other threads are very very fast, they could have finished one whole cycle, while our slow thread is just passing from wait to clear.
+Once our slow thread clears, it is already too late and it is lagging 1 iteration behind.
```
georgelok, line 43: Very true.

My solution (I used semaphores, which have equivalent functionality) were to have two event handlers for each line.  Then one would signal the higher line and the other would signal the lower line.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,51 @@
...
+But luckily it did not happen on my computer.
+Honestly, I don't think I understand the pointer business under the hood that makes:
+"tmpA, tmpB = tmpB, tmpA" work at a thread level where we actually only changed a few lines.
```
georgelok, line 46: Note that tmpA, tmpB are local copies for each thread, whose values are pointers to matrices.  (filtering.median_3x3(tmpA,tmpB, offset, numThreads) then update the lines which do not collide with other threads because of offset, so they're safe to perform in parallel.   Then, we update the local values of tmpA and tmpB by swapping the pointers (which don't chance since we haven't changed the location of the matrices).

Hopefully that is clears things a bit.   I'm assuming you know the definition of pointers and how memory related to them, but if you want a clarification, feel free to message. :)

HW2/P5/P5.txt
```diff
@@ -0,0 +1,50 @@
...
+I  actually have an intuition that the 2 mode shape of the 4 thread distribution might be related to memory access.
+There must be some kind of race conflicts. We see that things will get better with sorting and locks.
+
+q3 I used Hilbert curves. I am a big fan of Hilbert as a mathematician, but I did not know he also was an artist.
+The Hilbert curves are so pleasing to see, it's almost hypnotizing you.
+As for the results, we see they are slightly better when sorting. For 1 thread we are at arounf 240,
+while with 4 threads we are at 360-370, and it is already fairly one mode.
+I actually spent way more time implementing it than trying to understand why it matters.
+As a result, I see the improvement in performance, but I still don't clearly understand why it is important.
+My intuition is that the collisions will be between balls that are close in index so it will be faster.
```
georgelok, line 42: This is actually very close to the answer.

Computer have a limited amount of really really fast memory (the cache).  When we read from memory, the processor will put the data into a small cache and evicts data from it if it can't find something in the cache.  Data is usually added into the cache in chunks, and these chunks are contiguous in main memory.  By doing this spatial sorting, we deliberately try to put data that will be accessed together in a single chunk (thus, closer data points have a higher chance to be in a chunk).

HW2/P2/parallel_vector.pyx
hyunkwang, line 86: But, I think he did it correctly in terms of avoiding double-locking. I also did the same way for this part. Is there any opportunity of optimizing this part?

HW2/P2/parallel_vector.pyx
```diff
@@ -105,11 +115,19 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+               acquire(&(locks[min(didx,sidx)]))
+               acquire(&(locks[max(didx,sidx)]))
+           if counts[src[idx]] > 0:
+               counts[dest[idx]] += 1
+               counts[src[idx]] -= 1
+           if sidx == didx:
+               release(&(locks[didx]))
+           else:
+               release(&(locks[min(didx,sidx)]))
+               release(&(locks[max(didx,sidx)]))
```
hyunkwang, line 133: you did a good job with regard to prevent double-locking and dead-locking from occurring!

HW2/P3/P3.txt
hyunkwang, line 19: Wow, my performance is 4x better than yours. what laptop are you using for the homework? I used a macbook air with i7 cpu and 8GB RAM. I am just curious.

HW2/P3/mandelbrot.pyx
```diff
@@ -5,73 +5,168 @@ import numpy
...
+
+				# update iter_AVX:
+				# AVX.bitwise_and(mask,  AVX.float_to_float8(1.)) gives a one only where mask is not zero
+				iter_AVX = AVX.add(iter_AVX, AVX.bitwise_and(mask,  AVX.float_to_float8(1.)))
+
+
+			# Assign the iterations to out_counts
+			for k in range(8):
+				out_counts[i,j+k]= <np.uint32_t> (<np.float32_t *> &iter_AVX)[k]
+
```
hyunkwang, line 82: Does it work? When I used the below, i encountered the error (I didn't remember what error it was.)
for k in range(8):
-               out_counts[i,j+k]= <np.uint32_t> (<np.float32_t *> &iter_AVX)[k]
  So, in my case, I used the function that is recommended to be used on the piazza.

HW2/P4/P4.txt
hyunkwang, line 43: Good explanation! Your performance is 3x slower than mine.

HW2/P5/physics.pyx
```diff
@@ -113,17 +144,40 @@ cpdef update(FLOAT[:, ::1] XY,
...
+            new_y_grid =  max( min(new_y_grid, Grid.shape[1] - 1), 0)
+
+            #if the particule changed location
+            if (x_grid != new_x_grid) or (y_grid != new_y_grid):
+                #empty the former location
+                Grid[x_grid, y_grid] = -1
+                #populate the new one
+                Grid[new_x_grid, new_y_grid] = i
+
+
```
hyunkwang, line 180: Nice Implementation!

HW2/P3/P3.txt
rpestourie, line 19: Yeah it's an old computer.

HW2/P3/mandelbrot.pyx
rpestourie, line 82: Did you make sure you were using the pointer?
My code runs. You can try it! :)

HW2/P2/parallel_vector.pyx
```diff
@@ -105,11 +115,19 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
-       for r in range(repeat):
-           for idx in range(src.shape[0]):
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
-
-   free_N_locks(num_locks, locks)
+   for r in range(repeat):
+       for idx in prange(src.shape[0], nogil = True , num_threads = 4 ):
+           sidx, didx = src[idx]/N, dest[idx]/N
```
zezhouliu: what if the values of src[idx]/N and dest[idx]/N changes between now and acquiring the lock? You usually want to check the same condition again after acquiring the lock

HW2/P5/P5.txt
```diff
@@ -0,0 +1,50 @@
+results
```
zezhouliu, line 1: nice graphs and discussion

PR #324, opened 2015-11-04
Hw2

nithintumma:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,44 @@
...
+
+N = 80
+Medium grained uncorrelated: 10.0822679996 seconds
+Medium grained correlated: 4.63935303688 seconds
+
+N = 100
+Medium grained uncorrelated: 11.7373850346 seconds
+Medium grained correlated: 5.50020599365 seconds
+
+In every case, the serial version of the algorithm performed much faster than the parallel version with locks. This suggests that the overhead of locks is not worth it for the level of parralelism that we are achieving. In the correlated case, it appeared that the level of locking in the medium grained case had no large impact on the runtime (it slightly went down from N=5 to N=50 than went up for every incrase in N through 100). It seemed that uncorrelated scaled poorly as we increased N for medium grained locking. This suggests that we should use a smaller N, perhaps around 10, which gives us a good result in the uncorrelated case and th ecorrelated case. Fine grained locking appeared to perform similarly to the medium grained locking.
```
twkillian, line 44: I think that a useful comparison that needs to be considered is the difference between uncorrelated and correlated methods of this code. It's clear that locking doesn't necessarily speed up your code but that's mostly due to the fact that we're doing trivial memory swapping exercises. Do you think that locks would help speed up the code if we ran more compute intensive commands per thread?

HW2/P2/parallel_vector.pyx
```diff
@@ -80,12 +80,37 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
 
+               # release the locks, reverse order
+               if src_dst_comp == 0:
+                    release(&locks[src[idx]])
+               elif src_dst_comp == 1:
+                    release(&locks[dest[idx]])
+                    release(&locks[src[idx]])
+               else:
+                    release(&locks[src[idx]])
+                    release(&locks[dest[idx]])
```
twkillian, line 111: You could consolidate this large if/else block and avoid rewriting the same code by just using something along the lines of 'aquire(&(locks[min(src[idx],dest[idx])]))' etc. But other than that, this works great!

HW2/P3/P3.txt
```diff
@@ -0,0 +1,22 @@
+Experiment results:
+4 threads
+Original
+1074.656613 Million Complex FMAs in 4.06431293488 seconds, 264.412861465 million Complex FMAs / second
+AVX
+1074.694415 Million Complex FMAs in 0.191869974136 seconds, 5601.15995135 million Complex FMAs / second
+
```
twkillian, line 7: Speedy! Good job!

HW2/P3/mandelbrot.pyx
```diff
@@ -44,6 +45,89 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                  tmp_imag_z = AVX.add(AVX.mul(twos, AVX.mul(real_z, imag_z)), imag_c)
+                  real_z = tmp_real_z
+                  imag_z = tmp_imag_z
+
+                # copy the iters to out 
+                AVX.to_mem(iters, &out[0])
+                # copy out to out_counts 
+                for k in range(8):
+                  out_counts[i, j + k] = <int> out[k] 
+
```
twkillian, line 129: This is a really clean and well written solution. Good job.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,20 @@
...
+1.63110995293 seconds for 10 filter passes.
+1.6989338398 seconds for 10 filter passes.
+1.72469592094 seconds for 10 filter passes.
+
+1 thread 
+3.3609380722 seconds for 10 filter passes.
+3.37282991409 seconds for 10 filter passes.
+3.37703895569 seconds for 10 filter passes.
+
+Above we have the runtimes for each of the configurations (1, 2, and 4 threads) in 3 runs. We see that increasing from 1 to 2 threads almost achieves the optimal speedup of 2x, but further increasing from  2 to 4 threads does not have as large of an impact on the runtime (it does go down slightly). This might me because the overhead with 4 threads begins to outweigh the cost of parralelism, while the overhead with 2 threads is much less than the computation time. Note that the communication overhead will scale with the number of threads (we need more events to keep track of).
```
twkillian, line 18: This is similar to how I reasoned. One other consideration to make when you move from 2 to 4 threads is that, almost always, one of your threads is typically tasked with OS and I/O system processes. So you're not totally utilizing a full 4 threads. That said, you are right that there is added logistics overhead when you move to additional threads that can cause hang ups, especially with the event queuing method you implemented.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,55 @@
...
+
+For spatial sorting, I used the Morton ordering (or Z-order curve). This ordering is known to preserve locality of points from two dimensions down to one dimension (i.e. points close together in one-dimensional ordering will be close together in 2-dimensional originl space). Additionally, there was a seemingly efficient method to compute on the wikipedia page for the Z-order curve which used bitshifts, and required that the inputs be integers. Because we already mapped the space to a grid, we could use these integers along with this method to compute the ordering efficiently.
```
twkillian, line 55: I used Morton ordering as well. It was really nice how it conveniences your memory access as a result of this dimensionality reduction.

HW2/P5/physics.pyx
```diff
@@ -113,18 +140,29 @@ cpdef update(FLOAT[:, ::1] XY,
...
 
+            if XY[i, 0] >= 0 and XY[i, 0] < 1: 
+                if XY[i, 1] >= 0 and XY[i, 1] < 1:
+                    ix = <int>(XY[i, 0]/grid_spacing)
+                    iy = <int>(XY[i, 1]/grid_spacing)
+                    Grid[ix, iy] = i 
+
 
 def preallocate_locks(num_locks):
     cdef omp_lock_t *locks = get_N_locks(num_locks)
```
twkillian, line 168: This looks great. Good work!

HW2/P2/P2.txt
quezacot, line 44: I got the similar result.

HW2/P2/parallel_vector.pyx
quezacot, line 111: You can also just lock one, add, release, and then lock the other one, substract, release. It does not really need to lock both at the same time.

HW2/P2/parallel_vector.pyx
```diff
@@ -80,12 +80,37 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
+                  src_dst_comp = 1
+               elif src[idx] > dest[idx]:
+                  acquire(&locks[dest[idx]])
+                  acquire(&locks[src[idx]])
+                  src_dst_comp = 2
+               else:
+                # they are equal, acquire only once 
+                  acquire(&locks[src[idx]])
+                  src_dst_comp = 0
+
```
quezacot, line 97: I think you can just skip the case that adding and subtracting at the same target.

HW2/P2/parallel_vector.pyx
```diff
@@ -106,10 +132,38 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+                # release the locks in reverse order  
+                if src_dst_comp == 0:
+                    release(&locks[src_locki])
+                elif src_dst_comp == 1:
+                    release(&locks[dst_locki])
+                    release(&locks[src_locki])
+                else:
+                    release(&locks[src_locki])
+                    release(&locks[dst_locki])
    free_N_locks(num_locks, locks)
```
quezacot, line 169: Good to catch all cases!

HW2/P3/mandelbrot.pyx
quezacot, line 129: Very clear step by step.

HW2/P3/mandelbrot.pyx
```diff
@@ -44,6 +45,89 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+    assert in_coords.shape[1] % 8 == 0, "Input array must have 8N columns"
+    assert in_coords.shape[0] == out_counts.shape[0], "Input and output arrays must be the same size"
+    assert in_coords.shape[1] == out_counts.shape[1],  "Input and output arrays must be the same size"
+
+    in_co_real = np.real(in_coords)
+    in_co_imag = np.imag(in_coords)
+    # vector of fours that we use in comparisons 
+    fours = AVX.float_to_float8(4.0)
+    twos = AVX.float_to_float8(2.0)
+    ones = AVX.float_to_float8(1.0)
```
quezacot, line 72: What are ones and twos used for?

HW2/P5/physics.pyx
```diff
@@ -50,36 +52,61 @@ cdef inline void collide(FLOAT *x1, FLOAT *v1,
...
+                if overlapping(XY1, XY2, R): 
+                    if not moving_apart(XY1, V1, XY2, V2):
+                        collide(XY1, V1, XY2, V2)
+
+                # give a slight impulse to help separate them
+                for dim in range(2):
+                    V2[dim] += eps * (XY2[dim] - XY1[dim])
+                # release lock after updating 
+                release(&locks[Grid[ix, iy]])
+    release(&locks[i])
```
quezacot, line 110: Remember to consider not let the same two balls collide twice.

HW2/P5/physics.pyx
```diff
@@ -113,18 +140,29 @@ cpdef update(FLOAT[:, ::1] XY,
...
         # SUBPROBLEM 1: parallelize this loop over 4 threads (with static
         #    scheduling).
         # SUBPROBLEM 2: update the grid values.
-        for i in range(count):
+        for i in prange(count, num_threads=num_threads, schedule='static', chunksize=count/4):
+            # erase old position in grid 
+            if XY[i, 0] >= 0 and XY[i, 0] < 1: 
+                if XY[i, 1] >= 0 and XY[i, 1] < 1:
+                    Grid[<int>(XY[i, 0]/grid_spacing), <int>(XY[i, 1]/grid_spacing)] = -1  
+            # update the objects position s
```
quezacot: There might be other ball here after collision.

HW2/P5/physics.pyx
```diff
@@ -113,18 +140,29 @@ cpdef update(FLOAT[:, ::1] XY,
...
         #
         # SUBPROBLEM 1: parallelize this loop over 4 threads (with static
         #    scheduling).
         # SUBPROBLEM 2: update the grid values.
-        for i in range(count):
+        for i in prange(count, num_threads=num_threads, schedule='static', chunksize=count/4):
+            # erase old position in grid 
+            if XY[i, 0] >= 0 and XY[i, 0] < 1: 
+                if XY[i, 1] >= 0 and XY[i, 1] < 1:
+                    Grid[<int>(XY[i, 0]/grid_spacing), <int>(XY[i, 1]/grid_spacing)] = -1
```
quezacot, line 155: There might be other ball here after collision.

HW2/P2/parallel_vector.pyx
```diff
@@ -80,12 +80,37 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
    # data movement.  Be sure to avoid deadlock, and double-locking.
    ##########
    with nogil:
-       for r in range(repeat):
+       for r in prange(repeat, num_threads=4):
            for idx in range(src.shape[0]):
+               if src[idx] < dest[idx]:
```
zezhouliu: Usually, you want to do a check again after acquiring the lock, since the condition might have changed between checking the condition and acquiring the lock

HW2/P4/P4.txt
zezhouliu, line 18: good point

HW2/P5/physics.pyx
zezhouliu, line 110: good point (see the inline comment)

PR #326, opened 2015-11-04
HW2

sdaulton:


HW2/P2/P2.py
```diff
@@ -40,12 +40,12 @@
     # You should explore different values for the number of locks in the medium
     # grained locking
     ########################################
-    N = 10
-    counts[:] = orig_counts
-    with Timer() as t:
-        move_data_medium_grained(counts, src, dest, 100, N)
-    assert counts.sum() == total, "Wrong total after move_data_medium_grained"
-    print("Medium grained uncorrelated: {} seconds".format(t.interval))
+    for N in range(1,20,1):
```
sbalanovich, line 43: I don't know if this is a sufficient enough range for N, you should look into going into the 100s and 1000s for a better idea of what's going on.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,101 @@
...
+
+N=100
+Serial uncorrelated: 0.314683198929 seconds
+Fine grained uncorrelated: 5.27880001068 seconds
+Medium grained uncorrelated: 9.99726891518 seconds
+Serial correlated: 0.328505992889 seconds
+Fine grained correlated: 5.10451292992 seconds
+Medium grained correlated: 5.2350051403 seconds
+
+Medium grained uncorrelated: 6.39034605026 seconds with N = 1 -- START GOOD Range
```
sbalanovich, line 17: This seems unnecessary/hard to read and would be much better presented in a graph, I think

HW2/P2/parallel_vector.pyx
```diff
@@ -107,9 +120,23 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+               dest_lock_idx = dest[idx]/N
+               if src_lock_idx < dest_lock_idx:
+                   acquire(&(locks[src_lock_idx]))
+                   acquire(&(locks[dest_lock_idx]))
+               elif src_lock_idx > dest_lock_idx:
+                   acquire(&(locks[dest_lock_idx]))
+                   acquire(&(locks[src_lock_idx]))
+               else:
+                   #make sure we don't double lock
+                   acquire(&(locks[src_lock_idx]))
```
sbalanovich, line 135: Excellent!

HW2/P3/driver.py
```diff
@@ -22,12 +22,25 @@ def make_coords(center=(-0.575 - 0.575j),
...
+
+    x = np.linspace(start=(-width / 2), stop=(width / 2), num=count)
+    xx = center + (x + 1j * x[:, np.newaxis]).astype(np.complex64)
+    xx = np.array([xx[0]])
+    
     return xx, np.zeros_like(xx, dtype=np.uint32)
 
 
 if __name__ == '__main__':
+    #in_coords, out_counts = make_coords_test()
```
sbalanovich, line 40: Probably want to purge commented code out of final submission

HW2/P3/mandelbrot.pyx
```diff
@@ -31,16 +70,74 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+    in_imags = np.imag(in_coords)
+    
+    num_threads = 4
+     
+    # for reseting purposes
+    zeros8 = AVX.float_to_float8(0.0)
+
+    # used in comparison later
+    fours8 = AVX.float_to_float8(4.0)
+    nans8 = AVX.greater_than(fours8,zeros8)
```
sbalanovich, line 84: How is this any different than just an AVX float8 with all 0s? (As in, your zeros8 above)?

HW2/P3/mandelbrot.pyx
```diff
@@ -31,16 +70,74 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
-                c = in_coords[i, j]
-                z = 0
+        for i in prange(in_coords.shape[0],num_threads=num_threads,schedule='static',chunksize=1):
+            for j_idx in range(in_coords.shape[1]/8): # grab sets of 8 columns
+                j = j_idx * 8 # the real offset
+                
+                # get list of the complex values in the 8 columns
+                # and convert to float8
+                # there must be a better way to do this
+                reals8 = AVX.make_float8(in_reals[i,j+7], in_reals[i,j+6], in_reals[i,j+5], in_reals[i,j+4], in_reals[i,j+3], in_reals[i,j+2], in_reals[i,j+1], in_reals[i,j])
```
sbalanovich, line 95: Just a stylistic thing, this would be more readable in a column rather than this long row

HW2/P3/mandelbrot.pyx
```diff
@@ -31,16 +70,74 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                imags8 = AVX.make_float8(in_imags[i,j+7], in_imags[i,j+6], in_imags[i,j+5], in_imags[i,j+4], in_imags[i,j+3], in_imags[i,j+2], in_imags[i,j+1], in_imags[i,j])
+                
+                #reset float8 vectors for z's
+                z_reals8 = AVX.bitwise_and(zeros8, zeros8)
```
sbalanovich, line 99: This is an interesting approach. Why not just use "AVX.float_to_float8(0.0)" for both of these? Is there some perf gain you get out of using the AVX bitwise and?

HW2/P3/mandelbrot.pyx
```diff
@@ -31,16 +70,74 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                    # Note: we can quit if every value in tmp_out is > 0
+                    #AVX.greater_than(tmp_out8, zeros8) has NaN's where tmp_out has already been written to
+                    if AVX.signs(AVX.greater_than(tmp_out8, zeros8)) == 255: #2^8-1 = (binary) 11111111
+                        #write it out to out_counts
                         break
-                    z = z * z + c
-                out_counts[i, j] = iter
+                        
+                    # now do z = z * z + c
+                    new_z_reals8 = AVX.add(AVX.sub(AVX.mul(z_reals8, z_reals8), AVX.mul(z_imags8, z_imags8)), reals8)
```
sbalanovich, line 134: Consider simplifying these with fmadd and fmsub

HW2/P4/driver.py
```diff
@@ -19,6 +19,46 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
+    x = num_lines/num_threads
+    for tid in range(num_lines%num_threads):
+        lines_by_tid[tid].append(tid + num_threads * x)
+    '''  
+    #print threading.active_count()
+    for i in range(iterations):
+        threads = []
+        for tid in range(num_threads):
+            #offset = tid, step = num_threads
+            t = threading.Thread(target=filtering.median_3x3,args=(tmpA, tmpB, tid, num_threads))
```
sbalanovich, line 44: I might be missing something here, but what exactly is your locking mechanism here? How do you make sure the invariant of threads n-1 and n+1 only starting when n is complete is upheld? Does this code actually pass the assertion test in line 97 below?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,75 @@
...
+    
+    Performance with 4 threads:
+        1605.25901881 simulation frames per second
+        1616.10606007 simulation frames per second
+        1647.96407723 simulation frames per second
+        1634.40684691 simulation frames per second
+        1630.7097212 simulation frames per second
+        Average: 1626.889 simulation frames per second
+
+    On average, we achieve close to three times as many simulations per second using 4 threads instead of 1.  The addition of griding improves performance substantially from part 1.
```
sbalanovich, line 37: What gains do we get exactly? In this case it looks like it's about 50x in the 1-thread and 72x in the 4-thread case. What do you think causes this degree of perf boosts?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,75 @@
...
+    
+    Performance with 4 threads:
+        1817.21198347 simulation frames per second
+        1810.33293945 simulation frames per second
+        1821.81848218 simulation frames per second
+        1796.67335385 simulation frames per second
+        1823.8834323 simulation frames per second
+    Average: 1813.984 simulation frames per second
+    
+    Sorting improves performance over griding, with a proportionally larger increase when using 4 threads.  I chose to use Morton order rather than Hilbert for sorting; to quote wikipedia using Hilbert ordering, "points that have nearby d  values will also have nearby (x,y) values.  The converse can't always be true. There will sometimes be points where the (x,y) coordinates are close but their d values are far apart."  The whole point of sorting is to try to assign nearby (x,y) values nearby d values.  Thus, Hilbert's sorting may not be effective.
```
sbalanovich, line 56: Why do you suppose we attain a "proportionally larger increase" with 4 threads?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,75 @@
...
+    
+    Performance with 4 threads, griding and sorting and locking:
+        1405.08690725 simulation frames per second
+        1396.80359382 simulation frames per second
+        1377.08580646 simulation frames per second
+        1346.46682409 simulation frames per second
+        1461.19098775 simulation frames per second
+    Average: 1397.326 simulation frames per second
+
+    Using locking, we see performance decrease from performance without locking.  This is expected as threads will have to wait to access resources.
```
sbalanovich, line 75: So should we not use locking at all? Is there any derived benefit from introducing locks?

HW2/P5/physics.pyx
```diff
@@ -69,18 +73,44 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+                grid_x = current_grid_x + x
+                grid_y = current_grid_y + y
+                #make sure this square is in the grid
+                if ((grid_x >= 0) and (grid_x < grid_size) and (grid_y >= 0) and (grid_y < grid_size)):
+                    #j is the index of the object in Grid[grid_x,grid_y]
+                    j = Grid[grid_x,grid_y]
+                    #make sure the object is not the same as object as i, and only store objects that have index greater than j, that way we don't account for collision twice
+                    if (j > i):
+                        objects_within_2sqs[idx] = j
+        for j_idx in range(25):# we set object_within_2sqs to have length 25
```
sbalanovich, line 98: Seems a bit like you're using a "magic number", going through 25 iterations like this. Consider integrating this code into the nested for loop above

HW2/P5/physics.pyx
```diff
@@ -113,17 +145,30 @@ cpdef update(FLOAT[:, ::1] XY,
...
 
         # update positions
         #
         # SUBPROBLEM 1: parallelize this loop over 4 threads (with static
         #    scheduling).
         # SUBPROBLEM 2: update the grid values.
-        for i in range(count):
+        
+        #reset all values in grid
+        for idx_x in prange(grid_size,num_threads=num_threads,schedule='static'):
```
sbalanovich, line 159: This nested loop seems unnecessary since you could achieve the same effect by integrating this into the nested loops below

HW2/P2/parallel_vector.pyx
```diff
@@ -73,19 +73,31 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
    # Use parallel.prange() and a lock for each element of counts to parallelize
    # data movement.  Be sure to avoid deadlock, and double-locking.
    ##########
    with nogil:
+       #for r in prange(repeat, num_threads=4):
        for r in range(repeat):
-           for idx in range(src.shape[0]):
+           for idx in prange(src.shape[0], num_threads=4):
+               #always get lower numbered lock first
+               if src[idx] < dest[idx]:
```
Peilin-D, line 86: Why get lower numbered lock first?

sdaulton, line 86: To avoid deadlock

HW2/P3/mandelbrot.pyx
sdaulton, line 84: This is exactly the opposite!  It is a vector of float8's where each float8 as all bits set to 1!

HW2/P3/mandelbrot.pyx
sdaulton, line 99: To be honest, I am not sure, but the idea of copying a float to each spot in a float vector sounds more costs than a bitwise and.  I could be wrong though.  Interesting question

HW2/P3/mandelbrot.pyx
sbalanovich, line 84: Oh.. sorry I guess I must have misread this. In that case, why not just do `AVX.float_to_float8(1.0)`?

HW2/P4/driver.py
sdaulton, line 44: As stated in the problem, "to correctly compute iteration i for thread n, threads n, n − 1, and n + 1 must have
completed iteration i − 1."  Since I join all threads at the end of each iteration, all threads are started and killed in each iteration.  Thus all threads complete iteration i-1 before any thread starts computing iteration i.

Note: I did not complete the extra credit to allow threads to begin the next iteration asynchronously.  That would require locking as you describe.

sbalanovich, line 44: Ooh, I see that makes sense! Thanks for clearing that up :)

HW2/P5/P5.txt
sdaulton, line 37: Griding leads to the 50x boost for one thread.  Increased overhead leads to the 72x increase for 4 threads (better performance, but doesn't scale at 2x speed for 2x as many threads).

HW2/P5/P5.txt
sdaulton, line 56: Because we do not increase overhead per thread (overhead overall increases though as we have to sort).  Since the sorting cost is fixed regardless of threads, we see a greater performance gain using 4 threads because there is no additional cost and we can compute faster with sorted values.

HW2/P5/physics.pyx
sdaulton, line 159: Yeah but that would require executing an if statement every iteration to check if the i == 0.  Pick your poison.  Maybe that way would be faster

HW2/P5/P5.txt
sdaulton, line 75: Depends what the goal is.  Speed or correctness :)

HW2/P3/mandelbrot.pyx
sdaulton, line 84: Because (using 32 bit floats) 1.0 (decimal) = 00000000000000000000000000000001(binary)

The value I computed is 11111111111111111111111111111111(binary)

PR #333, opened 2015-11-04
HW2 Submission

apetschek:


HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +54,47 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+    # Use parallel.prange() and a lock for each element of counts to parallelize
+    # data movement.  Be sure to avoid deadlock, and double-locking.
+    ##########
+    
+    with nogil:              
+            for r in range(repeat):
+                
+                #Parallelize over 4 threads                
+                for idx in prange(src.shape[0],num_threads=4):
+                    if counts[src[idx]] > 0:
```
ghost, line 87: How are you ensuring that counts[src[idx]] > 0 continues to be met? With your locking strategy, it is possible for this to eventually be violated.
e.g. say counts[2] = 1 and 2 threads are operating on that same entry.  If they both reach (and pass the condition), then they will end up setting counts[2] = -1 once they are both done.

HW2/P2/parallel_vector.pyx
```diff
@@ -94,22 +102,32 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
+    #  to parallelize data movement.  Be sure to avoid deadlock, as well as
+    # double-locking.
+    ##########
+    
+    with nogil:
+            for r in range(repeat):
+                
+                #Parallelize over 4 threads                  
+                for idx in prange(src.shape[0],num_threads=4):  
+                    if counts[src[idx]] > 0:
```
ghost, line 122: Same comment as above...

HW2/P2/P2.txt
```diff
@@ -0,0 +1,26 @@
...
+
+#######################
+# Medium/Uncorrelated #
+#######################
+I experimented with various values of N locks. Specifically, I tried over the range(1,33). For medium grained uncorrelated, There seems to be a linear trend of increasing time as N increases. Please refer to my plots which depict the time it takes for various values of N. Since N continues to increase in this case, it appears that a lower value of N is best.
+
+#####################
+# Medium/Correlated #
+#####################
+For medium grained correlated, there appears to be a minimum around 14 locks. Please see my plots which depict the time it takes for various values of N. This makes sense that there exists a minimum whereas above I did not find one. In correlated movement, we are looking grabbing certain sets of locks at a time, so there is a way to optimize which/how many locks we grab for optimization. If we grab too few, we may not be taking full advantage of the correlated movement. If we take too many, we may be overdoing it, undermining the advantages of correlated. thus, we need to find the right balance which appears to be around 15-20 locks from my plot.
```
ghost, line 26: The shape of my graph is very similar (upward-sloping for medium uncorrelated, roughly U-shaped for medium correlated with a minimum at ~20), though my overall runtimes are higher as I locked both the src and dest positions to avoid any individual entry going < 0 (see my comment below). It's interesting to see that the shape remained the same despite our different locking strategies!

HW2/P3/mandelbrot.pyx
```diff
@@ -4,68 +4,114 @@ cimport cython
...
-                z = 0
+        
+        #Parallelize over 1,2, and 4 threads. 4 threads shown here
+        for i in prange(in_coords.shape[0],num_threads=4,schedule='static',chunksize=1):
+            for j in range(in_coords.shape[1]/8): #note: divide by 8 since we will grab 8 elements at a time 
+                
+                #create float8 constants to be used in operations below
+                const_zero = AVX.float_to_float8(0.0)
+                const_one = AVX.float_to_float8(1.0) 
+                const_four = AVX.float_to_float8(4.0)
```
ghost, line 50: Nice! In retrospect, a much better naming strategy than mine (counts, counts_add, and exit_check)! :)

HW2/P3/P3.txt
```diff
@@ -0,0 +1,23 @@
...
+# AVX #
+#######
+
+When instruction level parallelism is added, performance increases even more.
+
+1 Thread: 0.73 seconds
+2 Threads: 0.37 seconds
+4 Threads:  0.19 seconds
+
+Clearly, the addition of instruction level parallelism using AVX has drastically reduced the time. With a single thread, AVX appears to improve time by a factor of 6-7. It’s likely that a full 8 is not hit because we are not always working on all 8 registers at once. There are times when some registers are above our threshold and thus using a mask allows us to avoid operating on them. perhaps we could have swapped out completed parts of the register with new ones as soon as they are ready instead of waiting for all 8 locations in the register to meet our threshold specifications before inserting in a new set of 8 into the register.
```
ghost, line 23: Nice job! :+1:

HW2/P4/P4.txt
```diff
@@ -0,0 +1,23 @@
...
+
+My code runs slowest with only 1 thread and time reduces as threads increase. Note, there does seem to be diminishing returns as the number of threads increases.
+
+##################
+# Thread Control #
+##################
+
+My primitives strategy is as follows. The goal is to ensure that no thread begins iterating on another pass until neighboring threads n+1 and n-1 have finished the prior iteration pass. To do this, I first set up a 2D array of events that are all initially set to false that represent the 10 iterations per row in the image. Then, I only permit a thread to continue if the n+1 and n-1 threads had finished. To set a thread as finished, I set the iteration event for that thread to true with .set() when that iteration completed. 
+
+This methodology satisfies the extra credit condition that threads can continue working as long as the n+1 and n-1 condition has been met. Note, I had to account for a few edge cases. First, the 0th thread does not have a n-1 thread to worry about. Similarly, the nth thread does not have a n+1 thread to worry about. And, finally, if there is only 1 thread, then these n+1 and n-1 conditions no longer apply.
```
ghost, line 22: I interpreted this differently - e.g. for 4 threads, the 0th thread has to wait for the 1st thread and the 4th thread (i.e. if there is no thread_id - 1, then the wait condition "wraps around" to the highest thread id; the same logic applies for the greatest thread id).  This left me with a lot more if-then-else's when setting up the wait logic... did you find that you were able to pass the checks for multiple iterations, even without "wrapping around"?

HW2/P4/P4.txt
```diff
@@ -0,0 +1,23 @@
...
+##########################
+
+Below are the times I found for varying amounts of threads:
+
+3.16 seconds for 10 filter passes with 1 thread.
+1.61 seconds for 10 filter passes with 2 threads.
+0.87 seconds for 10 filter passes with 4 threads.
+0.78 seconds for 10 filter passes with 8 threads.
+
+My code runs slowest with only 1 thread and time reduces as threads increase. Note, there does seem to be diminishing returns as the number of threads increases.
```
ghost, line 14: I also saw diminishing returns.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,45 @@
...
+
+As we can see, there is a speed up relative to the speeds listed above in Subproblem 2.
+
+################
+# Subproblem 4 # 
+################
+
+Without Locks: ~2200 simulation frames per second (same benchmark as above)
+With Locks: ~2200 simulation frames per second 
+
```
ghost, line 44: That's interesting - wouldn't you expect the locking to create at least a tiny bit of overhead?  It shaved off ~5% of the simulation frames per second in my implementation.

HW2/P5/physics.pyx
```diff
@@ -91,40 +126,50 @@ cpdef update(FLOAT[:, ::1] XY,
...
-
+ 
+             #update ball i's new grid position to grid_xnew,grid_ynew
+             #make sure to use unsigned int for grid_x/y to convert from float           
+            if (XY[i,0] > 0) & (XY[i,0] < 1) & (XY[i,1] > 0) & (XY[i,1] < 1):
+                grid_xnew = <unsigned int>(XY[i,0]/grid_spacing)
+                grid_ynew = <unsigned int>(XY[i,1]/grid_spacing)
+                Grid[grid_xnew, grid_ynew] = i
 
 def preallocate_locks(num_locks):
```
ghost, line 174: Nice!

HW2/P4/P4.txt
apetschek, line 22: Oh, interesting. Yeah i had not considered the wrap around. Everything seemed to work fine with my code where if you are on the 0th thread or highest nth thread then you only worry about the adjacent thread in the list (1st and n-1 respectively) and don't worry about wrapping. If i recall correctly, I think this is OK because the threads'  iterations employ 3x3 filtering, so the cells you depend on are those in the rows immediately above and below . Thus, the top thread's 3x3 won't wrap to the bottom (i don't think).

HW2/P5/P5.txt
apetschek, line 44: Oh, that's possible. In hindsight, I probably should have created a histogram of simulation frames per second or something that could have better quantified the performance exactly. this way, a difference of something like 5% would have been more clear. it is likely that there was some overhead for creating this extra locking component.

HW2/P2/parallel_vector.pyx
zezhouliu, line 87: yes! this happens if counts[src[idx]] changes between now and acquiring the lock. You usually want to check the same condition again after acquiring the lock

PR #341, opened 2015-11-05
Hw2

jonathanhfriedman:


HW2/P2/P2.py
```diff
@@ -74,9 +77,19 @@
...
-    assert counts.sum() == total, "Wrong total after move_data_medium_grained"
-    print("Medium grained correlated: {} seconds".format(t.interval))
+    for N in range(1,21):
+		with Timer() as t:
+			move_data_medium_grained(counts, src, dest, 100, N)
+		assert counts.sum() == total, "Wrong total after move_data_medium_grained"
+		Nlistc.append(t.interval)
+		print("N: {}  Medium grained correlated: {} seconds".format(N, t.interval))
+		
+plt.plot(np.arange(1,21), Nlist, color='b', label='Uncorrelated')
```
neelspatel, line 89: This code is very clean and looks great; the only change I would suggest are more descriptive variable names for Nlistc and Nlist

HW2/P2/P2.txt
```diff
@@ -0,0 +1,5 @@
+Both parallel approaches were at least 30 times slower than serial code (0.3s vs 10-18s), presumably due to overhead. The fine-grained and medium-grained approaches took about the same amount of time.
+
+P2graph.png shows a plot of medium-grained locking performance for N from 1 to 20. For some reason, even values of N produced a better runtime than odd values. In addition, for correlated data, N=10 resulted in much worse performance than any other value of N, though this may be random noise.
+
+N=4 produced good runtimes (relatively speaking) for both correlated and uncorrelated data. N values greater than 15 also worked well for correlated data. This is probably because one lock frequently covered both the source and destination, thus reducing locking overhead.
```
neelspatel, line 5: I agree with your answers here, however I observed a different result for the number of locks (10-20 locks for either uncorrelated or correlated data), but i agree that N values greater than 15 are ideal for correlated data.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,5 @@
+Both parallel approaches were at least 30 times slower than serial code (0.3s vs 10-18s), presumably due to overhead. The fine-grained and medium-grained approaches took about the same amount of time.
+
+P2graph.png shows a plot of medium-grained locking performance for N from 1 to 20. For some reason, even values of N produced a better runtime than odd values. In addition, for correlated data, N=10 resulted in much worse performance than any other value of N, though this may be random noise.
```
neelspatel, line 3: Great graph, makes it pretty clear what regions are ideal for each set of data. I wonder what a smoothed grapha averaged over multiple runs would look like

HW2/P2/parallel_vector.pyx
```diff
@@ -51,65 +51,80 @@ cdef void free_N_locks(int N, omp_lock_t *locks) nogil:
...
+	# I just lock each memory address immediately before accessing it, and release it 
+	# immediately after.
+	##########
+	with nogil:
+		for r in range(repeat):
+			for idx in prange(src.shape[0], num_threads=4):
+				acquire(&locks[src[idx]])
+					
+				if counts[src[idx]] > 0:
+					release(&locks[src[idx]])
```
neelspatel, line 87: A comment explaining the order of the acquire/release sequence on the locks here would be really helpful in understanding the purpose of this code. It looks logical, but explaining the logic would be awesome

HW2/P3/P3.txt
```diff
@@ -0,0 +1,18 @@
...
+
+With AVX
+# of Threads	Millions of Complex FMAs / sec
+---------------------------------------------
+1		101.646625339
+2		196.69971906
+4		342.408732116
+
+
+I assume the slowdown for AVX is because my computer has to simulate AVX. I don’t get the GIL at any point in my computation loop, and I don’t see any other things that would cause slowdown. Aside from that, more threads still makes things faster.
```
neelspatel, line 18: That's interesting, I observed a speedup for AVX, so the slowdown coming from simulating AVX could make sense - I'd be curious to know what platform you're using, and it would be really interesting to study the relative speed of simulated vs real AVX

HW2/P3/mandelbrot.pyx
```diff
@@ -1,42 +1,76 @@
...
+            imags[i, j+3],
+            imags[i, j+2],
+            imags[i, j+1],
+            imags[i, j])
+
+          z_re = AVX.float_to_float8(0.)
+          z_im = AVX.float_to_float8(0.)
+          counts = AVX.float_to_float8(0.)
+
+          # The computation takes place in this loop. The counts are only updated
```
neelspatel, line 63: Great commenting, made the logic very easy to follow

HW2/P5/P5.txt
```diff
@@ -0,0 +1,41 @@
...
+
+With Locking:
+
+
+# threads             frames/sec
+--------------------------------
+1                     ~570
+4                     ~930
+
+Locking slows the code down a little bit, but the slowdown is really not bad.
```
neelspatel, line 41: Agree with your answers here, but I saw a much larger slowdown for locking, which is interesting

HW2/P3/P3.txt
jonathanhfriedman, line 18: I'm using a late 2013 MacBook Pro, which seems pretty standard. I was surprised by the slowdown.

HW2/P2/P2.txt
nicodri, line 5: Observed the sam results nice!

HW2/P3/P3.txt
nicodri, line 18: Observed a speed up too

HW2/P4/driver.py
```diff
@@ -14,16 +14,47 @@
...
+    def process_image(threadnum, tmpA, tmpB, offset=num_threads):
+        ''' do filtering for the threadnum-th thread '''
+        for iter in range(iterations):
+            # We don't have to check for this thread finishing iteration i-1 because
+            # it does each of its iterations in sequence.
+            if iter > 0:
+                if threadnum != 0:
+                    events[(threadnum-1)*iterations+iter-1].wait()
+                if threadnum != num_threads-1:
+                    events[(threadnum+1)*iterations+iter-1].wait()
```
nicodri, line 42: Code still really easy to follow. To avoid the corner case here you could have set num_threads + 2 events per iteration and having the first and the last event always set (they will correspond to the two borders)

HW2/P5/P5.txt
nicodri, line 41: Same range of results and also I got much more slow down with locking

HW2/P5/physics.pyx
```diff
@@ -55,31 +56,53 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+    # Need to check for collisions with the object o in grid squares marked with *
+    # (since the balls can collide from up to two squares away):
+    # *  *  *  *  *
+    # *  *  *  *  *
+    # *  *  o  *  *
+    # *  *  *  *  *
+    # *  *  *  *  *
+    #
```
nicodri, line 82: Nice drawing, I havent thought about it but it makes things much clearer!

HW2/P5/P5.txt
jonathanhfriedman, line 41: I think I should have put in another lock somewhere...

HW2/P2/parallel_vector.pyx
oyuchoi, line 87: yes, it looks like you're always locking src first (even though you might want to lock dest first instead?)

PR #347, opened 2015-11-05
HW2

stephenshifang:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,30 @@
...
+Medium grained correlated, N = 3: 5.79343700409 seconds
+Medium grained correlated, N = 4: 6.34704494476 seconds
+Medium grained correlated, N = 5: 7.08802008629 seconds
+Medium grained correlated, N = 6: 7.9741590023 seconds
+Medium grained correlated, N = 7: 8.97585201263 seconds
+Medium grained correlated, N = 8: 9.94677901268 seconds
+Medium grained correlated, N = 9: 11.0350589752 seconds
+Medium grained correlated, N = 10: 11.9213550091 seconds
+
+Serial vector update runs fastest followed by fine-grained locking and then medium-grained locking. In this case, the overhead of creating and destroying locks outweighs the parallel speedup.
```
samighoche, line 27: I think if you tried a much larger range for N you would see that the time goes down until an N of around 20, and then goes up drastically after.

HW2/P2/parallel_vector.pyx
```diff
@@ -54,39 +54,47 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+    with nogil:
+        for r in range(repeat):
+            for idx in prange(src.shape[0], num_threads=4):
+                src_idx, dest_idx = src[idx], dest[idx]
+                if counts[src_idx] > 0:
+                    # first increment dest[idx]
+                    acquire(&(locks[dest_idx]))
```
samighoche, line 88: I think you want to acquire the src lock before the if statement, because otherwise, you might be inside the execution under the assumption that the condition holds, but some other thread will have invalidated the condition in the meantime before you acquire the lock.
If you acquire the lock before the if statement and release it after, you will also have to deal with deadlocks by enforcing an order of which lock to acquire first

HW2/P3/mandelbrot.pyx
```diff
@@ -32,17 +70,50 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                    # increment counts where magnitude_squared is less than 4
+                    iter_count = AVX.add(iter_count, AVX.bitwise_and(mask, AVX.float_to_float8(1.)))
+                    # z = z * z + c
+                    # z_real = z_real**2 - z_imag**2 + c_real
+                    # z_imag = 2 * z_real * z_imag * i + c_imag * i
+                    tmp_real = AVX.add(AVX.sub(z_real_sq, z_imag_sq), c_real)
+                    tmp_imag = AVX.add(AVX.mul(AVX.float_to_float8(2.), AVX.mul(z_real, z_imag)), c_imag)
+                    z_real, z_imag = tmp_real, tmp_imag
+                
+                AVX.to_mem(iter_count, &(out_vals[0]))
```
samighoche, line 114: nice solution to the problem of writing back

HW2/P3/mandelbrot.pyx
```diff
@@ -32,17 +70,50 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
-                    z = z * z + c
-                out_counts[i, j] = iter
-
-
+                    # increment counts where magnitude_squared is less than 4
+                    iter_count = AVX.add(iter_count, AVX.bitwise_and(mask, AVX.float_to_float8(1.)))
+                    # z = z * z + c
+                    # z_real = z_real**2 - z_imag**2 + c_real
+                    # z_imag = 2 * z_real * z_imag * i + c_imag * i
+                    tmp_real = AVX.add(AVX.sub(z_real_sq, z_imag_sq), c_real)
```
samighoche, line 110: important that you make temporary variables to avoid overwriting variables that you still need

HW2/P4/P4.txt
```diff
@@ -0,0 +1,11 @@
...
+3.39166116714 seconds for 10 filter passes.
+
+num_threads 2
+1.74576187134 seconds for 10 filter passes.
+
+num_threads 4
+1.59977984428 seconds for 10 filter passes.
+
+With 2 threads the speedup is around 2 times relative to serial. However, with 4 threads, the run time is only slightly faster than 2 threads.
+In this implemention of the 3x3 median filter, we wait until iteration i finishes before starting iteration i + 1. Therefore, no threads start an iteration of filtering before the data for that iteration is ready.
```
samighoche, line 11: It's nice that you get a speedup, but you're making all threads wait for each other at every iteration. Instead, what I did was have each thread run a for loop over all iterations, but I have thread N wait for a signal that threads N-1 and N+1 are done with the previous iteration that thread N was working on, to make sure that it's using up to date information. By doing that, a thread only waits for the two threads surrounding it.

HW2/P5/physics.pyx
```diff
@@ -69,17 +73,28 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+
+    # get grid index of XY1
+    gridx = int(XY1[0]/grid_spacing)
+    gridy = int(XY1[1]/grid_spacing)
+    for x in range(max(0, gridx-search_size), min(Grid.shape[0], gridx+search_size+1)):
+        for y in range(max(0, gridy-search_size), min(Grid.shape[1], gridy+search_size+1)):
+            j = Grid[x, y]
+            # to avoid collision with itself and double counting
```
samighoche, line 83: nice

HW2/P5/P5.txt
```diff
@@ -0,0 +1,83 @@
...
+Sample simulation frames per second with 1 thread:
+263.170364421 simulation frames per second
+258.759107426 simulation frames per second
+255.524323007 simulation frames per second
+230.91725239 simulation frames per second
+260.40794702 simulation frames per second
+
+Sample simulation frames per second with 4 threads:
+475.39494994 simulation frames per second
+510.438802761 simulation frames per second
```
samighoche, line 33: I'm surprised that the grid didn't make an even bigger difference, maybe my comment on your double for loops partly explains why (that would account for about a 4x speedup)

HW2/P5/physics.pyx
```diff
@@ -69,17 +73,28 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-                collide(XY1, V1, XY2, V2)
-
-            # give a slight impulse to help separate them
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
+
+    # get grid index of XY1
+    gridx = int(XY1[0]/grid_spacing)
+    gridy = int(XY1[1]/grid_spacing)
+    for x in range(max(0, gridx-search_size), min(Grid.shape[0], gridx+search_size+1)):
```
samighoche, line 80: Instead of checking if i < j, you can make sure to only look three steps ahead of x (not x-3 to x+3), and same for y. That way, you know you never double count, and you run fewer computations

HW2/P3/mandelbrot.pyx
```diff
@@ -32,17 +70,50 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+                    tmp_imag = AVX.add(AVX.mul(AVX.float_to_float8(2.), AVX.mul(z_real, z_imag)), c_imag)
+                    z_real, z_imag = tmp_real, tmp_imag
+                
+                AVX.to_mem(iter_count, &(out_vals[0]))
+                for k in range(8):
```
kevinkeyjkw, line 115: I did this before too. But I realized if I store them in backwards for c_real,c_imag to start with, then I could just call AVX.to_mem(iter_count, &(out_vals[0])) without any for loop later. This could optimize the code a bit.

HW2/P4/driver.py
```diff
@@ -19,12 +19,14 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
-
     for i in range(iterations):
-        filtering.median_3x3(tmpA, tmpB, 0, 1)
-        # swap direction of filtering
+        threads = []
+        for tid in range(num_threads):
+            thread = threading.Thread(target=filtering.median_3x3, args=(tmpA, tmpB, tid, num_threads))
+            threads.append(thread)
+            thread.start()
+        [thread.join() for thread in threads]
```
kevinkeyjkw, line 28: Could you have put thread.join() inside the for loop?

HW2/P2/P2.txt
```diff
@@ -0,0 +1,30 @@
...
+Fine-grained locking runs faster than medium-grained locking as it allows for more concurrency.
+With the medium-grained locking scheme, a smaller N is associated with better the performance. This is consistent with the observation that fine-grained locking outperforms coarse-grained locking.
+With coarse-grained locking, as N increases, correlated data transfer is generally faster than uncorrelated data transfer. Since the source and destination are close, concurrency is not affected as much when N increases.
```
kevinkeyjkw, line 30: Good observations!

HW2/P3/mandelbrot.pyx
```diff
@@ -4,19 +4,57 @@ cimport cython
...
 from cython.parallel import prange
-
+cdef: 
+    int NUM_THREADS = 4
 
 cdef np.float64_t magnitude_squared(np.complex64_t z) nogil:
     return z.real * z.real + z.imag * z.imag
 
 @cython.boundscheck(False)
 @cython.wraparound(False)
```
kevinkeyjkw, line 14: I found some other compiler directives that could speed up my program such as @cython.initializedcheck(False)
"If set to True, Cython checks that a memoryview is initialized whenever its elements are accessed or assigned to. Setting this to False disables these checks. Default is True." and @cython.nonecheck(False)

You can find other directives here: http://docs.cython.org/src/reference/compilation.html#compiler-directives

HW2/P2/P2.txt
stephenshifang, line 27: I see. Good point.

PR #353, opened 2015-11-05
Hw2

jonahkall:


HW2/P2/parallel_vector.pyx
```diff
@@ -70,23 +70,40 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
+        else:
+          omp_set_lock(&locks[minlock])
+          omp_set_lock(&locks[maxlock])
+          if counts[src[idx]] > 0:
+            counts[dest[idx]] += 1
+            counts[src[idx]] -= 1
+          omp_unset_lock(&locks[minlock])
+          omp_unset_lock(&locks[maxlock])
+
+  free_N_locks(counts.shape[0], locks)
```
Elena-Zhao, line 107: This is great that you considered many cases when posing the lock. I understand that if src and dest are the same we only need to lock once, but do we really have to always lock the small one first? I don't see any necessities here since it's just data movement that doesn't involve any comparisons

HW2/P3/mandelbrot.pyx
```diff
@@ -42,7 +45,89 @@ cpdef mandelbrot(np.complex64_t [:, :] in_coords,
...
+
+    assert in_coords.shape[1] % 8 == 0, "Input array must have 8N columns"
+    assert in_coords.shape[0] == out_counts.shape[0], "Input and output arrays must be the same size"
+    assert in_coords.shape[1] == out_counts.shape[1],  "Input and output arrays must be the same size"
+
+    with nogil:
+        # Define constants which we will use repeatedly.
+        fours = AVX.float_to_float8(4.0)
+        ones = AVX.float_to_float8(1.0)
+        twos = AVX.float_to_float8(2.0)
```
Elena-Zhao, line 77: This is really good habit! Defining the frequently used variable in the first place and reuse them later... I forgot to do do that and later the codes get really messy...

HW2/P4/driver.py
```diff
@@ -15,15 +15,44 @@
...
+def median_worker(tmpa, tmpb, j, events, num_threads, iters):
+    for i in range(iters):
+        # Wait on all the necessary conditions in order to execute (extra credit).
+        if bounds_check(j - 1, num_threads) and i - 1 >= 0:
+            events[i - 1][j - 1].wait()
+        if bounds_check(j, num_threads) and i - 1 >= 0:
+            events[i - 1][j].wait()
+        if bounds_check(j + 1, num_threads) and i - 1 >= 0:
+            events[i - 1][j + 1].wait()
+        filtering.median_3x3(tmpa, tmpb, j, num_threads)
```
Elena-Zhao, line 32: I did this using exactly the same way by creating a 2-D array of events to record the thread status. While the 'bounds_check()' is necessary, I don't think it is a good idea to check in every iteration, simply changing the starting point could be a better solution that avoids redundant checks.

HW2/P5/physics.pyx
```diff
@@ -113,18 +134,27 @@ cpdef update(FLOAT[:, ::1] XY,
...
+            # Actually update the positions.
             for dim in range(2):
                 XY[i, dim] += V[i, dim] * t
 
+            # Tell the grid where the new ball is.
+            if (<int> (XY[i, 0]/grid_spacing) < grid_size) and (<int> (XY[i, 1]/grid_spacing) < grid_size) and (<int> (XY[i, 1]/grid_spacing) >= 0) and (<int> (XY[i, 0]/grid_spacing)) >= 0:
+                Grid[<int> (XY[i, 0]/grid_spacing), <int> (XY[i, 1]/grid_spacing)] = i
 
 def preallocate_locks(num_locks):
     cdef omp_lock_t *locks = get_N_locks(num_locks)
```
Elena-Zhao, line 160: Generally nice. would be better if there's a graph to illustrate the efficiency increase.

HW2/P2/P2.txt
```diff
@@ -0,0 +1,10 @@
+Based on the data (see medium.svg, which is I git added to this directory), we see that
+in the uncorrelated case, we should use a small number of locks, around 10.  In the 
+correlated case, it makes the most sense to use a number of locks around 30-40.  This allows the right
+balance in the tradeoff between the computational overhead of having a lot of locks around and locking them
+vs. the benefit of being able to lock a single lock to modify two data elements.  This makes sense --
+in the uncorrelated case, we are very unlikely to have the same lock be able to modify both data elements,
+while in the correlated case, this looks to be happneing all the time with a lock size of around 10. Note:
+in the correlated model, things being switched are within 10 spots of each other, so it makes sense that
+a size of around 20-40 will be good, such that it is unlikely for us to have to grab another lock group
+(only necessary across halves).
```
xidexia, line 10: My best N is about 10, but we have similar observation. Your choice makes sense based on your results.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,13 @@
...
+
+I decided to go for the extra credit approach of allowing asynchronous
+progress by mandating only that the requisite set of the threads
+(n-1), (n+1), and n had completed their progress on iteration i-1 before
+allowing thread n to advance to iteration i.  I did this with a 2D array of
+events (the Event primitive from the python threading library).  Thus I was
+able to use a relatively simple design where if thread n wanted to do
+iteration i, it would wait on the events that threads n, n-1, and n+1 had
+completed their respective iteration i-1s, and then it would set the event
+flag for its own event after doing the computation in filtering.median_3x3.
```
xidexia, line 13: Nice explanation! In my result, the amount of time decreased dramatically when the number of threads changed from 1 to 2. However, when it came to 4, the difference of time was slight.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,99 @@
...
+Part 2:
+
+(for these two numbers, same as above)
+1 thread, no gridding: 6.4436254365 simulation frames per second
+4 threads, no gridding: 9.73795787025 simulation frames per second
+1 thread, gridding: ~890.23471 simulation frames per second
+4 threads, gridding: ~2007.10168 simulation frames per second
+
+*MASSIVE* performance increase! This is what we expect though
+in going from an O(N^2) to an O(N) algorithm with N very large.
```
xidexia, line 17: Nice! I think the griding largely helps to improve the performance because the time complexity for sub_update() improves to O(N) while it is originally O(N^2).

PR #357, opened 2015-11-05
P2 Completed

Sdeppen:


HW2/P2/parallel_vector.pyx
```diff
@@ -99,17 +99,17 @@ cpdef move_data_medium_grained(np.int32_t[:] counts,
...
                if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+                  acquire(&(locks[dest[idx] / N]))
+                  counts[dest[idx]] += 1
+                  release(&(locks[dest[idx] / N]))
+
+                  acquire(&(locks[src[idx] / N])) 
+                  counts[src[idx]] -= 1
+                  release(&(locks[src[idx] / N]))
```
k-ye, line 113: Did you forget to commit the rest part of this homework?

HW2/P2/parallel_vector.pyx
```diff
@@ -74,21 +74,21 @@ cpdef move_data_fine_grained(np.int32_t[:] counts,
...
-   # Use parallel.prange() and a lock for each element of counts to parallelize
-   # data movement.  Be sure to avoid deadlock, and double-locking.
-   ##########
    with nogil:
-       for r in range(repeat):
-           for idx in range(src.shape[0]):
+      for r in range(repeat):
+         for idx in prange(src.shape[0], num_threads = 4):
+            if src[idx] != dest[idx]:   
                if counts[src[idx]] > 0:
```
sdaulton, line 81: There is currently a race condition.

Consider the following scenario:
Suppose counts[src[idx]] equals 1. Assume for simplicity that we only have 1 core (i.e. only one thread can be running at a time).
Thread1 checks counts[src[idx]] > 0, which is true, and thus enters the body of the if clause.

Interrupt! Thread1 stops, Thread2 runs.
Then Thread 2 checks counts[src[idx]] > 0, which is true, and it enters the body of the if clause.
Thread2 gets the lock for counts[dest[idx]], increments it, releases the lock.
Thread2 gets the lock for counts[src[idx]], decrements it, and releases the lock.
Interrupt! Thread2 stops, Thread1 runs.
Thread1 gets the lock for counts[dest[idx]], increments it, releases the lock.
Thread1 gets the lock for counts[src[idx]], decrements it, and releases the lock.
But now counts[src[idx]] equals -1! Not good.

You can fix this by getting both locks before checking if counts[src[idx]] > 0. Note, you need to hold both locks when you decrement dest and increment src in order to avoid the race condition.

Note: With my recommended fix, you will also need to acquire locks in order to avoid deadlock.

HW2/P2/parallel_vector.pyx
Sdeppen, line 113: No. Some stuff came up and I ended up not being able to finish this pset at all.

HW2/P2/parallel_vector.pyx
Sdeppen, line 81: Oh, that makes a lot of sense. I was having trouble thinking about it, so this really helps. Thanks.

PR #358, opened 2015-11-05
HW2 finished

nicodri:


HW2/P2/P2.txt
```diff
@@ -0,0 +1,19 @@
...
+# ## Locking explained
+
+Fine grained: one lock per ceil.
+
+Medium grained: the same lock is shared for each group of ceils of indices between k * N and (k+1) * N in the array.
+
+# ## Medium Grained Value analysis
+
+I plot over 4 iterations the minimum and the average value of the execution time for 12 values of N in both cases, correlated and uncorrelated.
+It seems that for the uncorrelated a bigger value, around 10 is more suited, and for the correlated mostly around 5.
```
andrewreece, line 19: I'd be curious to see what speed differences you see for correlated data with a larger N, eg. N=20.  I noticed a substantial speed gain at this configuration, which I attributed to the fact that almost all of the src/dest values are within a 10-point spread of each other.  So with an N that exceeds the point spread, we only have to acquire and release one lock instead of two.  Not sure if that's correct, but over a lot of iterations that would explain a speed gain under this regime.

HW2/P3/P3.txt
```diff
@@ -0,0 +1,7 @@
+# ## Multithreading without AVX
+
+Over 4 iterations I plot the min and the average execution time (which are actually pretty close). 4 threads run 4 times faster than 1 thread, it seems that only a few amount of time is lost in communications and settings.
+
+# ## Multithreading with AVX
+
+On average, this version is 6 times faster than the one without AVX for the same number of threads. Its less than 8 times because of the time taken to do the consecutive operations and objects creation in AVX. Then, comparing the time over the number of thread show more time used for communications and system works than without AVX.
```
andrewreece, line 7: I saw about a 6-7x speedup as well.

HW2/P4/P4.txt
```diff
@@ -0,0 +1,33 @@
...
+Sync version in 0.358500003815 seconds with 4 threads.
+Async version in 0.36866402626 seconds with 4 threads.
+
+_________ 10 iterations
+
+Sync version in 3.46021699905 seconds with 1 threads.
+Async version in 3.43095612526 seconds with 1 threads.
+Sync version in 1.82300400734 seconds with 2 threads.
+Async version in 1.81348586082 seconds with 2 threads.
+Sync version in 1.65510702133 seconds with 4 threads.
```
andrewreece, line 28: I also did not see much speed gain moving from 2 to 4 threads here.

HW2/P4/driver.py
```diff
@@ -27,6 +28,77 @@ def py_median_3x3(image, iterations=10, num_threads=1):
...
+
+def py_median_3x3_async(image, iterations=10, num_threads=1):
+    ''' repeatedly filter with a 3x3 median '''
+    tmpA = image.copy()
+    tmpB = np.empty_like(tmpA)
+
+    # List of event for each iteration & thread
+    events = []
+    for i in xrange(iterations):
+        events_list = [threading.Event()]
```
andrewreece, line 60: I used a very similar implementation with Event() objects.  Glad to see I wasn't alone!

HW2/P2/P2.txt
stephenshifang, line 19: I agree. In my implementation I only tried N = 1 to 11 so I didn't observe the plateau but the reasoning makes sense.

HW2/P3/P3.txt
stephenshifang, line 7: I got the same result as well. AVX seems like the way to go.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,42 @@
...
+
+
+## Locking performance (with 4 threads, sorting and griding)
+
+With 1 thread: average simulation frames per second of 424
+With 4 threads: average simulation frames per second of 680
+
+The execution time is larger in the locks case because some threads need to wait to acquire the lock. Otherwise the code is exactly the same in the two case. 
+But the time lost enables to come up with an accurate collision updates for each of the situation. In particular if three balls are in a collision (A hits B and B hits), we are sure that the velocities are correctly updated
+
```
stephenshifang, line 42: Great visualizations below!

PR #361, opened 2015-11-08
HW2

LeonhardFS:


HW2/P2/parallel_vector.pyx
```diff
@@ -54,62 +54,83 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+    cdef:
+        int idx, r
+        omp_lock_t * locks = get_N_locks(counts.shape[0])
+        int thread_count = 4  # omp_get_num_threads()
+
+    ##########
+    # Your code here
+    # Use parallel.prange() and a lock for each element of counts to parallelize
+    # data movement.  Be sure to avoid deadlock, and double-locking.
+    ##########
```
samighoche, line 84: I think you should acquire the source lock before the if statement because otherwise, the condition might be true when evaluating it, but once you're actually within the if statement execution (which only happens with some assumptions about the value of the src[idx]), some other thread may have already invalidated your assumptions

samighoche, line 84: Therefore, you'll also need to make sure to avoid deadlock because you won't be able to acquire and release only one lock at a time

HW2/P2/P2.txt
```diff
@@ -0,0 +1,14 @@
...
+
+Generally speaking, the correlated data was always faster except for the coarse-grained case. There, the correlated data needed around ~20s to finish whereas the uncorrelated data finsihed in ~7-8s. This could be due to a lot of cache misses.
+
+Comparing the fine grained algorithm vs. the medium grained one yields that often the fine grained algorithm is faster, especially for larger N. In the uncorrelated case the fine grained algorithm outperformed (in contrary to my presumption) the medium grained algorithm. However, in the correlated case, setting i.e. N=16 yielded a little better performance. (as we can assume elements are not farther than 10 distanced in the correlated case, everything up to N=20 seems reasonable (10 in each direction!))
+
+In conclusion, except for the comparison coarsed vs. fine/medium results tended to be highly varying. To get more reliable results, a larger dataset, a better defined testing environment and more detailed problem specification should be used. I.e. I think that my results are highly disturbed due to the OS.
```
dekatzenel, line 14: Which OS are you using?

HW2/P2/parallel_vector.pyx
```diff
@@ -17,33 +17,33 @@ cimport numpy as np
...
-cdef void free_N_locks(int N, omp_lock_t *locks) nogil:
+cdef void free_N_locks(int N, omp_lock_t * locks) nogil:
     cdef int idx
 
     for idx in range(N):
-        omp_destroy_lock(&(locks[idx]))
+        omp_destroy_lock(& (locks[idx]))
 
-    free(<void *> locks)
+    free( < void * > locks)
```
dekatzenel, line 46: Are these whitespace changes for PEP8?

HW2/P2/parallel_vector.pyx
```diff
@@ -54,62 +54,83 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
-               if counts[src[idx]] > 0:
-                   counts[dest[idx]] += 1
-                   counts[src[idx]] -= 1
+    assert src.size == dest.size, "Sizes of src and dest arrays must match"
+    with nogil:
+        for r in range(repeat):
+            for idx in range(src.shape[0]):
+                # Assumption: Code here does not need to be changed
+                # for coarse-grained locking, one lock should be used
+                if counts[src[idx]] > 0:
```
dekatzenel, line 66: I believe your assumption is correct

HW2/P2/parallel_vector.pyx
```diff
@@ -54,62 +54,83 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+        int thread_count = 4  # omp_get_num_threads()
+
+    ##########
+    # Your code here
+    # Use parallel.prange() and a lock for every N adjacent elements of counts
+    # to parallelize data movement.  Be sure to avoid deadlock, as well as
+    # double-locking.
+    ##########
+    for r in range(repeat):
+        for idx in prange(src.shape[0], nogil=True, num_threads=thread_count):
```
dekatzenel, line 116: Why move the nogil?

HW2/P2/parallel_vector.pyx
```diff
@@ -54,62 +54,83 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+            if counts[src[idx]] > 0:
+                sidx = src[idx] / N
+                didx = dest[idx] / N
+
+                # are the indices equal? --> use only one lock!
```
dekatzenel, line 121: Why use the one-lock optimization here and not in fine-grained?

HW2/P2/stats.txt
```diff
@@ -0,0 +1,61 @@
...
+Medium grained correlated: 7.27928590775 seconds
+
+N = 5:
+-------
+Serial uncorrelated: 0.255202054977 seconds
+Fine grained uncorrelated: 6.25651311874 seconds
+Medium grained uncorrelated: 8.13269591331 seconds
+Serial correlated: 0.383741855621 seconds
+Fine grained correlated: 5.90306687355 seconds
+Medium grained correlated: 7.02958297729 seconds
```
dekatzenel, line 60: Why so few values of N? Could check anything from 1 to 1000

HW2/P3/P3.txt
```diff
@@ -0,0 +1,35 @@
...
+
+Two threads:
+------------
+1074.656613 Million Complex FMAs in 2.07837510109 seconds,
+517.065765673 million Complex FMAs / second
+
+Four threads:
+-------------
+1074.656613 Million Complex FMAs in 1.28325009346 seconds,
+837.449082199 million Complex FMAs / second
```
dekatzenel, line 26: Would have been useful to explicitly say that these were without instruction level parallelism

HW2/P3/P3.txt
```diff
@@ -0,0 +1,35 @@
...
+
+AVX + 4 threads:
+----------------
+1043.531618 Million Complex FMAs in 0.250422000885 seconds,
+4167.09240527 million Complex FMAs / second
```
dekatzenel, line 31: What about AVX with 1 or 2 threads?

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
-                out_counts[i, j] = iter
-
+    assert in_coords.shape[0] == out_counts.shape[
+        0], "Input and output arrays must be the same size"
+    assert in_coords.shape[1] == out_counts.shape[
+        1],  "Input and output arrays must be the same size"
+
+    # get real / imag parts
+    real_coords = np.real(in_coords)
+    imag_coords = np.imag(in_coords)
```
dekatzenel, line 71: Elegant and I'm surprised this works with Cython memoryviews.

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+
+    # version multithreading and AVX instructions...
+    for i in prange(in_coords.shape[0], nogil=True, num_threads=4, schedule='static', chunksize=1):
+      for j in range(in_coords.shape[1] / 8):
```
dekatzenel, line 75: Alternatively, keep the range the same and modify the step size from 1 to 8.

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+
+        # parallelize 8 complex numbers
+        # first separate real and imaginary parts
+        realc = real_coords[i, j*8]
+        imagc = imag_coords[i, j*8]
```
dekatzenel, line 80: nit: unused variables

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+                                   real_coords[i, j * 8 + 1],
+                                   real_coords[i, j * 8 + 0])
+        avximagc = AVX.make_float8(imag_coords[i, j * 8 + 7],
+                                   imag_coords[i, j * 8 + 6],
+                                   imag_coords[i, j * 8 + 5],
+                                   imag_coords[i, j * 8 + 4],
+                                   imag_coords[i, j * 8 + 3],
+                                   imag_coords[i, j * 8 + 2],
+                                   imag_coords[i, j * 8 + 1],
+                                   imag_coords[i, j * 8 + 0])
```
dekatzenel, line 97: nit: explicit +0?

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+
+        avxiters = AVX.make_float8(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
+        avxones  = AVX.make_float8(1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0, 1.0)
+        avxfours = AVX.make_float8(4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0, 4.0) # to check for magnitude
+
+        # now do the loop
+        # init the series with c (we skip over the first step
+        # which is unnecessary in the original implemntation)
+        avxrealz = avxrealc
+        avximagz = avximagc
```
dekatzenel, line 107: Throws off your count compared to the original implementation

HW2/P2/P2.txt
```diff
@@ -0,0 +1,14 @@
...
+
+Here the results of some sample runs shall be discussed
+
+In the file stats.txt some runs are listed. In general, the locking introduced overhead s.t. the serial version runned always a lot faster. It shall be thereof excluded form the comparison, so only algorithms with locks are compared. To simulate the coarse-grained case, N is set to the number of samples (this will yield one lock). 
+
+Generally speaking, the correlated data was always faster except for the coarse-grained case. There, the correlated data needed around ~20s to finish whereas the uncorrelated data finsihed in ~7-8s. This could be due to a lot of cache misses.
+
+Comparing the fine grained algorithm vs. the medium grained one yields that often the fine grained algorithm is faster, especially for larger N. In the uncorrelated case the fine grained algorithm outperformed (in contrary to my presumption) the medium grained algorithm. However, in the correlated case, setting i.e. N=16 yielded a little better performance. (as we can assume elements are not farther than 10 distanced in the correlated case, everything up to N=20 seems reasonable (10 in each direction!))
```
samighoche, line 12: I think I agree with most of what you said, just that in the case of medium-grained correlated, there seems to be an optimal range of values for N (around 20 like you said), lower than that and time goes up because the same thread was more likely to acquire only one lock for both src and dest indices, but for higher than N=20, time also went up because different threads were more likely to be waiting on each other for the same lock since locks spanned a higher range of indices.

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+        avximagc = AVX.make_float8(imag_coords[i, j * 8 + 7],
+                                   imag_coords[i, j * 8 + 6],
+                                   imag_coords[i, j * 8 + 5],
+                                   imag_coords[i, j * 8 + 4],
+                                   imag_coords[i, j * 8 + 3],
+                                   imag_coords[i, j * 8 + 2],
+                                   imag_coords[i, j * 8 + 1],
+                                   imag_coords[i, j * 8 + 0])
+
+        avxiters = AVX.make_float8(0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0)
```
samighoche, line 99: You can alternatively set it like this:
avxiters = AVX.float_to_float8(0.0)

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+        for iter in range(max_iterations):
+          # for a given complex number z = x + iy
+          # it holds z^2 = (x^2 - y^2) + i(2xy)
+          # we can use that to compute z^2 + c for 8 complex numbers in parallel!
+
+          # now one step
+          avxrealz_tmp = AVX.sub(AVX.mul(avxrealz, avxrealz), AVX.mul(avximagz, avximagz))
+          avximagz_tmp = AVX.mul(avxrealz, avximagz)
+          avximagz_tmp = AVX.add(avximagz_tmp, avximagz_tmp) # adding is faster and equal to * 2
+
```
samighoche, line 118: nice job creating temp variables to avoid overwriting data that you still need and using the wrong values

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+    # # version using multithreading
+    # for i in prange(in_coords.shape[0], nogil=True, num_threads=4, schedule='static', chunksize=1):
+    #   for j in range(in_coords.shape[1]):
+    #     c = in_coords[i, j]
+    #     z = 0
+    #     for iter in range(max_iterations):
+    #       if magnitude_squared(z) > 4:
+    #         break
+    #       z = z * z + c
+    #     out_counts[i, j] = iter
```
samighoche, line 164: Very nice solution to problem 3:)

HW2/P4/P4.txt
```diff
@@ -0,0 +1,11 @@
...
+==========
+
+threads | time
+------------------------------------------------------
+1		| 3.59052395821 seconds for 10 filter passes.
+2		| 1.83024907112 seconds for 10 filter passes.
+4		| 1.687541008 seconds for 10 filter passes.
+
+
+For the implementation I choose a very basic approach. In every iteration I start N threads, which will work on the image according to the description. As for the example runs, we see that switching from serial (1 thread) to 2 threads yielded a significant performance gain. However, using 2 more threads turned out to be rather disappointing. The performance gain was very low. This is due to the great amount of overhead my implementation uses. Starting a thread typically is very costly. A better implementation therefore would start the 4 threads only once and use signals to signal them that a iteration is over (i.e. notifying them to wake up and process the next load of data).
```
samighoche, line 11: It's nice that you get a speedup, but you're making all threads wait for each other at every iteration. Instead, what I did was have each thread run a for loop over all iterations, but I have thread N wait for a signal that threads N-1 and N+1 are done with the previous iteration that thread N was working on, to make sure that it's using up to date information. By doing that, a thread only waits for the two threads surrounding it.

HW2/P5/P5.txt
```diff
@@ -0,0 +1,112 @@
...
+13.6152620115 simulation frames per second
+
+Using four threads, the performance increased to 18-19 frames per second.
+
+17.505442404 simulation frames per second
+18.330466401 simulation frames per second
+19.0755964671 simulation frames per second
+19.3292133848 simulation frames per second
+18.6936934528 simulation frames per second
+18.2928918469 simulation frames per second
```
samighoche, line 26: nice job taking the average of many runs

HW2/P5/P5.txt
```diff
@@ -0,0 +1,112 @@
...
+26.3824230568 simulation frames per second
+26.9027747489 simulation frames per second
+25.1775567414 simulation frames per second
+25.9558151912 simulation frames per second
+24.340345522 simulation frames per second
+26.1122358771 simulation frames per second
+25.8786248427 simulation frames per second
+25.2276822769 simulation frames per second
+
+which roughly resembles a speed up of 2x. This means, that indeed data locality seems to have an enormous effect. With a better sorting of the data, also in the case of larger chunksizes better speedups should be achievable.
```
samighoche, line 59: nice analysis

HW2/P5/P5.txt
```diff
@@ -0,0 +1,112 @@
...
+
+Spatial decomposition:
+----------------------
+Reducing the complexity from O(N^2) to O(N) yields a tremendous speed up. The feactor compared to multithreaded versions is around 6-10x! This can be seen like curing with a better medication (better algorithm) vs. fighting the symptoms only (MT).
```
samighoche, line 63: it's more like 100x right?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,112 @@
...
+1972.5432071 simulation frames per second
+
+Spatially coherent sorting:
+---------------------------
+Sorting the data and using the locality gives another little boost of around 2-300 fps. However, in the measurement below the time the sorting needs is not considered and has to be carefully traded off. In the implementation a Hilbert curve was choosen because it is considered the curve with the better locality. 
+
+2179.60372161 simulation frames per second
+2156.40830316 simulation frames per second
+2316.16303124 simulation frames per second
+2174.6190811 simulation frames per second
```
samighoche, line 83: are these for one thread or 4 threads?

HW2/P5/P5.txt
```diff
@@ -0,0 +1,112 @@
...
+2449.80431933 simulation frames per second
+2202.23428742 simulation frames per second
+2065.62870969 simulation frames per second
+2265.85375632 simulation frames per second
+
+Complete version using locking:
+-------------------------------
+Introducing locking takes off some of the performance (fps) due to the unpredictable nature of the scheduler and the overhead, but overall it did not impact the performance heavily. Still fps around 2000 are achieved, which is great!
+
+1079.9083406 simulation frames per second
```
samighoche, line 98: mine actually impacted it substantially (33%), maybe you're showing the results for 1 thread? (where there's actually no problems due to locking)

HW2/P5/driver.py
```diff
@@ -13,12 +13,41 @@
 from animator import Animator
 from physics import update, preallocate_locks
 
+# taken from http://blog.notdot.net/2009/11/Damn-Cool-Algorithms-Spatial-indexing-with-Quadtrees-and-Hilbert-Curves
+# all copyright belongs there
```
samighoche, line 17: nice!

HW2/P5/physics.pyx
```diff
@@ -55,32 +56,79 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+    V1 = &(V[i, 0])
+
+    # update 5x5 neighborhood
+    for x in range(max(0, gx - 2), min(gx + 3, grid_size)):
+        for y in range(max(0, gy - 2), min(gy + 3, grid_size)):
+            # retrieve index of cell
+            j = Grid[x, y]
+
+            # only update if pair not visited yet
+            # (this here excludes also the case where gx == x, gy == y)
```
samighoche, line 106: nice

HW2/P3/mandelbrot.pyx
```diff
@@ -9,39 +9,158 @@ from cython.parallel import prange
...
+          avxrealz_tmp = AVX.sub(AVX.mul(avxrealz, avxrealz), AVX.mul(avximagz, avximagz))
+          avximagz_tmp = AVX.mul(avxrealz, avximagz)
+          avximagz_tmp = AVX.add(avximagz_tmp, avximagz_tmp) # adding is faster and equal to * 2
+
+          # get new avxrealz by adding c
+          avxrealz = AVX.add(avxrealz_tmp, avxrealc)
+          avximagz = AVX.add(avximagz_tmp, avximagc)
+
+          # compute magnitude
+          avxmagnitude = AVX.add(AVX.mul(avxrealz, avxrealz), AVX.mul(avximagz, avximagz))
```
dekatzenel, line 124: could have stored and reused the squared values from avxrealz_temp

HW2/P4/P4.txt
dekatzenel, line 11: Agreed that creating a new set of threads every iteration is not the best way to go about this.

HW2/P5/driver.py
```diff
@@ -80,3 +114,31 @@ def randcolor():
...
             # grid if objects' indices change!  Also be sure to sort the
             # velocities with their object positions!
+
+            # Solution here uses the hilbert curve
+            # compute the hilbert ranks for all points
+            ranks = np.zeros(positions.shape[0])
+            for i in range(positions.shape[0]):
+                x = int(min(grid_size - 1, max(0, positions[i, 0] / grid_spacing)))
+                y = int(min(grid_size - 1, max(0, positions[i, 1] / grid_spacing)))
+                ranks[i] = point_to_hilbert(x, y, order=10)
```
dekatzenel, line 124: the numpy function apply_along_axis() cleans this code up a bit

HW2/P5/driver.py
```diff
@@ -80,3 +114,31 @@ def randcolor():
...
+                x = int(min(grid_size - 1, max(0, positions[i, 0] / grid_spacing)))
+                y = int(min(grid_size - 1, max(0, positions[i, 1] / grid_spacing)))
+                grid[x,y] = i
+
+            # in order to limit total computation
+            total_steps +=1
+
+        # quit after some measurements
+        if total_steps > 20:
+            break
```
dekatzenel, line 144: Is this just to prevent the simulation from running forever?

HW2/P5/physics.pyx
```diff
@@ -55,32 +56,79 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
-
-            # give a slight impulse to help separate them
-            for dim in range(2):
-                V2[dim] += eps * (XY2[dim] - XY1[dim])
 
+    # get grid index for current i
+    grid_size = int((1.0 / grid_spacing) + 1)
+
+    gx = min(grid_size - 1, max(0, <int>(XY[i, 0] / grid_spacing)))
+    gy = min(grid_size - 1, max(0, <int>(XY[i, 1] / grid_spacing)))
```
dekatzenel, line 94: elegant

HW2/P5/physics.pyx
```diff
@@ -55,32 +56,79 @@ cdef void sub_update(FLOAT[:, ::1] XY,
...
+            # (this here excludes also the case where gx == x, gy == y)
+            if j > i:
+                XY2 = &(XY[j, 0])
+                V2 = &(V[j, 0])
+                if overlapping(XY1, XY2, R):
+                    # SUBPROBLEM 4: Add locking
+                    if not moving_apart(XY1, V1, XY2, V2):
+
+                        # use locks for both i, j
+                        acquire(&locks[i])
```
dekatzenel, line 115: May be useful to acquire these locks prior to overlapping and moving_apart checks to prevent another thread from modifying in between.

dekatzenel:
Overall, strong submission

HW2/P2/parallel_vector.pyx
zezhouliu, line 84: @samighoche is correct.  Another way to do it is to check the condition again after acquiring the lock (since the condition may have changed after acquiring the lock.

PR #368, opened 2015-11-16
Hw2

ymiyamot:


HW2/P2/parallel_vector.pyx
```diff
@@ -65,51 +65,96 @@ cpdef move_data_serial(np.int32_t[:] counts,
...
+
+        
+   for r in range(repeat):
+       for idx in prange(src.shape[0], 
+                         nogil=True, 
+                         num_threads=4):
+            src_loc = src[idx]
+            dest_loc = dest[idx]
+
+            lock_num_src = (src_loc / N)
```
zezhouliu: what if these values change (e.g. src[idx] changes between now and acquiring the lock?)

PR #374, opened 2015-11-17
Hw3

1Reinier:


HW3/P3/sum.cl
```diff
@@ -0,0 +1,69 @@
...
+    // Partial sum
+
+    for (size_t j = i * k; j < i * k + k ; j++) {
+        if (j < n) {
+            sum += x[j];
+        }
+    }
+
+    fast[local_id] = sum;
+    barrier(CLK_LOCAL_MEM_FENCE);
```
oyuchoi, line 57: why do you need this barrier?

1Reinier, line 57: All `fast` sums need to be ready before we can start reducing them into `partial` sums.

PR #376, opened 2015-11-17
HW3a

Peilin-D:


HW3/P3/sum.cl
```diff
@@ -0,0 +1,86 @@
...
+    // thread with global_id 0 should add 0..k-1
+    // thread with global_id 1 should add k..2k-1
+    // thread with global_id 2 should add 2k..3k-1
+    // ...
+    //     with k = ceil(N / get_global_size()).
+    // 
+    // Be careful that each thread stays in bounds, both relative to
+    // size of x (i.e., N), and the range it's assigned to sum.
+    for (j=0;j<L;j++) { // YOUR CODE HERE
+        if(L*i+j<N)
```
adisuissa, line 57: You can add this to the above 'for' loop (to avoid the break).

HW3/P4/median_filter.cl
```diff
@@ -0,0 +1,114 @@
...
+    //
+    // We've given you median9.h, and included it above, so you can
+    // use the median9() function.
+
+    // Each thread in the valid region (x < w, y < h) should write
+    // back its 3x3 neighborhood median.
+
+    // stay in bounds
+    if ((x < w) && (y < h)) {
+        out_values[y*w+x] = median9(localGETPIX(buffer, buf_w, buf_h, buf_x-1, buf_y-1), localGETPIX(buffer, buf_w, buf_h, buf_x, buf_y-1), localGETPIX(buffer, buf_w, buf_h, buf_x+1, buf_y-1),
```
adisuissa, line 108: The use of localGETPIX is unnecessary because the buffer is already allocated with an halo, making the code safe to access the surrounding pixels inside the buffer.

PR #380, opened 2015-11-18
finished HW3

zhiqianseah:


HW3/P3/sum.cl
```diff
@@ -24,11 +27,24 @@ __kernel void sum_coalesced(__global float* x,
...
-    }
 
+    //
+
+    int offset = get_local_size(0);
+
+     while (offset != 1)
+    {
+        offset = offset >> 1;
+        fast[local_id] = fast[local_id] + fast[local_id+offset];
```
adisuissa, line 38: You are missing an if statement before this line.
You should only increment fast[local_id] if local_id < offset, otherwise, unpredicted result may occur (reading from fast[local_id+offset] that is being updated by the thread at index local_id+offset.

HW3/P4/median_filter.cl
```diff
@@ -28,7 +100,20 @@ median_3x3(__global __read_only float *in_values,
...
+    if ((y < h) && (x < w)) // stay in bounds
+
+
+        //global version
+        //out_values[y * w + x] = median9(in_values[GetPIX(x-1,y-1,w,h)], in_values[GetPIX(x-1,y,w,h)], in_values[GetPIX(x-1,y+1,w,h)],
+        //                                in_values[GetPIX(x,y-1,w,h)], in_values[GetPIX(x,y,w,h)], in_values[GetPIX(x,y+1,w,h)],
+        //                                in_values[GetPIX(x+1,y-1,w,h)], in_values[GetPIX(x+1,y,w,h)], in_values[GetPIX(x+1,y+1,w,h)]);
+
+        //local version
+        out_values[y * w + x] = median9(buffer[GetPIX(buf_x-1,buf_y-1,buf_w,buf_h)], buffer[GetPIX(buf_x-1,buf_y,buf_w,buf_h)], buffer[GetPIX(buf_x-1,buf_y+1,buf_w,buf_h)],
```
adisuissa, line 116: You don't need to call GetPIX here, because of the way the buffer was allocated and populated (for example, you'll never get buf-x to be negative or above the buffer's width).

HW3/P5/label_regions.cl
```diff
@@ -81,6 +81,49 @@ propagate_labels(__global __read_write int *labels,
...
+
+    //use only one thread
+    if (lx == 0 && ly == 0)
+    {
+        for (int y_iterator = halo; y_iterator < buf_h - halo; y_iterator++)
+        {
+            for (int x_iterator = halo; x_iterator < buf_w - halo; x_iterator++)
+           {
+
+                int temp = buffer[y_iterator * buf_w + x_iterator];
```
adisuissa, line 96: Your implementation of part 4 is wrong.
It is the same as part 2 (but with a single thread).

HW3/P5/label_regions.cl
```diff
@@ -88,10 +131,33 @@ propagate_labels(__global __read_write int *labels,
         // to adjust this for correctness.
         new_label = old_label;
```
adisuissa, line 132: After parts 2 and 4, you should use buffer[buf_w \* buf_y + buf_x] instead of old_label.

HW3/P5/label_regions.cl
```diff
@@ -88,10 +131,33 @@ propagate_labels(__global __read_write int *labels,
...
+
         if (new_label != old_label) {
             // CODE FOR PART 3 HERE
             // indicate there was a change this iteration.
             // multiple threads might write this.
+
+            atomic_min(&labels[old_label], new_label);
+
             *(changed_flag) += 1;
             labels[y * w + x] = new_label;
```
adisuissa, line 162: This should also be done atomically.

PR #401, opened 2015-11-19
Hw3

abhishekmalali:
Submission for HW3.


HW3/P3/P3.txt
```diff
@@ -0,0 +1,6 @@
+run 1- configuration ('coalesced', 128, 128): 0.00248136 seconds
```
oyuchoi, line 1: did you try other configs?

PR #406, opened 2015-11-20
Submitting, without image

neelspatel:


HW3/P3/P3.txt
```diff
@@ -0,0 +1,2 @@
+The best configuration I found was
```
oyuchoi, line 1: Any other numbers?

PR #424, opened 2015-11-20
Hw3

jacobbaron:


HW3/P3/sum.cl
```diff
@@ -24,8 +28,15 @@ __kernel void sum_coalesced(__global float* x,
...
     // See http://www.nehalemlabs.net/prototype/blog/2014/06/16/parallel-programming-with-opencl-and-python-parallel-reduce/
-    for (;;) { // YOUR CODE HERE
-        ; // YOUR CODE HERE
+    j=1;
+    offset=get_local_size(0)>>j;
+    while(offset>1) {
+    //for (j = 1; j <= lim; j++) { // YOUR CODE HERE
+        offset=get_local_size(0)>>j;
+        //printf("global_id[%d], offset=%d\n",global_id,offset);
+        fast[local_id]=fast[local_id]+fast[local_id+offset];
```
adisuissa, line 37: If you compare this to the code in the link above, you'll see that an if statement is missing.
This causes 2 issues:
1. Some threads may access data that is beyond the end of the input array.
2. The output may differ because in some iteration, given two threads A and B (where B_id = A_id + offset), B may update its value, and then A will read the new value.
This may also explain why you saw an error when using more than 32 threads, or when the global size was larger than 4096.

HW3/P3/sum.cl
```diff
@@ -48,13 +64,14 @@ __kernel void sum_blocked(__global float* x,
     // 
     // Be careful that each thread stays in bounds, both relative to
     // size of x (i.e., N), and the range it's assigned to sum.
-    for (;;) { // YOUR CODE HERE
-        ; // YOUR CODE HERE
+    for (i=global_id*k; i <= (k*(global_id+1)-1); i++) { // YOUR CODE HERE
+        sum=sum+x[i];
```
adisuissa, line 68: You should verify that you access only valid elements from x (i.e., that i < N).

HW3/P4/median_filter.cl
```diff
@@ -1,4 +1,76 @@
...
+    return s4;
+}
+
+//Return image value at point (x,y), where the indicies are relative to the global image. If the indicies are out of bounds, choose closest in bounds value instead. 
+inline float fetch_point(__global __read_only float *img, 
+                   int w, int h,
+                   int x, int y)
+{
+    float out_img;
+    while (x<0) {
```
adisuissa, line 56: It is unclear why you use a while loop, if what you are aiming for is setting the value of 'x' to 0 if it is less than 0.

HW3/P4/median_filter.cl
```diff
@@ -13,7 +85,27 @@ median_3x3(__global __read_only float *in_values,
...
+    // coordinates of the upper left corner of the buffer in image
+    // space, including halo
+    const int buf_corner_x = x - lx - halo;
+    const int buf_corner_y = y - ly - halo;
+
+    const int idx_1D = ly * get_local_size(0) + lx;
+
+    int row;
+
+    if (x < w && y < h){
```
adisuissa, line 108: Adding this line here may cause a partial population of the buffer (for example, without the column that is to the right of the right-most column), but then when reading from the buffer (lines 135-143), you may read an invalid value.

HW3/P4/median_filter.cl
```diff
@@ -21,14 +113,38 @@ median_3x3(__global __read_only float *in_values,
...
+        if (idx_1D < buf_w) {
+            for (row = 0; row < buf_h; row++) {
+                buffer[row * buf_w + idx_1D] = \
+                    fetch_point(in_values, w, h,
+                          buf_corner_x + idx_1D,
+                          buf_corner_y + row);
+          }
+        }
 
+        barrier(CLK_LOCAL_MEM_FENCE);
```
adisuissa, line 125: A barrier should be performed by all threads (not only the ones that enter the if statement).

PR #428, opened 2015-11-20
Hw3

vinayps:


HW3/P3/sum.cl
```diff
@@ -5,11 +5,15 @@ __kernel void sum_coalesced(__global float* x,
...
     size_t local_id = get_local_id(0);
+    size_t global_id = get_global_id(0);
+    uint gs = get_local_size(0);
 
     // thread i (i.e., with i = get_global_id()) should add x[i],
     // x[i + get_global_size()], ... up to N-1, and store in sum.
-    for (;;) { // YOUR CODE HERE
-        ; // YOUR CODE HERE 
+    for (uint c = global_id; c < N; c = c + get_global_size(0)) { 
+              if( global_id < N ){
```
adisuissa, line 14: Not sure if/why this is necessary.
If initially c = global_id, and the for loop condition is c<N, then why would a thread with a global_id >= N enter the loop execution?

HW3/P4/median_filter.cl
```diff
@@ -21,14 +54,36 @@ median_3x3(__global __read_only float *in_values,
...
+    barrier(CLK_LOCAL_MEM_FENCE);
 
     // Compute 3x3 median for each pixel in core (non-halo) pixels
     //
     // We've given you median9.h, and included it above, so you can
     // use the median9() function.
+       
+    if ((y < h) && (x < w)) // stay in bounds
+    {
+        buffer[buf_y * buf_w + buf_x] = median9( buffer[(buf_y - 1) * buf_w + buf_x - 1], buffer[(buf_y - 1) * buf_w + buf_x], buffer[(buf_y - 1) * buf_w + buf_x + 1],
```
adisuissa, line 77: You write into a buffer place that is concurrently being accessed by another thread.
You should either avoid this or use barriers between the reads and the write.
In this case you could have pushed the new value to the output array.

HW3/P5/label_regions.cl
```diff
@@ -80,20 +83,54 @@ propagate_labels(__global __read_write int *labels,
...
 
     // CODE FOR PARTS 2 and 4 HERE (part 4 will replace part 2)
+    if( lx == 4 && ly == 4 ) // let (4,4) be the one thread that fetches grandparents
+    {
+        // force thread (4,4) to fetch grandparents for every work-group pixel that requires it 
+        // (excluding background and repeat immediate fetches)
+        for( int y_index = 0; y_index < get_local_size(0); y_index++ )
+            for( int x_index = 0; x_index < get_local_size(0); x_index++ )
+                if( buffer[(y_index + halo) * buf_w + halo + x_index] != w*h )
+                    if( buffer[(y_index + halo) * buf_w + halo + x_index] != last_read )
```
adisuissa, line 93: If the values are equal, you should update the buffer location with the grandparent value (of the last_read location).

PR #430, opened 2015-11-21
Hw3

draj36:


HW3/P3/sum.cl
```diff
@@ -24,8 +29,17 @@ __kernel void sum_coalesced(__global float* x,
...
     //
     // See http://www.nehalemlabs.net/prototype/blog/2014/06/16/parallel-programming-with-opencl-and-python-parallel-reduce/
-    for (;;) { // YOUR CODE HERE
-        ; // YOUR CODE HERE
+    loglSize = 1;
+    temp  = lSize >> 1;
+    while (temp > 1){
+        temp = temp >> 1;
+        loglSize = loglSize + 1;
+    }
```
adisuissa, line 37: Instead of using the above while loop, you can incorporate this to following for loop,
by setting the initial j to lSize >> 1 or (lSize/2), and decrease it by half on each iteration.

HW3/P3/sum.cl
```diff
@@ -48,8 +63,16 @@ __kernel void sum_blocked(__global float* x,
...
-    for (;;) { // YOUR CODE HERE
-        ; // YOUR CODE HERE
+    lSize = get_local_size(0);
+    gID = get_global_id(0);
+    if (k-1 < N - k*gID){
+        minS = k;
+    }
+    else{
+        minS = N - k*gID;
+    }
```
adisuissa, line 73: You can simplify the above if statement, and just put in the following for loop condition:
...; j < k_(gID +1) && (k_gID + j) < N; ...

HW3/P4/median_filter.cl
```diff
@@ -12,23 +13,81 @@ median_3x3(__global __read_only float *in_values,
...
 
+    if ((y < h) && (x < w)){
+        if (lID < buf_w){ // only work with buf_w threads
+            xTemp = buf_corner_x + lID;
+            xUse = xTemp;
+            if (xTemp < 0){ // if pixel out of bounds, add compensation steps to find closest in bound pixel
+                    xUse += 1;
+            }
+            if (xTemp > w - 1){
+                xUse -= 1;
```
adisuissa, line 47: This would lead to accessing wrong values if the buffer size has more than 1 value out of bounds with the right side of the input matrix.

HW3/P4/median_filter.cl
```diff
@@ -12,23 +13,81 @@ median_3x3(__global __read_only float *in_values,
...
+            }
+            for (row = 0; row < buf_h; row++) {
+                yTemp = buf_corner_y + row;
+                yUse = yTemp;
+                if (yTemp < 0){
+                    yUse += 1;
+                }
+                if (yTemp > h - 1){
+                    yUse -= 1;
```
adisuissa, line 56: This would lead to accessing wrong values if the buffer size has more than 1 value out of bounds with the bottom of the input matrix.

HW3/P4/median_filter.cl
```diff
@@ -12,23 +13,81 @@ median_3x3(__global __read_only float *in_values,
...
-    //
-    // It may be helpful to consult HW3 Problem 5, and
-    // https://github.com/harvard-cs205/OpenCL-examples/blob/master/load_halo.cl
-    //
-    // Note that globally out-of-bounds pixels should be replaced
-    // with the nearest valid pixel's value.
+    buf_x = lx + halo;
+    buf_y = ly + halo;
 
+    if ((y < h) && (x < w)){
```
adisuissa, line 39: This if statement should not be here.
It may cause a partial population of the buffer (specifically, avoid loading the column next to the right-most input matrix column).
You should look again at HW3 Problem 5, and https://github.com/harvard-cs205/OpenCL-examples/blob/master/load_halo.cl

HW3/P5/label_regions.cl
```diff
@@ -94,6 +139,7 @@ propagate_labels(__global __read_write int *labels,
             // multiple threads might write this.
             *(changed_flag) += 1;
             labels[y * w + x] = new_label;
```
adisuissa, line 141: This update should also be done using atomic_min.

